Issue #17, opened 2020-07-22
Add caching via redis


rmeissner:
### Tasks
- [x] Add redis to service
- [x] Add possibility to enable simple caching (set/ get cache by key with timeout)
- [x] Add possibility to invalidate cache (by key)

PR #22, opened 2020-07-24
Caching via redis (#17)

rmeissner:
See #17 

- Added redis
- Added caching macro

### DIFF
--- a/src/routes/transactions.rs
+++ b/src/routes/transactions.rs
@@ -6,8 +6,8 @@ use anyhow::Result;
 
 #[get("/transactions/<safe_address>")]
 pub fn all(cache: ServiceCache, safe_address: String) -> Result<content::Json<String>> {
-    cache_resp!(cache, &safe_address, 60 * 2, { 
-        transactions::get_all_transactions(&safe_address)?
+    cache.cache_resp(&safe_address, 60 * 2, || { 
+        transactions::get_all_transactions(&safe_address)
     })
 }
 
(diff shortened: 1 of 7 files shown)


src/services/transactions.rs
```diff
@@ -8,15 +8,15 @@ use crate::models::commons::Page;
 use reqwest::Url;
 use anyhow::Result;
 
-pub fn get_about() -> String {
+pub fn get_about() -> Result<String> {
```
jpalvarezl, line 11: I will replace the `unwrap` calls with `?` in a different pull request, as we don't need to be so drastic anymore.

src/routes/transactions.rs
```diff
@@ -13,6 +17,6 @@ pub fn details(tx_hash: String) -> content::Json<String> {
 }
 
 #[get("/transactions/about")]
-pub fn about() -> String {
-    transactions::get_about()
+pub fn about(cache: ServiceCache) -> Result<content::Json<String>> {
+    cache.cache_resp(&"about_page".to_owned(), 60 * 200, transactions::get_about)
```
jpalvarezl, line 21: This is cached for 200 minutes?

src/routes/transactions.rs
```diff
@@ -1,10 +1,14 @@
...
+use crate::cache::ServiceCache;
 use crate::services::transactions;
 use rocket::response::content;
 use anyhow::Result;
 
 #[get("/transactions/<safe_address>")]
-pub fn all(safe_address: String) -> Result<content::Json<String>> {
-    Ok(content::Json(transactions::get_all_transactions(safe_address)?))
+pub fn all(cache: ServiceCache, safe_address: String) -> Result<content::Json<String>> {
+    cache.cache_resp(&safe_address, 60 * 2, || {
```
jpalvarezl, line 9: Maybe it would be good to parametrise the cache timeouts to have a single place of control for these values.

src/routes/transactions.rs
rmeissner: yes :D was just an example :P

Issue #28, opened 2020-07-27
SafeInfo fetching and caching


jpalvarezl:
**Is your feature request related to a problem? Please describe.**
The `SafeInfo` is required for calculating the status (according to #24 ). Particularly, the `threshold` and `nonce` are necessary to determine certain statuses. 

**Describe the solution you'd like**
`SafeInfo` should be fetch from the `/safe/{address}` endpoint and cached.

**Additional context**
The class layout should look like in the swagger. Here is the Kotlin equivalent:

```kotlin
data class SafeInfo(
    val address: Solidity.Address,
    val nonce: BigInteger,
    val threshold: Int,
    val owners: List<Solidity.Address>,
    val masterCopy: Solidity.Address,
    val modules: List<Solidity.Address>,
    val fallbackHandler: Solidity.Address?
)
```

PR #30, opened 2020-07-28
Closes #24 and closes #25 - Domain types and Module Transaction mapping into custom

### LINKED ISSUE #24: Transaction domain types
The clients currently deserialise the polymorphic response from the `/safe/{address}/all-transactions` into `Multisig`, `Module` and `Ethereum` transactions. Then map each of those types into `Transfer`, `Settings` or `Custom` transaction types. 

Deserialization is already largely implemented. The domain types (`Transfer`, `Settings` and `Custom`) should be defined according to the needs of the apps.

**Additional context**
For reference, the domain transaction data layouts expected by Android are as follow:

```kotlin
sealed class Transaction {
    abstract val status: TransactionStatus

    // If status is Successful, Failed or Canceled, the confirmations can be null
    abstract val confirmations: Int?

    data class Custom(
        override val status: TransactionStatus,
        override val confirmations: Int?,
        val nonce: BigInteger?,
        val address: Solidity.Address,
        val dataSize: Long,
        val date: String?,
        val value: BigInteger
    ) : Transaction()

    data class SettingsChange(
        override val status: TransactionStatus,
        override val confirmations: Int?,
        val dataDecoded: DataDecodedDto,
        val date: String?,
        val nonce: BigInteger
    ) : Transaction()

    data class Transfer(
        override val status: TransactionStatus,
        override val confirmations: Int?,
        val recipient: Solidity.Address,
        val sender: Solidity.Address,
        val value: BigInteger,
        val date: String?,
        val tokenInfo: ServiceTokenInfo?,
        val nonce: BigInteger?
    ) : Transaction()
}

enum class TransactionStatus() {
    AwaitingConfirmations,
    AwaitingExecution,
    Cancelled,
    Failed,
    Success,
    Pending // Not supported yet as these correspond to the ones issued from the device
}

```

### LINKED ISSUE #25: Map ModuleTransaction into transaction domain types
Module transactions should be mapped (at least so far) only into `Custom` as defined in #24 

There should be a mechanism to detect and map any `ModuleTransaction` into a `CustomTransaction` 

**Additional context**
For reference, the classes described in #24 will be used. For mapping logic the `TransactionRepository` implementation will be used and filling out of data will be done as needed.

jpalvarezl:
- Status calculation still missing (depends on #28)
Closes #25 (and maybe #24 too?)

### DIFF
--- a/src/models/converters/transactions.rs
+++ b/src/models/converters/transactions.rs
@@ -14,7 +14,7 @@ impl TransactionDto {
             TransactionDto::Ethereum(transaction) => Ok(transaction.to_service_transaction()),
             TransactionDto::Module(transaction) => Ok(transaction.to_service_transaction()),
             TransactionDto::Unknown => {
-                Error::msg("Unknown transaction type from backend")
+                Err(Error::msg("Unknown transaction type from backend"))
             }
         }
     }
@@ -24,7 +24,7 @@ impl MultisigTransaction {
     fn to_service_transaction(&self) -> Vec<ServiceTransaction> {
         vec!(ServiceTransaction {
             id: String::from("multisig_<something_else_eventually>"),
-            timestamp: self.execution_date.unwrap().timestamp_millis().as_(),
+            timestamp: self.execution_date.unwrap().timestamp_millis(),
             tx_status: TransactionStatus::Success,
             execution_info: None,
             tx_info: self.transaction_info(),
@@ -125,8 +125,8 @@ impl MultisigTransaction {
     fn to_custom(&self) -> Custom {
         Custom {
             to: self.safe,
-            data_size: self.data.unwrap_or(String::new()).len().as_(),
-            value: self.value.unwrap_or(String::from("0")),
+            data_size: self.data.as_ref().unwrap_or(&String::new()).len().to_string(),
+            value: self.value.as_ref().unwrap().into(),
         }
     }
 }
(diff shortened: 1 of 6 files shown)


src/models/converters/transactions.rs
```diff
@@ -115,27 +123,48 @@ impl MultisigTransaction {
     }
 
     fn to_custom(&self) -> Custom {
-        Custom { to: self.safe }
+        Custom {
+            to: self.safe,
+            data_size: self.data.as_ref().unwrap_or(&String::new()).len().to_string(),
```
rmeissner, line 128: length of data = (length of string - 2)

src/models/service/transactions.rs
```diff
@@ -3,15 +3,44 @@ use serde::Serialize;
...
+
+#[derive(Serialize, Debug)]
+#[serde(rename_all = "SCREAMING_SNAKE_CASE")]
+pub enum TransactionStatus {
+    AwaitingConfirmations,
+    AwaitingExecution,
+    Cancelled,
+    Failed,
+    Success,
+    Pending, //is this really necessary as it's a state only reachable in devices, afaik?
```
rmeissner, line 24: get rid of it

src/services/transactions.rs
```diff
@@ -45,9 +45,14 @@ pub fn get_all_transactions(safe_address: &String) -> Result<Vec<ServiceTransact
...
-    let transactions: Vec<ServiceTransaction> = transactions.results.into_iter()
-        .flat_map(|transaction| transaction.to_service_transaction())
+    let backend_transactions: Page<TransactionDto> = serde_json::from_str(&body)?;
+    let service_transactions: Vec<ServiceTransaction> = backend_transactions.results.into_iter()
+        .flat_map(|transaction| transaction.to_service_transaction().unwrap())
         .collect();
-    Ok(transactions)
+    Ok(Page {
+        count: service_transactions.len(),
+        next: backend_transactions.next,
```
rmeissner, line 54: https://github.com/rmeissner/safe-merc-service/blob/master/src/renderer/overview.rs#L44

jpalvarezl: Will address in #32

PR #33, opened 2020-07-28
Closes #32 query parameter extraction for pagination links

### LINKED ISSUE #32: Map "next" and "previous" URLs coming from the backend
Currently paged data returns the links provided by the backend as the `next` and `previous` links for pagination. This will direct the clients to the service we are trying to map from.

We want to map these URLs so that we can process them in this service.

**Additional context**
https://github.com/rmeissner/safe-merc-service/blob/master/src/renderer/overview.rs#L44

jpalvarezl:
Closes #32

### DIFF
--- a/src/services/transactions.rs
+++ b/src/services/transactions.rs
@@ -1,6 +1,6 @@
 extern crate reqwest;
 
-use crate::config::{base_transaction_service_url, request_cache_duration, scheme};
+use crate::config::{base_transaction_service_url, request_cache_duration};
 use crate::models::backend::about::About;
 use crate::models::backend::transactions::Transaction as TransactionDto;
 use crate::models::service::transactions::Transaction as ServiceTransaction;
@@ -64,7 +64,7 @@ pub fn get_all_transactions(context: &Context, safe_address: &String, next: &Opt
 fn build_paging_link(link: &String, context: &Context) -> Option<String> {
     Some(format!("{}{}?next={}",
                  context.host().unwrap(),
-                 context.origin(),
-                 Uri::percent_encode(Absolute::parse(link).ok()?.origin()?.query()?))
-    )
+                 context.path(),
+                 Uri::percent_encode(Absolute::parse(link).ok()?.origin()?.query()?)
+    ))
 }
--- a/src/services/transactions.rs
+++ b/src/services/transactions.rs
@@ -53,7 +53,6 @@ pub fn get_all_transactions(context: &Context, safe_address: &String) -> Result<
         .flat_map(|transaction| transaction.to_service_transaction(&mut info_provider).unwrap_or(vec!()))
         .collect();
     Ok(Page {
-        count: service_transactions.len(),
         next: backend_transactions.next.as_ref().and_then(|link| {
             Some(Absolute::parse(link).ok()?.origin()?.query()?.to_string())
         }),
--- a/src/services/transactions.rs
+++ b/src/services/transactions.rs
@@ -9,6 +9,7 @@ use crate::utils::context::Context;
 use crate::providers::info::InfoProvider;
 use reqwest::Url;
 use anyhow::Result;
+use rocket::http::uri::Absolute;
 
 pub fn get_about() -> Result<String> {
     let url_string = format!("{}{}", base_transaction_service_url(), "/about");
@@ -53,8 +54,12 @@ pub fn get_all_transactions(context: &Context, safe_address: &String) -> Result<
         .collect();
     Ok(Page {
         count: service_transactions.len(),
-        next: backend_transactions.next,
-        previous: backend_transactions.previous,
+        next: backend_transactions.next.as_ref().and_then(|link| {
+            Some(Absolute::parse(link).ok()?.origin()?.query()?.to_string())
+        }),
+        previous: backend_transactions.previous.as_ref().and_then(|link| {
+            Some(Absolute::parse(link).ok()?.origin()?.query()?.to_string())
+        }),
         results: service_transactions,
     })
 }
--- a/src/services/transactions.rs
+++ b/src/services/transactions.rs
@@ -38,7 +38,7 @@ pub fn get_transactions_details(tx_hash: String) -> String {
 }
 
 
-pub fn get_all_transactions(context: &Context, safe_address: &String) -> Result<Page<ServiceTransaction>> {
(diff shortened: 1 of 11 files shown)


src/services/transactions.rs
```diff
@@ -37,24 +38,37 @@ pub fn get_transactions_details(tx_hash: String) -> String {
...
     let service_transactions: Vec<ServiceTransaction> = backend_transactions.results.into_iter()
         .flat_map(|transaction| transaction.to_service_transaction(&mut info_provider).unwrap_or(vec!()))
         .collect();
     Ok(Page {
-        count: service_transactions.len(),
-        next: backend_transactions.next,
-        previous: backend_transactions.previous,
+        next: backend_transactions.next.as_ref().and_then(|link| {
+            Some(format!("{}{}?next={}",
+                         context.host().unwrap_or(String::new()),
```
rmeissner, line 60: I would throw if we cannot unwrap host as it would become an invalid url

src/services/transactions.rs
```diff
@@ -37,24 +38,37 @@ pub fn get_transactions_details(tx_hash: String) -> String {
...
     println!("{:#?}", body);
     let backend_transactions: Page<TransactionDto> = serde_json::from_str(&body)?;
     let service_transactions: Vec<ServiceTransaction> = backend_transactions.results.into_iter()
         .flat_map(|transaction| transaction.to_service_transaction(&mut info_provider).unwrap_or(vec!()))
         .collect();
     Ok(Page {
-        count: service_transactions.len(),
-        next: backend_transactions.next,
-        previous: backend_transactions.previous,
+        next: backend_transactions.next.as_ref().and_then(|link| {
```
rmeissner, line 58: Can we add a function/macro to build this url

src/services/transactions.rs
```diff
@@ -37,24 +38,37 @@ pub fn get_transactions_details(tx_hash: String) -> String {
...
+            Some(format!("{}{}?next={}",
+                         context.host().unwrap_or(String::new()),
+                         context.origin(),
+                         Uri::percent_encode(Absolute::parse(link).ok()?.origin()?.query()?)
+            ))
+        }),
+        previous: backend_transactions.previous.as_ref().and_then(|link| {
+            Some(format!("{}{}?next={}",
+                         context.host().unwrap_or(String::new()),
+                         context.origin(),
```
rmeissner, line 68: This is not correct, we should use the rocket macro else we just prepend next to the url which potentially already contains next

src/utils/context.rs
```diff
@@ -21,8 +23,18 @@ impl<'a, 'r> Context<'a, 'r> {
...
+        self.request.headers().get_one("Host").map(|host| {
+            if host.starts_with("localhost:") {
+                String::from(host)
+            } else {
+                format!("{}://{}", scheme(), host)
+            }
+        })
+    }
+
+    pub fn origin(&self) -> String {
```
rmeissner, line 36: I would call it path, but not super important

src/utils/context.rs
```diff
@@ -21,8 +23,18 @@ impl<'a, 'r> Context<'a, 'r> {
...
     }
 
-    pub fn host(&self) -> Option<&str> {
-        self.request.headers().get_one("Host")
+    pub fn host(&self) -> Option<String> {
+        self.request.headers().get_one("Host").map(|host| {
+            if host.starts_with("localhost:") {
+                String::from(host)
+            } else {
+                format!("{}://{}", scheme(), host)
```
rmeissner, line 31: Not sure if I would include the scheme here, I wouls leave it as plain host and prepend it when building the absolute uri as even for localhost a scheme should be defined

jpalvarezl: hmm I tried prepending it for localhost but firefox wouldn't recognise the string as a url

src/utils/context.rs
```diff
@@ -21,12 +22,18 @@ impl<'a, 'r> Context<'a, 'r> {
...
+        })
     }
 
-    pub fn path(&self) -> &str {
-        self.request.uri().path()
+    pub fn path(&self) -> String {
+        self.request.uri().path().to_string()
+    }
+
+    pub fn uri(&self) -> String {
```
jpalvarezl, line 35: Had to add this for the cache key. Check in your redis-cli with `$ KEYS *` you will see that now keys containing path data with the correct cache. Right now with `http://localhost:8000/transactions/0x1230B3d59858296A31053C1b8562Ecf89A2f888b?next=limit%3D20%26offset%3D20` it should crash (I have implemented a fix for this in the local branch, that's related to #36 )

src/services/transactions.rs
```diff
@@ -37,24 +38,33 @@ pub fn get_transactions_details(tx_hash: String) -> String {
...
+        next: backend_transactions.next.as_ref().and_then(|link| build_paging_link(link, context)),
+        previous: backend_transactions.previous.as_ref().and_then(|link| build_paging_link(link, context)),
         results: service_transactions,
     })
+}
+
+fn build_paging_link(link: &String, context: &Context) -> Option<String> {
+    Some(format!("{}{}?next={}",
+                 context.host().unwrap(),
+                 context.path(),
```
rmeissner, line 67: I would still  recommend to use the rocket macro here instead building the path ourselves. I am.also not sure if this solution perform url encoding on the next param (which it should)

jpalvarezl: Oh yeah, sorry forgot about it, I will address it.

src/utils.rs
```diff
@@ -48,3 +49,9 @@ pub fn hex_hash<T: Hash>(t: &T) -> String {
     t.hash(&mut s);
     format!("{:#x}", s.finish())
 }
+
+//TODO verify we are only touching 'offset' and 'limit'
```
rmeissner, line 53: I would probably call is extract query string as we can also use it for other cases then the next. Also I don't think we need to check that it is only offset and limit, if backend changes the pagination it would still work ... Probably

src/utils/context.rs
```diff
@@ -22,19 +23,23 @@ impl<'a, 'r> Context<'a, 'r> {
...
-
     pub fn path(&self) -> String {
         self.request.uri().path().to_string()
     }
 
     pub fn uri(&self) -> String {
         self.request.uri().to_string()
     }
+
+    pub fn build_paging_link(&self, origin: Origin) -> String {
```
rmeissner, line 34: I would also change this name to build_absolute_url or build_service_link ... Just more general vasically

PR #43, opened 2020-07-30
Closes #18: Dockerize project

### LINKED ISSUE #18: Dockerize project
### Tasks
- [x] Add docker support
- [ ] Check requirements by devops for deployment

### Notes
- Configure redis (see #17)

rmeissner:
Closes #18: Dockerize project
- Add `Dockerfile`
- Add `docker-compose.yml`


Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+
+COPY . .
+RUN cargo build --release
+
+# Image Stage
+FROM debian:buster-slim
+
+RUN useradd rust
+
+WORKDIR "/gateway"
```
Uxio0, line 20: I would use `/app` as workdir. Not important at all, but a lot of dockerized apps (like our services) using that "standard"

Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+
+ENV ROCKET_ENV=production
```
Uxio0, line 22: Maybe everything on the same line? `ROCKET_ENV=production ROCKET_ADDRESS=0.0.0.0 ROCKET_PORT=3666`

Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+ENV ROCKET_ADDRESS=0.0.0.0 ROCKET_PORT=3666
+EXPOSE 3666
+
+COPY --from=builder /app/target/release/safe-client-gateway ./
+
+RUN set -ex; \
```
Uxio0, line 28: I would run this before the `COPY`, so it can be cached. If not, every time a file is changed on the project `ca-certificates` and `libssl-dev` will be installed again

docker-compose.yml
```diff
@@ -0,0 +1,23 @@
...
+      - "6379:6379"
+
+  web:
+    build:
+      context: .
+      dockerfile: ./Dockerfile
+    env_file:
+      - .env
+    environment:
+      - ROCKET_DATABASES={service_cache={url="redis://redis"}}
```
Uxio0, line 19: Shouldn't this be on the `.env` file?

rmeissner: I moved it from the Rocket.toml to env for the dev configuration, but I will still overwrite it here as this is specific to this docker-compose config. Or would you recomend to create a separate env file for the docker setup (e.g. `.env.docker` for the docker-compose setup and `.env` is used for local setup)?

Uxio0: I think is better, but for now it's ok this way

PR #52, opened 2020-07-31
Closes #39: Add transaction details endpoint

### LINKED ISSUE #39: Implement detail endpoint
TBD

rmeissner:
- Closes #39: Add transaction details endpoint

### ToDos
- [x] Add token info for transfer details
- [x] Add id macro
- [x] verifiable multisig tx information (created #53)
- [x] information to execute tx (e.g. signatures) (created #53 complete support)

### DIFF
--- a/src/models/converters/transactions/details.rs
+++ b/src/models/converters/transactions/details.rs
@@ -3,7 +3,8 @@ extern crate chrono;
 use crate::models::backend::transactions::{ModuleTransaction, MultisigTransaction};
 use crate::models::commons::Operation;
 use crate::models::service::transactions::details::{
-    DetailedExecutionInfo, TransactionData, TransactionDetails,
+    DetailedExecutionInfo, ModuleExecutionDetails, MultisigExecutionDetails, TransactionData,
+    TransactionDetails,
 };
 use crate::models::service::transactions::TransactionStatus;
 use crate::providers::info::InfoProvider;
@@ -25,20 +26,24 @@ impl MultisigTransaction {
                 data_decoded: self.data_decoded.clone(),
             }),
             tx_hash: self.transaction_hash.as_ref().map(|hash| hash.to_owned()),
-            detailed_execution_info: Some(DetailedExecutionInfo {
-                nonce: self.nonce,
-                operation: self.operation.unwrap_or(Operation::CALL),
-                safe_tx_hash: self.safe_tx_hash.to_owned(),
-                signers: safe_info.owners,
-                confirmations_required: self.confirmations_required.unwrap_or(safe_info.threshold),
-                confirmations: self
-                    .confirmations
-                    .as_ref()
-                    .unwrap_or(&vec![])
-                    .into_iter()
-                    .map(|confirmation| confirmation.owner.to_owned())
-                    .collect(),
-            }),
+            detailed_execution_info: Some(DetailedExecutionInfo::Multisig(
+                MultisigExecutionDetails {
+                    nonce: self.nonce,
+                    operation: self.operation.unwrap_or(Operation::CALL),
+                    safe_tx_hash: self.safe_tx_hash.to_owned(),
+                    signers: safe_info.owners,
+                    confirmations_required: self
+                        .confirmations_required
+                        .unwrap_or(safe_info.threshold),
+                    confirmations: self
+                        .confirmations
+                        .as_ref()
+                        .unwrap_or(&vec![])
+                        .into_iter()
+                        .map(|confirmation| confirmation.owner.to_owned())
+                        .collect(),
+                },
+            )),
         })
     }
 }
@@ -55,7 +60,11 @@ impl ModuleTransaction {
                 data_decoded: self.data_decoded.clone(),
             }),
             tx_hash: Some(self.transaction_hash.to_owned()),
-            detailed_execution_info: None,
+            detailed_execution_info: Some(DetailedExecutionInfo::Module(
+                ModuleExecutionDetails {
+                    address: self.module.to_owned()
+                },
+            )),
         })
     }
 }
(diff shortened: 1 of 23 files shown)


src/models/converters/transfers.rs
```diff
@@ -1,59 +1,175 @@
...
+        &self,
+        info_provider: &mut InfoProvider,
+    ) -> Result<TransactionDetails> {
+        Ok(TransactionDetails {
+            executed_at: self.get_execution_time(),
+            submitted_at: None,
+            tx_status: TransactionStatus::Success,
+            tx_info: self.to_transfer(info_provider),
+            tx_data: None,
+            tx_hash: self.get_transaction_hash(),
```
jpalvarezl, line 42: This feels like it should be part of the `executionInfo`. Maybe I am misunderstanding here what's what.

src/models/converters/transfers.rs
```diff
@@ -1,59 +1,175 @@
...
+            _ => None,
+        }
+    }
+
+    fn get_transaction_hash(&self) -> Option<String> {
+        match self {
+            TransferDto::Erc721(transfer) => Some(transfer.transaction_hash.to_owned()),
+            TransferDto::Erc20(transfer) => Some(transfer.transaction_hash.to_owned()),
+            TransferDto::Ether(transfer) => Some(transfer.transaction_hash.to_owned()),
+            _ => None,
```
jpalvarezl, line 61: Every type appears to contain it as non-nullable, why do we have to make it nullable? Same goes for execution time. Could it be that we should probably shift where this data should go elsewhere?

src/models/service/transactions/details.rs
```diff
@@ -0,0 +1,56 @@
...
+}
+
+#[derive(Serialize, Debug)]
+#[serde(rename_all = "camelCase")]
+pub struct MultisigExecutionDetails {
+    pub nonce: u64,
+    pub safe_tx_hash: String,
+    pub signers: Vec<String>,
+    pub confirmations_required: u64,
+    pub confirmations: Vec<MultisigConfirmation>,
```
jpalvarezl, line 32: Don't we have already an implicit list of `signers` within `confirmations`, just by doing `confirmations.map { it.signer }` ? Feels a bit duplicated.

src/services/transactions_details.rs
```diff
@@ -0,0 +1,134 @@
...
+            id_parts.get(1).ok_or(anyhow::anyhow!("No safe adress"))?,
+            id_parts
+                .get(2)
+                .ok_or(anyhow::anyhow!("No module tx block"))?
+                .parse()?,
+            id_parts
+                .get(3)
+                .ok_or(anyhow::anyhow!("No module tx details hash"))?,
+        ),
+        &_ => Err(anyhow::anyhow!("Invalid details type")),
```
jpalvarezl, line 132: was not aware of that macro, nice!

src/models/converters/transactions/details.rs
```diff
@@ -0,0 +1,76 @@
+extern crate chrono;
+
+use crate::models::backend::transactions::{ModuleTransaction, MultisigTransaction};
```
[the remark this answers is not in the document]
rmeissner: Currently there is nothing meaningful to provide here

src/models/converters/transfers.rs
rmeissner: every type of transaction can have a transaction hash (of the ethereum transaction that contained it). Only case were it is not always present is for multisig transactions that have not been executed. Therefore I would have it here as it would be a field that is present for all types.

src/models/converters/transfers.rs
rmeissner: The object was UI focused. On the Ui these are separate sections that can easily be hidden if not present. Therefore the information was grouped this way.

src/models/service/transactions/details.rs
rmeissner: `signers` are all potential signers (aka `owners` ... we can rename it to that). `confirmations` refers to already submitted confirmations.

src/models/service/transactions/mod.rs
```diff
@@ -1,17 +1,16 @@
...
+pub mod details;
+pub mod summary;
+
+pub const ID_SEPERATOR: &str = "_";
+pub const ID_PREFIX_MULTISIG_TX: &str = "multisig";
+pub const ID_PREFIX_MODULE_TX: &str = "module";
+pub const ID_PREFIX_ETHEREUM_TX: &str = "ethereum";
 
 #[derive(Serialize, Debug)]
+#[serde(rename_all = "SCREAMING_SNAKE_CASE")]
```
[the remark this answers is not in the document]
rmeissner: This was to conform to our other enums. Also the clients should map this to enums on their side and not use the string directly. Therefore I would not handle that here right now (we should change it for the whole service in that case)

src/models/converters/transactions/summary.rs
```diff
@@ -0,0 +1,129 @@
...
+        );
+    };
+}
+
+macro_rules! create_id {
+    ($tx_type:expr, $($parts:expr),+) => {
+        // `stringify!` will convert the expression *as it is* into a string.
+        format!("{}{}", $tx_type, concat_parts!($($parts),+));
+    };
+}
```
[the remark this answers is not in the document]
rmeissner: I can move them there, was primarily here, because it was not used anywhere else.

src/models/converters/transactions/mod.rs
```diff
@@ -0,0 +1,204 @@
...
+                token_name: Some(token.name.to_owned()),
+                token_symbol: Some(token.symbol.to_owned()),
+                token_id: self
+                    .data_decoded
+                    .as_ref()
+                    .and_then(|it| match it.get_parameter_value("tokenId") {
+                        Some(e) => Some(e),
+                        None => it.get_parameter_value("value"),
+                    })
+                    .unwrap_or(String::from("0")),
```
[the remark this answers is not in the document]
rmeissner: https://github.com/gnosis/safe-client-gateway/issues/55

src/models/service/transactions/details.rs
jpalvarezl: ah I see, `confirmations` is a list of submitted signatures.

PR #59, opened 2020-08-04
Tests for Module and Ethereum TXs for generating TX Summary

jpalvarezl:


src/utils/context.rs
```diff
@@ -23,10 +23,6 @@ impl<'a, 'r> Context<'a, 'r> {
         self.get::<ServiceCache>()
     }
 
-    pub fn path(&self) -> String {
```
jpalvarezl, line 26: Removed unused

src/models/converters/mod.rs
```diff
@@ -1,7 +1,5 @@
 pub mod transactions;
-pub mod transfers;
-use crate::models::service::transactions::TransferDirection;
-
+pub mod transfers;use crate::models::service::transactions::TransferDirection;
```
rmeissner, line 2: ugly :P

Issue #40, opened 2020-07-30
Implement cache invalidation


rmeissner:
- [ ] Add webhook target with auth token
- [ ] Invalidate cache with pattern
- [ ] Invalidate cache for Safe when webhook is triggered

### References
- Webhook payloads: https://github.com/gnosis/safe-transaction-service/blob/2cf8365eb7af8a01b1012515ffe0e304d532741e/safe_transaction_service/history/signals.py

PR #61, opened 2020-08-04
Added endpoint for webhook (#40)

rmeissner:
Added endpoint for webhook (#40)
- Added new env variable `WEBHOOK_TOKEN` that is used to verify interactions via the webhook endpoint

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -4,3 +4,4 @@ SCHEME=http
 ROCKET_SECRET_KEY=Qt6DPFUU8qO4BKTCQnKAgt9FBBJxIWAYUGyHuruVfpE=
 ROCKET_LOG=normal
 ROCKET_DATABASES={service_cache={url="redis://127.0.0.1:6379"}}
+WEBHOOK_TOKEN=some_random_token
--- a/src/models/backend/webhooks.rs
+++ b/src/models/backend/webhooks.rs
@@ -1,0 +1,57 @@
+use serde::Deserialize;
+
+#[derive(Deserialize, Debug, Hash)]
+#[serde(tag = "type")]
+pub struct Payload {
+    pub address: String,
+    #[serde(flatten)]
+    pub details: Option<PayloadDetails>,
+}
+
+#[derive(Deserialize, Debug, Hash)]
+#[serde(tag = "type", rename_all = "SCREAMING_SNAKE_CASE")]
+pub enum PayloadDetails {
+    NewConfirmation(NewConfirmation),
+    ExecutedMultisigTransaction(ExecutedMultisigTransaction),
+    PendingMultisigTransaction(PendingMultisigTransaction),
+    IncomingEther(IncomingEther),
+    IncomingToken(IncomingToken),
+    #[serde(other)]
+    Unknown,
+}
+
+#[derive(Deserialize, Debug, Hash)]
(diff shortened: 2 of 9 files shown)


.env.sample
```diff
@@ -3,4 +3,5 @@ SCHEME=http
 # Random string (generated with openssl rand -base64 32)
 ROCKET_SECRET_KEY=Qt6DPFUU8qO4BKTCQnKAgt9FBBJxIWAYUGyHuruVfpE=
 ROCKET_LOG=normal
-ROCKET_DATABASES={service_cache={url="redis://127.0.0.1:6379"}}
\ No newline at end of file
+ROCKET_DATABASES={service_cache={url="redis://127.0.0.1:6379"}}
+WEBHOOK_TOKEN=some_random_token
```
jpalvarezl, line 7: Maybe it would be nice to mention something about this new env values in the README. Can be forked to a different issue though. Maybe for 1.0.0

src/models/backend/webhooks.rs
```diff
@@ -0,0 +1,57 @@
...
+    pub value: String,
+}
+
+#[derive(Deserialize, Debug, Hash)]
+#[serde(rename_all = "camelCase")]
+pub struct IncomingToken {
+    pub tx_hash: String,
+    pub token_address: String,
+    pub token_id: Option<String>,
+    pub value: Option<String>,
```
jpalvarezl, line 56: Maybe including `Options` for the hashing isa bit dangerous, in a sense that we don't have clear control of the nullability of values?

src/routes/hooks.rs
```diff
@@ -0,0 +1,15 @@
...
+use crate::services::hooks::invalidate_caches;
+use rocket_contrib::json::{Json};
+use anyhow::{bail, Result};
+
+#[post("/v1/hook/update/<token>", format = "json", data = "<update>")]
+pub fn update(context: Context, token: String, update: Json<Payload>) -> Result<()> {
+    if token != webhook_token() {
+        bail!("Invalid token");
+    }
+    println!("{:?}", update.details);
```
jpalvarezl, line 13: Maybe remove or change to `dbg!` like Nick suggested in a different PR?

src/services/hooks.rs
```diff
@@ -0,0 +1,7 @@
+use crate::utils::context::Context;
+use anyhow::Result;
+
+pub fn invalidate_caches(context: &Context, safe: &String) -> Result<()> {
+    println!("{}", safe);
```
jpalvarezl, line 5: Same as above, remove or change to `dbg!`?

src/routes/hooks.rs
rmeissner: That isbpart of the other pr ;)

src/services/hooks.rs
rmeissner: Same as above

src/models/backend/webhooks.rs
rmeissner: Nullability is regarded by the hash trait ... Also we could remove the hash trait here as it is not required

.env.sample
rmeissner: Yes we need to adjust our docs for 1.0.0

PR #62, opened 2020-08-04
Closes #40: Cache invalidation via webhook

### LINKED ISSUE #40: Implement cache invalidation

rmeissner:
Closes #40: Cache invalidation via webhook

### ToDos
- [x] Rebase when #61 is merged

### DIFF
--- a/src/routes/hooks.rs
+++ b/src/routes/hooks.rs
@@ -10,6 +10,5 @@ pub fn update(context: Context, token: String, update: Json<Payload>) -> Result<
     if token != webhook_token() {
         bail!("Invalid token");
     }
-    println!("{:?}", update.details);
     invalidate_caches(&context, &update.address)
 }
--- a/src/services/hooks.rs
+++ b/src/services/hooks.rs
@@ -2,7 +2,6 @@ use crate::utils::context::Context;
 use anyhow::Result;
 
 pub fn invalidate_caches(context: &Context, safe: &String) -> Result<()> {
-    println!("{}", safe);
     context.cache().invalidate_pattern(&format!("*{}*", safe));
     Ok(())
 }
--- a/src/services/hooks.rs
+++ b/src/services/hooks.rs
@@ -3,5 +3,6 @@ use anyhow::Result;
 
 pub fn invalidate_caches(context: &Context, safe: &String) -> Result<()> {
     println!("{}", safe);
+    context.cache().invalidate_pattern(&format!("*{}*", safe));
     Ok(())
 }
--- a/src/utils/cache.rs
+++ b/src/utils/cache.rs
@@ -1,8 +1,8 @@
-use rocket_contrib::databases::redis::{self, Commands};
-use serde_json;
-use serde::ser::{Serialize};
 use anyhow::Result;
 use rocket::response::content;
+use rocket_contrib::databases::redis::{self, pipe, Commands, Iter, PipelineCommands};
+use serde::ser::Serialize;
+use serde_json;
 
 #[database("service_cache")]
 pub struct ServiceCache(redis::Connection);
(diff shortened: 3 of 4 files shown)


src/utils/cache.rs
```diff
@@ -19,12 +19,29 @@ impl ServiceCache {
         let _: () = self.set_ex(id, dest, timeout).unwrap();
     }
 
+    pub fn invalidate_pattern(&self, pattern: &String) {
+        let keys: Iter<String> = self.scan_match(pattern).unwrap();
+        let pipeline = &mut pipe();
+        for key in keys {
+            pipeline.del(key);
+        }
+        // I don't know why I have to do this .... 3 f*n hours ...
```
rmeissner, line 28: Now I know a lot more about database and pooling in rocket ... also how redis macros work and what r2d2 does ...

src/services/hooks.rs
```diff
@@ -3,5 +3,6 @@ use anyhow::Result;
 
 pub fn invalidate_caches(context: &Context, safe: &String) -> Result<()> {
     println!("{}", safe);
```
rmeissner, line 5: remove debug print

src/routes/hooks.rs
```diff
@@ -0,0 +1,15 @@
...
+use crate::services::hooks::invalidate_caches;
+use rocket_contrib::json::{Json};
+use anyhow::{bail, Result};
+
+#[post("/v1/hook/update/<token>", format = "json", data = "<update>")]
+pub fn update(context: Context, token: String, update: Json<Payload>) -> Result<()> {
+    if token != webhook_token() {
+        bail!("Invalid token");
+    }
+    println!("{:?}", update.details);
```
rmeissner, line 13: remove debug print as we have implemented the functionality

src/utils/cache.rs
rmeissner: maybe @nlordell can explain me why DeRef doesn't work here ... but might be too Redis/R2D2/Rocket specific

src/services/hooks.rs
nlordell: Cool protip:
```suggestion
    dbg!(safe);
```

src/utils/cache.rs
```diff
@@ -19,12 +19,29 @@ impl ServiceCache {
...
+        let con: &redis::Connection = &self.0;
+        pipeline.execute(con);
```
nlordell, line 30: I think this could be written like this as well:
```suggestion
        pipeline.execute(&self.0);
```

However, if I understand your question correctly, its "why can't I write `pipeline.execute(self)`?" (since I think the `#[database(...)]` attribute generates a `Deref` implementation). This is because of `Deref` semantics. Usually, the compiler is smart enough to figure out that you want deref `self` into its `&Self::Target` type. However, in this case it gets confused since `execute` expects a trait object:

```rust
pub fn execute(&self, con: &dyn ConnectionLike);
```

In this particular case, there is no auto-`Deref` happening because of this. You can always force a `Deref` with the `*` operator:

```suggestion
        pipeline.execute(&*self);
        // equivalent to:
        // use std::ops::Deref;
        // pipeline.execute(self.deref());
```

Note that if this were some other method that used a concrete type as a parameter, then it would Just Work™:
```rust
fn foo(&self, con: &redis::Connection);

// This should compile:
foo(self)
```

src/utils/cache.rs
nlordell: See :point_down:

src/utils/cache.rs
```diff
@@ -19,12 +19,22 @@ impl ServiceCache {
...
     }
 
-    pub fn cache_resp<R>(&self, key: &String, timeout: usize, resp: impl Fn() -> Result<R>) -> Result<content::Json<String>>
-        where R: Serialize
+    pub fn cache_resp<R>(
+        &self,
+        key: &String,
+        timeout: usize,
+        resp: impl Fn() -> Result<R>,
+    ) -> Result<content::Json<String>>
```
jpalvarezl, line 35: Should we already return something serialized? Maybe we can make this a bit more "rocket" agnostic. Although, this beyond the scope of this PR but maybe it's better so somehow get rid of `content::Json<?>`

PR #63, opened 2020-08-05
Closes #41: About endpoint

### LINKED ISSUE #41: Update about endpoint
- [x] transaction  service endpoint
- [x] name
- [x] version
- ~~(Optional) service version~~
- ~~(Optional) buildnumber or commit~~

jpalvarezl:
Closes #41 

Changes:
- Added json structure: 
```rust
pub struct About {
    pub transaction_service_base_url: String, // base url defined in .env file
    pub name: String, // crate name
    pub version: String, // crate version
    pub transaction_service_version: Option<String>, // version returned by about endpoint of transaction service
    pub build_number: Option<String>, //commit count of master
} 
```

### DIFF
--- a/src/routes/transactions.rs
+++ b/src/routes/transactions.rs
@@ -22,5 +22,5 @@ pub fn details(context: Context, details_id: String) -> Result<content::Json<Str
 
 #[get("/about")]
 pub fn about(context: Context) -> Result<content::Json<String>> {
-    context.cache().cache_resp(&context.uri(), 60 * 200, about::get_about)
+    context.cache().cache_resp(&context.uri(), request_cache_duration(), about::get_about)
 }
--- a/src/routes/transactions.rs
+++ b/src/routes/transactions.rs
@@ -1,4 +1,4 @@
-use crate::config::request_cache_duration;
+use crate::config::{request_cache_duration, about_cache_duration};
 use crate::utils::context::Context;
 use crate::services::about;
 use crate::services::transactions_details;
@@ -22,5 +22,5 @@ pub fn details(context: Context, details_id: String) -> Result<content::Json<Str
 
 #[get("/about")]
 pub fn about(context: Context) -> Result<content::Json<String>> {
-    context.cache().cache_resp(&context.uri(), request_cache_duration(), about::get_about)
+    context.cache().cache_resp(&context.uri(), about_cache_duration(), about::get_about)
 }
--- a/src/services/about.rs
+++ b/src/services/about.rs
@@ -1,6 +1,6 @@
 extern crate reqwest;
 
-use crate::config::{base_transaction_service_url};
+use crate::config::{base_transaction_service_url, version, build_number};
 use anyhow::Result;
 use crate::models::service::about::About;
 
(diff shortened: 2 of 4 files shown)


src/services/about.rs
```diff
@@ -1,17 +1,30 @@
...
     let url = Url::parse(&url_string)?;
     let body = reqwest::blocking::get(url)?.text()?;
-    let about: About = serde_json::from_str(&body)?;
-    Ok(format!(
-        "This is an API wrapper for {}, version {}\nNo guarantees in terms of availability.",
-        about.name, about.api_version
-    ))
+    let about_dto: AboutDto = serde_json::from_str(&body)?;
+
+    let command = Command::new("git")
```
rmeissner, line 16: This should probably be a value that is set on compile time by the ci (as we cannot assume that the web worker has git or the git history). Maybe we should leave this out for now

src/services/about.rs
```diff
@@ -1,17 +1,30 @@
...
 use anyhow::Result;
 use reqwest::Url;
+use crate::models::service::about::About;
+use std::process::Command;
 
-pub fn get_about() -> Result<String> {
+pub fn get_about() -> Result<About> {
     let url_string = format!("{}{}", base_transaction_service_url(), "/about");
     let url = Url::parse(&url_string)?;
     let body = reqwest::blocking::get(url)?.text()?;
```
rmeissner, line 13: I am not sure if I would include a service version. If we don't include it we can avoid the request making this endpoint a lot simpler.

src/services/about.rs
```diff
@@ -1,17 +1,30 @@
...
 use crate::config::{base_transaction_service_url};
-use crate::models::backend::about::About;
+use crate::models::backend::about::About as AboutDto;
 use anyhow::Result;
 use reqwest::Url;
+use crate::models::service::about::About;
+use std::process::Command;
 
-pub fn get_about() -> Result<String> {
+pub fn get_about() -> Result<About> {
```
rmeissner, line 10: Can we define the cache timeout for this endpoint via the config?

rmeissner:
Theoretically in the future we could use https://doc.rust-lang.org/std/macro.option_env.html to pull the BUILD_NUMBER and VERSION number from the service at compile time. (Similar to what we do on android)

src/routes/transactions.rs
```diff
@@ -22,5 +22,5 @@ pub fn details(context: Context, details_id: String) -> Result<content::Json<Str
 
 #[get("/about")]
 pub fn about(context: Context) -> Result<content::Json<String>> {
-    context.cache().cache_resp(&context.uri(), 60 * 200, about::get_about)
+    context.cache().cache_resp(&context.uri(), request_cache_duration(), about::get_about)
```
rmeissner, line 25: I would use a different cache duration here (as this can theoretically be cached until we deploy a new service). So I would introduce a new one that defaults to a reasonable value (maybe even to the `request_cache_duration()`)

jpalvarezl: Do you mean, we should have an additional env variable `ABOUT_CACHE_DURATION` that defines that?

rmeissner: would be fine for me

PR #78, opened 2020-08-06
Closes #56 Creation transaction

### LINKED ISSUE #56: Safe creation transaction
Currently, the safe creation transaction is available as a separate endpoint in the transaction service. 
For example:  https://safe-transaction.gnosis.io/api/v1/safes/0xA063Cb7CFd8E57c30c788A0572CBbf2129ae56B6/creation/

In order to integrate it into the transaction list on the client, the client app needs then to make additional request when the last page of the all-transactions is reached, and then augment the response with the response from the endpoint above.

The solution would be to receive the safe creation transaction from the transactions response. If the creation endpoint response structure is preserved, then we need to augment it at least with a new "txType" for the creation transaction. Alternatively, if there's a way to format the response as another 'settings' transaction, that could also work.

Alternative: to request the creation transaction data from the client

**Additional context**
This appeared as part of working on requirements for the issue https://github.com/gnosis/safe/issues/315

jpalvarezl:
Closes #56 

Changes:
- Added `CreationTransaction` for backend deserialisable types
- Added `Creation` concretization of `TransactionInfo` for `TransactionSummary`
- Cached results
- Improved test structure and added test for mapping
- Moved macros to root of the crate because of visibility issues in other modules

### DIFF
--- a/src/services/transactions_list.rs
+++ b/src/services/transactions_list.rs
@@ -2,7 +2,6 @@ extern crate reqwest;
 
 use crate::config::{base_transaction_service_url, request_cache_duration};
 use crate::models::backend::transactions::{Transaction, CreationTransaction};
-use crate::models::service::transactions::{ID_PREFIX_CREATION_TX, TransactionStatus, TransactionInfo, Creation};
 use crate::models::service::transactions::summary::TransactionSummary;
 use crate::models::commons::Page;
 use crate::utils::context::Context;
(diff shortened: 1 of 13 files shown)


src/services/transactions_list.rs
```diff
@@ -41,3 +46,22 @@ pub fn get_all_transactions(context: &Context, safe_address: &String, next: &Opt
         results: service_transactions,
     })
 }
+
+fn get_creation_transaction_summary(
+    context: &Context,
+    safe: &String,
+) -> Result<TransactionSummary> {
+    let url = format!(
+        "{}/safes/{}/creation",
```
rmeissner, line 55: we should have a trailing slash here

PR #91, opened 2020-08-12
Closes #69

### LINKED ISSUE #69: Add tests for MultiSig transaction mapping to TransactionSummary
Add tests to existing `test.rs` file. Extract json.

- [x] ERC20 transfer summary
- [x] ERC721 transfer summary
- [x] Ether transfer summary
- [x] Custom transaction summary
- [x] Settings change summary

jpalvarezl:
Closes #69 
Test for Multisig mapping into:
- [x] ERC20 transfer summary
- [x] ERC721 transfer summary
- [x] Ether transfer summary
- [x] Custom transaction summary
- [x] Settings change summary


src/models/converters/transactions/tests/summary.rs
```diff
@@ -203,4 +203,168 @@ fn creation_transaction_to_summary() {
...
+            "symbol": "USDT",
+            "decimals": 18,
+            "logoUri": "https://gnosis-safe-token-logos.s3.amazonaws.com/0xD9BA894E0097f8cC2BBc9D24D308b98e36dc6D02.png"
+          },
+          "from": "0x1230B3d59858296A31053C1b8562Ecf89A2f888b"
+        }
+      ],
+      "txType": "MULTISIG_TRANSACTION"
+    }"#;
+    let token_info = TokenInfo {
```
rmeissner, line 322: This could be json too, right?

jpalvarezl: It could. I will replace and extract to a json dumpster module or something. so we can re-use it too.

src/models/converters/transactions/tests/summary.rs
```diff
@@ -291,5 +291,44 @@ fn multisig_transaction_to_erc721_transfer_summary() {
...
+    let safe_info = serde_json::from_str::<SafeInfo>(crate::json::SAFE_WITH_MODULES).unwrap();
+
+    let mut mock_info_provider = MockInfoProvider::new();
+    mock_info_provider
+        .expect_safe_info()
+        .times(1)
+        .return_once(move |_| Ok(safe_info));
+    mock_info_provider
+        .expect_token_info()
+        .times(1) // isErc20 or isErc721 check requires the token to be checked, Question: why do we use MultisigTx.to for fetching the token info?
```
jpalvarezl, line 309: Hey @rmeissner while writing the tests for this I ran into something I don't understand. The line is here: https://github.com/gnosis/safe-client-gateway/blob/3448a490b3580539c3da0c3dd65861f98f65db01/src/models/converters/transactions/mod.rs#L52

jpalvarezl: Is the `to` address used as a cache key or something?

rmeissner: Under the hood this is used as a cache key (as it should be an address)

src/models/converters/transactions/tests/summary.rs
```diff
@@ -379,4 +379,41 @@ fn multisig_transaction_to_setttins_change_summary() {
...
+
+    let mut mock_info_provider = MockInfoProvider::new();
+    mock_info_provider
+        .expect_safe_info()
+        .times(1)
+        .return_once(move |_| Ok(safe_info));
+    mock_info_provider
+        .expect_token_info()
+        .times(1)
+        .return_once(move |_| Ok(token_info));
```
jpalvarezl, line 398: Here it also seems a bit counter intuitive to be asking for the `token_info`

src/models/converters/transactions/tests/summary.rs
jpalvarezl: Should I refactor to make this request only when we've confirmed this is not a `settingsChange` nor `custom` nor `etherTransfer` ? Or could this maybe mess things up?

jpalvarezl: does the type of token transfer checking order matter, is basically my question

src/models/converters/transactions/tests/summary.rs
rmeissner: Why? You can just return None

jpalvarezl: True. I was wondering if it was a better option to simply not having to call that method at all.

jpalvarezl: Just FYI, this is not an `Option` so I have to flat out return `Error` 😂

src/models/converters/transactions/tests/summary.rs
rmeissner: We should refactor this at some point but not in this issue. I would refactor the whole logic how we check what kind of transfer it is and so on. Can you create an issue for that?

jpalvarezl: Cool, will do 👍

PR #102, opened 2020-08-17
Closes #89 DataDecoded.Parameter to include optional ValueDecoded

### LINKED ISSUE #89: Add `dataDecoded` to parameters in DataDecoded
Currently we only display `name`, `type` and `value`, but the transaction service also returns `dataDecoded` if it possible. This should be forwarded to the clients, so that they can display it.

jpalvarezl:
Closes #89 
- ~~Changed `DataDecoded` to recursively contain itself~~
- ~~Made it so that when the reference to itself is null, serialisation is skipped.~~
- Added `ValueDecoded` for `Parameters` optionally
- `Operation` and `decodedData` are serialised in an "unexpected" way, maybe we should open an issue for the transaction service backend.


jpalvarezl:
I was just looking at the responses for:
- Gateway:`/v1/transactions/multisig_0xe1da4ef7552069404077a1c2d1c007629b263ed614fbeea426572c7dfe8a5771` 
- Transaction service: https://safe-transaction.rinkeby.gnosis.io/api/v1/safes/0x1230B3d59858296A31053C1b8562Ecf89A2f888b/all-transactions/?&trusted=true

I think that the spec of #89  Doesn't match that of what's returned by the Transaction backend. Basically, there is a slightly different structure `decodedValue` which seems to be like
```rust
struct DecodedValue {
   pub values: Vec<DecodedEntry>,
}

struct DecodedEntry {
  pub operation: Operation,
  pub to: String,
  value: u64,
  data: String,
  data_decoded: DataDecoded,
}
```

Is this correct? Maybe I should close this PR and move this task to spec to review so we can talk tomorrow about it? I kind of implemented this a bit blindly without looking at the `json` first.

src/models/commons.rs
```diff
@@ -31,6 +34,17 @@ pub enum ParamValue {
     ArrayValue(Vec<ParamValue>),
 }
 
+#[derive(Serialize, Deserialize, Debug, Clone, Hash, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub struct ValueDecoded {
```
rmeissner, line 39: Didn't we want to call it internal transactions?

src/models/commons.rs
```diff
@@ -31,6 +34,17 @@ pub enum ParamValue {
...
+    pub operation: String, //TODO use operation, usually this is received as 0 or 1 but here it is `CALL`, etc. as a string
```
rmeissner, line 40: Do we want to create an issue somewhere for this? ... not sure if we want to keep this TODO in here

src/models/commons.rs
```diff
@@ -31,6 +34,17 @@ pub enum ParamValue {
...
+    pub to: String,
+    pub value: Option<u64>,
+    pub data: Option<String>,
+    #[serde(rename(deserialize = "decodedData"))] //TODO maybe we can also open an issue for this to have names consistent across structs
```
rmeissner, line 44: Yes I would do so, but it will break the web clients until they switch to our service. But would love to create an issue on the transaction service to remove this todo

Issue #54, opened 2020-08-03
Improve error handling


rmeissner:
- Proper status codes and error messages should be used (e.g. 400 for invalid ids instead of 500)

- Improve how network errors are handled internally (e.g. https://github.com/gnosis/safe-client-gateway/pull/103/files#diff-38f42f16fddf9ebfffe7f77c441e1fd5R62) so that appropriate messages can be generated

PR #103, opened 2020-08-17
Closes #90: Cache in-memory if no token info could be found

### LINKED ISSUE #90: Cache if an address is NOT a token
Currently we only cache if an address is an token, but when an address is not an token we perform the request again.

rmeissner:
Closes #90

- We already store in redis if the response errored
- Added this to in-memory cache with an Option

We should improve how `request_cached` works:
- Only cache on success (or add a flag to specify this)
- Return reqwest error to caller so that they can react to it if necessary
- This could be partially handled in https://github.com/gnosis/safe-client-gateway/issues/54 and if we see more changes necessary we create a follow up issue

### DIFF
--- a/src/providers/info.rs
+++ b/src/providers/info.rs
@@ -19,8 +19,8 @@ pub trait InfoProvider {
 pub struct DefaultInfoProvider<'p> {
     client: &'p reqwest::blocking::Client,
     cache: &'p ServiceCache,
-    safe_cache: HashMap<String, SafeInfo>,
-    token_cache: HashMap<String, TokenInfo>,
+    safe_cache: HashMap<String, Option<SafeInfo>>,
+    token_cache: HashMap<String, Option<TokenInfo>>,
 }
 
 #[derive(Serialize, Deserialize, Clone, Debug, PartialEq)]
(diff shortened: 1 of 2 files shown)


src/providers/info.rs
```diff
@@ -76,22 +76,22 @@ impl DefaultInfoProvider<'_> {
...
         let url = url.into();
         match local_cache(self).get(&url) {
-            Some(value) => Ok(value.clone()),
+            Some(value) => value.clone().ok_or(anyhow::anyhow!("Could not decode cached")),
             None => {
                 let data =
                     self.cache
                         .request_cached(self.client, &url, info_cache_duration())?;
-                let value: T = serde_json::from_str(&data)?;
+                let value: Option<T> = serde_json::from_str(&data).unwrap_or(None);
```
jpalvarezl, line 92: Aren't we drowning this panic here? Maybe we can use `unwrap_or_else` and throw a message in `dbg!` if we fail?

src/utils/cache.rs
```diff
@@ -57,9 +57,12 @@ impl ServiceCache {
...
             None => {
-                let response = client.get(url).send()?.text()?;
-                self.create(&url, &response, timeout);
-                response
+                let response = client.get(url).send()?;
+                // Don't cache if it is a Server error
+                 if response.status().is_server_error() { anyhow::bail!("Got server error for {}", url); };
+                let raw_data = response.text()?;
+                self.create(&url, &raw_data, timeout);
+                raw_data
```
jpalvarezl, line 65: Looks good. I wish I knew about the existence of `bail!` sooner xD

src/providers/info.rs
rmeissner: We are expecting a panic here, therefore I would not lock it ... this should be refactored in the future (that is what I added to #54 )

Issue #55, opened 2020-08-03
Refactor extraction of params from DataDecoded


rmeissner:
Currently it is quite tedious to extract params from DataDecode with fallbacks (e.g. `to` and `_to`). It would be nice to have a easy way to handle this.

### Note
It should be safer to use the position of the argument rather than the name of the argument.

jpalvarezl:
I got as feedback from iOS specifically that they have types defined for `SettingsChange` instead of using `dataDecoded` directly. Maybe we should factor in, including some types and enums into the `SettingsChange` definition.

rmeissner:
this would be a different issue, and question would be if they want this in a first version then we should add this to soon, as Android would require changes there too.

Theoretically we can also have this in a v2, but it would be good to know. I would suggest to create a separate issue for that.

PR #112, opened 2020-08-18
Closes #110 Include SettingsInfo in SettingsChange variant of TransactionInfo

### LINKED ISSUE #110: Refactor: SettingsChange to include struct with types for each observed method
The clients currently rely on `data_decoded` to figure out which method has been called and unpack it. This unpacking logic could be centralised in the gateway providing, in addition to `data_decoded`, a polymorphic structure representing the method call. 

Currently the `SettingsChange` variant of `TransactionInfo` looks like this:

```rust
pub struct SettingsChange {
    pub data_decoded: DataDecoded
}
```

The proposed change would make this look like so:
```rust
pub struct SettingsChange {
  pub data_decoded: DataDecoded,
  pub settings_info: SettingsInfo,
} 

pub enum SettingsInfo {
  setFallbackHandler{ address: String },
  addOwnerWithThreshold { owner: String, threshold: u64},
  removeOwner { owner: String},
  swapOwner, // missing params
  changeThreshold, // missing params
  changeMasterCopy, // missing params
  enableModule, // missing params
  disableModule, // missing params
}
```
The continued usage of `DataDecoded` in the clients seems like a reasonable cost, however, if we ever decide to extend which method call constitutes a `SettingsChange` then both apps would have to release an update, whereas with the proposed change it simply entails a small update in the gateway.

jpalvarezl:
Closes #110 Changes:
- [x] Added json for each case case:
  - [x] setFallbackHandler
  - [x] addOwnerWithThreshold
  - [x] removeOwner
  - [x] swapOwner
  - [x] changeThreshold
  - [x] changeMasterCopy
  - [x] enableModule
  - [x] disableModule
- [x] Add unit tests to deserialiser
- [x] Define enum and add structure to match expected parameters

### DIFF
--- a/src/json/commons/data_decoded_change_master_copy.json
+++ b/src/json/commons/data_decoded_change_master_copy.json
@@ -1,0 +1,10 @@
+{
+  "method": "changeMasterCopy",
+  "parameters": [
+    {
+      "name": "_masterCopy",
+      "type": "address",
+      "value": "0xb6029EA3B2c51D09a50B53CA8012FeEB05bDa35A"
+    }
+  ]
+}
--- a/src/models/converters/data_decoded.rs
+++ b/src/models/converters/data_decoded.rs
@@ -1,58 +1,54 @@
 use crate::models::commons::DataDecoded;
 use crate::models::service::transactions::SettingsInfo;
+use crate::utils::SETTINGS_CHANGE_METHODS;
 
 impl DataDecoded {
-    pub(super) fn to_settings_info(&self) -> SettingsInfo {
+    pub(super) fn to_settings_info(&self) -> Option<SettingsInfo> {
         match self.method.as_str() {
-            "setFallbackHandler" => {
-                SettingsInfo::SetFallbackHandler {
-                    handler: self.get_parameter_single_value_at(0).unwrap()
-                }
+            method_name if method_name == SETTINGS_CHANGE_METHODS[0] => {
+                Some(SettingsInfo::SetFallbackHandler {
+                    handler: self.get_parameter_single_value_at(0)?
+                })
             }
-            "addOwnerWithThreshold" => {
-                SettingsInfo::AddOwnerWithThreshold {
-                    owner: self.get_parameter_single_value_at(0).unwrap(),
-                    threshold: self.get_parameter_single_value_at(1).unwrap().parse().unwrap(),
-                }
+            method_name if method_name == SETTINGS_CHANGE_METHODS[1] => {
+                Some(SettingsInfo::AddOwner {
+                    owner: self.get_parameter_single_value_at(0)?,
+                    threshold: self.get_parameter_single_value_at(1)?.parse().ok()?,
+                })
             }
-            "removeOwner" => {
-                SettingsInfo::RemoveOwner {
-                    prev_owner: self.get_parameter_single_value_at(0).unwrap(),
-                    owner: self.get_parameter_single_value_at(1).unwrap(),
-                    threshold: self.get_parameter_single_value_at(2).unwrap().parse().unwrap(),
-                }
+            method_name if method_name == SETTINGS_CHANGE_METHODS[2] => {
+                Some(SettingsInfo::RemoveOwner {
+                    owner: self.get_parameter_single_value_at(1)?,
+                    threshold: self.get_parameter_single_value_at(2)?.parse().ok()?,
+                })
             }
-            "swapOwner" => {
-                SettingsInfo::SwapOwner {
-                    prev_owner: self.get_parameter_single_value_at(0).unwrap(),
-                    old_owner: self.get_parameter_single_value_at(1).unwrap(),
-                    new_owner: self.get_parameter_single_value_at(2).unwrap(),
-                }
+            method_name if method_name == SETTINGS_CHANGE_METHODS[3] => {
+                Some(SettingsInfo::SwapOwner {
+                    old_owner: self.get_parameter_single_value_at(1)?,
+                    new_owner: self.get_parameter_single_value_at(2)?,
+                })
             }
-            "changeThreshold" => {
-                SettingsInfo::ChangeThreshold {
-                    threshold: self.get_parameter_single_value_at(0).unwrap().parse().unwrap(),
-                }
+            method_name if method_name == SETTINGS_CHANGE_METHODS[4] => {
+                Some(SettingsInfo::ChangeThreshold {
+                    threshold: self.get_parameter_single_value_at(0)?.parse().ok()?,
+                })
             }
-            "changeMasterCopy" => {
-                SettingsInfo::ChangeImplementation {
-                    implementation: self.get_parameter_single_value_at(0).unwrap(),
-                }
+            method_name if method_name == SETTINGS_CHANGE_METHODS[5] => {
+                Some(SettingsInfo::ChangeImplementation {
+                    implementation: self.get_parameter_single_value_at(0)?,
+                })
             }
-            "enableModule" => {
-                SettingsInfo::EnableModule {
-                    module: self.get_parameter_single_value_at(0).unwrap()
-                }
+            method_name if method_name == SETTINGS_CHANGE_METHODS[6] => {
+                Some(SettingsInfo::EnableModule {
+                    module: self.get_parameter_single_value_at(0)?
+                })
             }
-            "disableModule" => {
-                SettingsInfo::DisableModule {
-                    prev_module: self.get_parameter_single_value_at(0).unwrap(),
-                    module: self.get_parameter_single_value_at(1).unwrap(),
-                }
+            method_name if method_name == SETTINGS_CHANGE_METHODS[7] => {
+                Some(SettingsInfo::DisableModule {
+                    module: self.get_parameter_single_value_at(1)?,
+                })
             }
-            _ => {
-                SettingsInfo::Unknown
-            }
+            _ => None
         }
     }
 }
--- a/src/models/converters/data_decoded.rs
+++ b/src/models/converters/data_decoded.rs
@@ -6,48 +6,48 @@ impl DataDecoded {
         match self.method.as_str() {
             "setFallbackHandler" => {
                 SettingsInfo::SetFallbackHandler {
-                    handler: self.get_parameter_single_value("handler").unwrap()
+                    handler: self.get_parameter_single_value_at(0).unwrap()
                 }
             }
             "addOwnerWithThreshold" => {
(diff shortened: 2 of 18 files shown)


src/json/commons/data_decoded_change_master_copy.json
```diff
@@ -0,0 +1,10 @@
+{
+  "method": "changeMasterCopy",
+  "parameters": [
+    {
+      "name": "_masterCopy",
```
rmeissner, line 5: always with `_`? 🤔 interesting

jpalvarezl: Here is the link if you want to investigate more: https://safe-transaction.staging.gnosisdev.com/api/v1/safes/0x1230B3d59858296A31053C1b8562Ecf89A2f888b/all-transactions/?limit=20&offset=20

jpalvarezl: I also don't get it :/

rmeissner: it is a common practise in smart contract development to differentiate between the local variable and the parameter ;) but it is theoretically not required in the abi ... but no problem :)

jpalvarezl: so should I be defensive when checking names? As in, checking with `contains` rather than exact `equals` ?

src/models/service/transactions/mod.rs
```diff
@@ -91,8 +91,16 @@ pub struct SettingsChange {
 
 #[derive(Serialize, Debug, PartialEq)]
 #[serde(rename_all = "camelCase")]
+#[serde(tag = "type")]
 pub enum SettingsInfo {
-
+    SetFallbackHandler { handler: String },
+    AddOwnerWithThreshold { owner: String, threshold: u64 },
```
rmeissner, line 97: Can we align the remove owner and add owner naming?

src/models/service/transactions/mod.rs
```diff
@@ -91,8 +91,16 @@ pub struct SettingsChange {
...
+    RemoveOwner { prev_owner: String, owner: String, threshold: u64 },
```
rmeissner, line 98: I would not return prev_owner ... this is not required in the UI

src/models/service/transactions/mod.rs
```diff
@@ -91,8 +91,16 @@ pub struct SettingsChange {
...
+    SwapOwner { prev_owner: String, old_owner: String, new_owner: String },
```
rmeissner, line 99: same here

src/models/service/transactions/mod.rs
```diff
@@ -91,8 +91,16 @@ pub struct SettingsChange {
...
+    ChangeThreshold { threshold: u64 },
+    ChangeMasterCopy { master_copy: String },
```
rmeissner, line 101: this should be ChangeImplementation and have the parameter `implementation`

src/models/service/transactions/mod.rs
```diff
@@ -91,8 +91,16 @@ pub struct SettingsChange {
...
+    EnableModule { module: String },
+    DisableModule { prev_module: String, module: String },
```
rmeissner, line 103: We do not need `prev_module` as it is not relevant in the UI

src/models/service/transactions/mod.rs
```diff
@@ -91,8 +91,16 @@ pub struct SettingsChange {
 
 #[derive(Serialize, Debug, PartialEq)]
 #[serde(rename_all = "camelCase")]
+#[serde(tag = "type")]
 pub enum SettingsInfo {
```
rmeissner, line 95: When do we need structs inside the enum and when not? (I always forget, but there were specific cases when this was required)

src/json/commons/data_decoded_change_master_copy.json
rmeissner: no, I think our backend always returns the same. but this can also be improved with #55 so that we automatically try with and without `_` and maybe allow case insensitive matching or similar

src/models/service/transactions/mod.rs
jpalvarezl: I am unsure too now. Maybe this is something we could homogenise in a refactor after the release. Generally if you use tuple enums you don't get to fiddle with the fields casings. Also, enum variants themselves cannot be used as types. So if you want to return a specific type with a method, like we do when we map from backend transaction to service transaction, a nested struct is better, there we can differentiate types and return `Transfer`, `SettingsChange` or `Custom` and not always `ServiceTransaction` . You cannot have a variant of an enum as a return type.

src/models/service/transactions/mod.rs
jpalvarezl: I am not sure I understand what you mean by this

rmeissner: One has the ending `WithThreshold` the other doesn't ... either both have it or neither

src/models/service/transactions/mod.rs
```diff
@@ -85,7 +85,48 @@ pub struct EtherTransfer {
...
+    SetFallbackHandler {
+        handler: String
+    },
+    AddOwnerWithThreshold {
+        owner: String,
+        threshold: u64,
+    },
+    RemoveOwner {
+        #[serde(skip_serializing)]
+        prev_owner: String,
```
rmeissner, line 105: I would not even have the field if we don't use it

src/models/service/transactions/mod.rs
```diff
@@ -85,7 +85,48 @@ pub struct EtherTransfer {
...
+        owner: String,
+        threshold: u64,
+    },
+    SwapOwner {
+        #[serde(skip_serializing)]
+        prev_owner: String,
```
rmeissner, line 111: same here

src/models/service/transactions/mod.rs
```diff
@@ -85,7 +85,48 @@ pub struct EtherTransfer {
...
+    },
+    ChangeImplementation {
+        implementation: String
+    },
+    EnableModule {
+        module: String
+    },
+    DisableModule {
+        #[serde(skip_serializing)]
+        prev_module: String,
```
rmeissner, line 126: and here

src/models/service/transactions/mod.rs
```diff
@@ -85,7 +85,48 @@ pub struct EtherTransfer {
...
+}
+
+#[derive(Serialize, Debug, PartialEq)]
+#[serde(rename_all = "camelCase")]
+#[serde(tag = "type")]
+pub enum SettingsInfo {
+    SetFallbackHandler {
+        handler: String
+    },
+    AddOwnerWithThreshold {
```
rmeissner, line 99: I would call this only AddOwner

src/models/converters/data_decoded.rs
```diff
@@ -0,0 +1,58 @@
+use crate::models::commons::DataDecoded;
+use crate::models::service::transactions::SettingsInfo;
+
+impl DataDecoded {
+    pub(super) fn to_settings_info(&self) -> SettingsInfo {
+        match self.method.as_str() {
+            "setFallbackHandler" => {
```
rmeissner, line 7: maybe we should make these constrants in the utils module and it would be nice to only have 1 place were we manage these string: https://github.com/gnosis/safe-client-gateway/blob/c21802dac969ad3a59b247dcd9548ef9dc6d359a/src/utils/mod.rs#L14

src/models/converters/tests/data_decoded.rs
```diff
@@ -0,0 +1,150 @@
...
+    assert_eq!(expected.settings_info, actual);
+}
+
+#[test]
+fn data_decoded_unknown_to_settings_info() {
+    let data_decoded = serde_json::from_str::<DataDecoded>(crate::json::DATA_DECODED_EXEC_TRANSACTION_FROM_MODULE).unwrap();
+
+    let expected = SettingsChange {
+        data_decoded: data_decoded.clone(),
+        settings_info: SettingsInfo::Unknown,
```
rmeissner, line 144: Would it be nicer to (or more idiomatic) to have this as an optional and if we don't know what it is we just return `null`?

src/models/converters/data_decoded.rs
```diff
@@ -0,0 +1,58 @@
+use crate::models::commons::DataDecoded;
+use crate::models::service::transactions::SettingsInfo;
+
+impl DataDecoded {
+    pub(super) fn to_settings_info(&self) -> SettingsInfo {
```
rmeissner, line 5: Should we return a result here? If any of the `unwrap` fail thenwe cannot render any tx for this safe, this seems a little risky. I would write it in a way that in case we cannot handle it or fail when trying to handle it we return our "Unknown" or `null`

src/models/service/transactions/mod.rs
jpalvarezl: I just matched the name according to the json. I  will take the rest off.

src/models/converters/tests/data_decoded.rs
rmeissner: just as a side note: I am fine either way. If at all I would ask what the clients prefer ;)

src/models/converters/data_decoded.rs
jpalvarezl: beware of the crap that I was forced to introduce because of  how the syntax for `match` arm works xD

src/models/converters/tests/data_decoded.rs
jpalvarezl: Maybe null is fine. I think I opened a ticket in the android repo where they should handle polymorphic types more gracefully, so I think it is on them how they handle "unknown"

src/models/converters/data_decoded.rs
```diff
@@ -0,0 +1,54 @@
+use crate::models::commons::DataDecoded;
+use crate::models::service::transactions::SettingsInfo;
+use crate::utils::SETTINGS_CHANGE_METHODS;
+
+impl DataDecoded {
+    pub(super) fn to_settings_info(&self) -> Option<SettingsInfo> {
+        match self.method.as_str() {
+            method_name if method_name == SETTINGS_CHANGE_METHODS[0] => {
```
rmeissner, line 8: this works for now, but I would have extracted the string inside the `SETTINGS_CHANGE_METHODS` array into separate constants and use them here. Now if we change the order in the `SETTINGS_CHANGE_METHODS` array it will break

rmeissner (approved):

src/models/converters/data_decoded.rs
jpalvarezl: I will address it

PR #116, opened 2020-08-19
Closes #115 confirmation submission date

### LINKED ISSUE #115: Add timestamp for confirmation
Confirmations of multisig transactions should have a timestamp indicating when the confirmation was submitted.

jpalvarezl:
Closes #115 

- Added optionally present `submission_date`

### DIFF
--- a/src/models/converters/transactions/tests/details.rs
+++ b/src/models/converters/transactions/tests/details.rs
@@ -70,10 +70,12 @@ fn multisig_custom_transaction_to_transaction_details() {
                     MultisigConfirmation {
                         signer: String::from("0x65F8236309e5A99Ff0d129d04E486EBCE20DC7B0"),
                         signature: Some(String::from("0x83b1506c409918f21031e93ed2f62310a5e0c05b1be89242a6a266a7de4af7bc6094e206b33387b8d4465af6087a4d2158815e613aeb186d88d9a1973e00bbe81b")),
+                        submission_time: multisig_tx.confirmations.as_ref().and_then(|confirmation| confirmation.get(0)?.submission_date.map(|it| it.timestamp_millis())),
                     },
                     MultisigConfirmation {
                         signer: String::from("0xF2CeA96575d6b10f51d9aF3b10e3e4E5738aa6bd"),
                         signature: Some(String::from("0x000000000000000000000000f2cea96575d6b10f51d9af3b10e3e4e5738aa6bd000000000000000000000000000000000000000000000000000000000000000001")),
+                        submission_time: multisig_tx.confirmations.as_ref().and_then(|confirmation| confirmation.get(1)?.submission_date.map(|it| it.timestamp_millis())),
                     },
                 ],
             })),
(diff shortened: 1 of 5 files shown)


src/models/converters/transactions/tests/details.rs
```diff
@@ -70,10 +70,12 @@ fn multisig_custom_transaction_to_transaction_details() {
                     MultisigConfirmation {
                         signer: String::from("0x65F8236309e5A99Ff0d129d04E486EBCE20DC7B0"),
                         signature: Some(String::from("0x83b1506c409918f21031e93ed2f62310a5e0c05b1be89242a6a266a7de4af7bc6094e206b33387b8d4465af6087a4d2158815e613aeb186d88d9a1973e00bbe81b")),
+                        submission_time: multisig_tx.confirmations.as_ref().and_then(|confirmation| confirmation.get(0)?.submission_date.map(|it| it.timestamp_millis())),
```
rmeissner, line 73: I would have hardcoded the number :D

src/models/service/transactions/details.rs
```diff
@@ -37,6 +37,7 @@ pub struct MultisigExecutionDetails {
 pub struct MultisigConfirmation {
     pub signer: String,
     pub signature: Option<String>,
+    pub submission_time: Option<i64>,
```
rmeissner, line 40: this should not be optional

rmeissner (changes requested):

src/models/converters/transactions/tests/details.rs
jpalvarezl: I don't know how to do that in tests :/ parsing dates with `chrono` is a bit of a pain. I would like to clean that up eventually.

src/models/service/transactions/details.rs
rmeissner: so I would expect this https://github.com/gnosis/safe-client-gateway/blob/ca68312297009dfe934043de0d96e674a41d3779/src/models/backend/transactions.rs#L86 also to be not optional

jpalvarezl: It is optional from the backend: https://github.com/gnosis/safe-client-gateway/blob/8491e453ea2e367300bd4098f87a5025e14d4060/src/models/backend/transactions.rs#L86

jpalvarezl: `confirmations` is a string in the `Multisig` variant in the docs 🤔 

https://safe-transaction.rinkeby.gnosis.io/

rmeissner: submission_date will always be present. See https://github.com/gnosis/safe-transaction-service/blob/master/safe_transaction_service/history/serializers.py#L267 (`created` is a database field that is set upon db object creation and will always be present)

src/models/converters/transactions/tests/details.rs
rmeissner: what do you mean? you know the expected number by converting the string `2020-06-22T14:58:34.055882Z` with some tool (e.g. https://codechi.com/dev-tools/date-to-millisecond-calculators/) ... and then you get: `1592837914055`

src/models/service/transactions/details.rs
jpalvarezl: alright, will remove the `Option<>` wrapper.

src/models/converters/transactions/tests/details.rs
jpalvarezl: aaah, I was trying to find how the deserialiser turns them into `DateTime<Utc>` , that's much better, thanks!

rmeissner (approved):

PR #125, opened 2020-08-24
Closes #124 : Invalidate cache with safe tx hash if possible

### LINKED ISSUE #124: Transaction Details invalidation for Multisig types
**Describe the bug**
The URL for fetching transaction details for a Multisig transaction type does not contain the safe address and therefore it is not invalidated by the webhook when there is an update to the safe

**To Reproduce**
- Propose a transaction in a safe with multiple owners
- Issue all the require signatures (do not execute)
- Request the transactions details for the particular transaction and see that the data return is out of date (e.g. status will be `AWAITING_CONFIRMATIONS` instead of `AWAITING_EXECUTION`)

**Expected behavior**
Cache should be invalidated and fresh results should be returned

rmeissner:
Closes #124 
- Invalidate cache with `safe_tx_hash` for new confirmations, for execution of multisig transactions and new pending multisig transactions

Created #125 for tests

### DIFF
--- a/src/services/hooks.rs
+++ b/src/services/hooks.rs
@@ -5,7 +5,7 @@ use anyhow::Result;
 pub fn invalidate_caches(context: &Context, payload: &Payload) -> Result<()> {
     let cache = context.cache();
     cache.invalidate_pattern(&format!("*{}*", &payload.address));
-    payload.details.map(|d| {
+    payload.details.as_ref().map(|d| {
         match d {
             PayloadDetails::NewConfirmation(data) => {
                 cache.invalidate_pattern(&format!("*{}*", data.safe_tx_hash));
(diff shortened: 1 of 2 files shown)


jpalvarezl (approved):

src/services/hooks.rs
```diff
@@ -1,7 +1,23 @@
 use crate::utils::context::Context;
+use crate::models::backend::webhooks::{Payload, PayloadDetails};
 use anyhow::Result;
 
-pub fn invalidate_caches(context: &Context, safe: &String) -> Result<()> {
-    context.cache().invalidate_pattern(&format!("*{}*", safe));
+pub fn invalidate_caches(context: &Context, payload: &Payload) -> Result<()> {
```
jpalvarezl, line 5: You didn't need to update `routes/hooks.rs` ?

PR #130, opened 2020-08-26
Closes #126 cache invalidation pattern tests for webhooks

### LINKED ISSUE #126: Add tests for webhooks
Currently we don't have any tests for webhooks. We should test that specific payload trigger specific cache invalidations.

Related to #124 and #125

jpalvarezl:
Closes #126 
- Adds tests and slight improvement in implementation leveraging traits
- Usage of `?Sized` is ok, read more [here](https://doc.rust-lang.org/nomicon/exotic-sizes.html). Seems necessary for trait objects.

### DIFF
--- a/src/services/hooks.rs
+++ b/src/services/hooks.rs
@@ -1,4 +1,4 @@
-use crate::utils::context::{Context, ContextCache};
+use crate::utils::context::ContextCache;
 use crate::models::backend::webhooks::{Payload, PayloadDetails};
 use anyhow::Result;
 use crate::utils::cache::Cache;
--- a/src/services/hooks.rs
+++ b/src/services/hooks.rs
@@ -6,10 +6,12 @@ use crate::utils::cache::Cache;
 pub fn invalidate_caches(context: &impl ContextCache, payload: &Payload) -> Result<()> {
     let cache = context.cache();
     cache.invalidate_pattern(&format!("*{}*", &payload.address));
+    println!("{:#?}", &payload);
 
     payload.details.as_ref().map(|d| {
         match d {
             PayloadDetails::NewConfirmation(data) => {
+                println!("{:#?}", &data);
                 cache.invalidate_pattern(&format!("*{}*", data.safe_tx_hash));
             }
             PayloadDetails::ExecutedMultisigTransaction(data) => {
--- a/src/services/hooks.rs
+++ b/src/services/hooks.rs
@@ -1,23 +1,28 @@
 use crate::utils::context::Context;
 use crate::models::backend::webhooks::{Payload, PayloadDetails};
 use anyhow::Result;
+use crate::utils::cache::Cache;
 
 pub fn invalidate_caches(context: &Context, payload: &Payload) -> Result<()> {
     let cache = context.cache();
+    invalidate(payload, cache);
+    Ok(())
+}
+
+pub fn invalidate(payload: &Payload, cache: &impl Cache) {
     cache.invalidate_pattern(&format!("*{}*", &payload.address));
     payload.details.as_ref().map(|d| {
         match d {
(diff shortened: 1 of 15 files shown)


src/services/hooks.rs
```diff
@@ -1,23 +1,27 @@
...
+use crate::utils::context::ContextCache;
 use crate::models::backend::webhooks::{Payload, PayloadDetails};
 use anyhow::Result;
+use crate::utils::cache::Cache;
 
-pub fn invalidate_caches(context: &Context, payload: &Payload) -> Result<()> {
+pub fn invalidate_caches(context: &impl ContextCache, payload: &Payload) -> Result<()> {
     let cache = context.cache();
     cache.invalidate_pattern(&format!("*{}*", &payload.address));
+    println!("{:#?}", &payload);
```
rmeissner, line 9: remove ;)

src/services/hooks.rs
```diff
@@ -1,23 +1,27 @@
...
+
     payload.details.as_ref().map(|d| {
         match d {
             PayloadDetails::NewConfirmation(data) => {
                 cache.invalidate_pattern(&format!("*{}*", data.safe_tx_hash));
-            },
```
rmeissner, line 12: the `,` is not needed?

src/services/hooks.rs
```diff
@@ -1,23 +1,27 @@
-use crate::utils::context::Context;
+use crate::utils::context::ContextCache;
 use crate::models::backend::webhooks::{Payload, PayloadDetails};
 use anyhow::Result;
+use crate::utils::cache::Cache;
 
-pub fn invalidate_caches(context: &Context, payload: &Payload) -> Result<()> {
+pub fn invalidate_caches(context: &impl ContextCache, payload: &Payload) -> Result<()> {
```
rmeissner, line 6: I am not the biggest fan of the `ContextCache`. Either I would say we make the whole `Context` into a trait (might be too much right now) or we just pass in `Cache` here.

If neither is easy to do right now we can stick with the `ContextCache`

src/services/tests/invalidate_caches.rs
```diff
@@ -0,0 +1,163 @@
...
+        .expect_invalidate_pattern()
+        .times(1)
+        .return_const(())
+        .with(eq(String::from("*0x65df8a1e5a40703d9c67d5df6f9b552d3830faf0507c3d7350ba3764d3a68621*")))
+        .in_sequence(&mut sequence);
+
+    let context = TestContext(mock_cache);
+
+    invalidate_caches(&context, &payload).unwrap();
+}
```
rmeissner, line 163: I like the tests already ... for completeness we should also test the other payloads in the future. 

Are there parametrized tests in rust?

src/services/hooks.rs
jpalvarezl: I think the auto formatter from Intellij decided it wasn't. It still runs the tests with no hiccups.

src/services/hooks.rs
jpalvarezl: Maybe the latter. The definition of `Context` :

```rust
pub struct Context<'a, 'r> {
    request: &'a Request<'r>,
    cache: ServiceCache,
}
```

It is coupled with the request. Ideally we have it as a trait, which could be a bit too much of a change now. Not sure what you have against though having `ContextCache`. I feel is more behaviour-driven. There is a `cache` aspect and a `request` aspect of the context, that we can selectively use.

src/services/tests/invalidate_caches.rs
jpalvarezl: I will keep this in mind for the next one :D I would have to read into it.

rmeissner (approved):

PR #137, opened 2020-08-31
Closes #54 Better error handling

### LINKED ISSUE #54: Improve error handling

jpalvarezl:
Closes #54 
- Added `ApiResult` type alias with type wrappers for most common errors
- Tests 
- Added polymorphic `message` field
- Attempted deserialisation of the backend error
- client errors are cached

### DIFF
--- a/src/utils/cache.rs
+++ b/src/utils/cache.rs
@@ -66,35 +66,40 @@ pub trait CacheExt: Cache {
         match self.fetch(&url) {
             Some(cached) => {
                 let cached_with_code = CachedWithCode::split(&cached);
-                if !cached_with_code.is_error() {
+                return if !cached_with_code.is_error() {
                     Ok(String::from(cached_with_code.data))
                 } else {
-                    return if let Ok(backend_error) = serde_json::from_str::<BackendError>(&cached_with_code.data) {
-                        Err(ApiError { status: cached_with_code.code, message: ApiErrorMessage::BackendError(backend_error) })
-                    } else {
-                        Err(ApiError { status: cached_with_code.code, message: ApiErrorMessage::SingleLine(cached_with_code.data) })
-                    };
-                }
+                    deserialise_backend_error(cached_with_code.code, &cached_with_code.data)
+                };
             }
             None => {
                 let response = client.get(url).send()?;
-                if response.status().is_server_error() {
-                    return Err(anyhow::anyhow!("Got server error for {}", url).into());
-                }
                 let status_code = response.status().as_u16();
+
+                if response.status().is_server_error() {
+                    return Err(ApiError {
+                        status: status_code,
+                        message: ApiErrorMessage::SingleLine(String::from("Got server error for {}")),
+                    });
+                }
                 let is_client_error = response.status().is_client_error();
                 let raw_data = response.text()?;
                 self.create(&url, CachedWithCode::join(status_code, &raw_data).as_str(), timeout);
-                if is_client_error {
-                    return if let Ok(backend_error) = serde_json::from_str::<BackendError>(&raw_data) {
-                        Err(ApiError { status: status_code, message: ApiErrorMessage::BackendError(backend_error) })
-                    } else {
-                        Err(ApiError { status: status_code, message: ApiErrorMessage::SingleLine(raw_data) })
-                    };
-                }
-                Ok(raw_data)
+                return if is_client_error {
+                    deserialise_backend_error(status_code, &raw_data)
+                } else {
+                    Ok(raw_data)
+                };
             }
         }
+    }
+}
+
+fn deserialise_backend_error(status_code: u16, raw_data: &str) -> ApiResult<String> {
+    if let Ok(backend_error) = serde_json::from_str::<BackendError>(&raw_data) {
+        Err(ApiError { status: status_code, message: ApiErrorMessage::BackendError(backend_error) })
+    } else {
+        Err(ApiError { status: status_code, message: ApiErrorMessage::SingleLine(raw_data.to_owned()) })
     }
 }
 
@@ -108,7 +113,7 @@ impl CachedWithCode {
     const SEPARATOR: &'static str = ";";
 
     pub(super) fn split(cached: &str) -> Self {
-        let cached_with_code: Vec<&str> = cached.split(CachedWithCode::SEPARATOR).collect();
+        let cached_with_code: Vec<&str> = cached.splitn(2, CachedWithCode::SEPARATOR).collect();
         CachedWithCode {
             code: cached_with_code.get(0).expect("Must have a status code").parse().expect("Not a valid Http code"),
             data: cached_with_code.get(1).expect("Must have data").to_string(),
@@ -120,7 +125,7 @@ impl CachedWithCode {
     }
 
     pub(super) fn is_error(&self) -> bool {
-        500 > self.code && self.code >= 400
+        200 < self.code && self.code >= 400
     }
 }
 
--- a/src/utils/cache.rs
+++ b/src/utils/cache.rs
@@ -3,7 +3,7 @@ use rocket_contrib::databases::redis::{self, pipe, Commands, Iter, PipelineComma
 use serde::ser::Serialize;
 use serde_json;
 use mockall::automock;
-use crate::utils::errors::{ApiResult, ApiError, ApiErrorMessage};
+use crate::utils::errors::{ApiResult, ApiError};
 
 #[database("service_cache")]
 pub struct ServiceCache(redis::Connection);
@@ -77,10 +77,7 @@ pub trait CacheExt: Cache {
                 let status_code = response.status().as_u16();
 
                 if response.status().is_server_error() {
-                    return Err(ApiError {
-                        status: status_code,
-                        message: ApiErrorMessage::SingleLine(String::from("Got server error for {}")),
-                    });
+                    return Err(ApiError::from_backend_error(42, format!("Got server error for {}", response.text()?).as_str()));
                 }
                 let is_client_error = response.status().is_client_error();
                 let raw_data = response.text()?;
--- a/src/utils/cache.rs
+++ b/src/utils/cache.rs
@@ -3,7 +3,7 @@ use rocket_contrib::databases::redis::{self, pipe, Commands, Iter, PipelineComma
 use serde::ser::Serialize;
 use serde_json;
 use mockall::automock;
-use crate::utils::errors::{ApiResult, ApiError, BackendError, ApiErrorMessage};
+use crate::utils::errors::{ApiResult, ApiError, ApiErrorMessage};
 
 #[database("service_cache")]
 pub struct ServiceCache(redis::Connection);
@@ -69,7 +69,7 @@ pub trait CacheExt: Cache {
                 return if !cached_with_code.is_error() {
                     Ok(String::from(cached_with_code.data))
                 } else {
-                    deserialise_backend_error(cached_with_code.code, &cached_with_code.data)
+                    Err(ApiError::from_backend_error(cached_with_code.code, &cached_with_code.data))
                 };
             }
             None => {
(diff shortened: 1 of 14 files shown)


src/utils/errors.rs
```diff
@@ -0,0 +1,92 @@
...
+pub type ApiResult<T, E = ApiError> = Result<T, E>;
+
+#[derive(Error, Debug, Serialize)]
+pub struct ApiError {
+    pub status: u16,
+    pub message: ApiErrorMessage,
+}
+
+#[derive(Debug, Serialize, PartialEq)]
+#[serde(untagged)]
```
jpalvarezl, line 21: Maybe should be tagged? For the clients' easy consumption.

src/utils/cache.rs
```diff
@@ -62,20 +62,65 @@ pub trait CacheExt: Cache {
...
+}
+
+impl CachedWithCode {
+    const SEPARATOR: &'static str = ";";
+
+    pub(super) fn split(cached: &str) -> Self {
+        let cached_with_code: Vec<&str> = cached.split(CachedWithCode::SEPARATOR).collect();
+        CachedWithCode {
+            code: cached_with_code.get(0).expect("Must have a status code").parse().expect("Not a valid Http code"),
+            data: cached_with_code.get(1).expect("Must have data").to_string(),
```
rmeissner, line 114: We should take the rest not just the second part right? E.g. if we cache the following `404:{"this_is_some_json":"And here we have a message with ;"}` then then we would have 3 parts and your `data` part would be invalid.

src/utils/cache.rs
```diff
@@ -62,20 +62,65 @@ pub trait CacheExt: Cache {
...
+        }
+    }
+
+    pub(super) fn join(code: u16, data: &str) -> String {
+        format!("{}{}{}", code, CachedWithCode::SEPARATOR, data)
+    }
+
+    pub(super) fn is_error(&self) -> bool {
+        500 > self.code && self.code >= 400
```
rmeissner, line 123: I would maybe check here if < 200 or >= 400 (so not a 2xx or 3xx status)

src/utils/cache.rs
```diff
@@ -1,19 +1,19 @@
...
 #[database("service_cache")]
 pub struct ServiceCache(redis::Connection);
 
 #[automock]
 pub trait Cache {
     fn fetch(&self, id: &str) -> Option<String>;
-    fn create(&self, id: &String, dest: &String, timeout: usize);
-    fn invalidate_pattern(&self, pattern: &String);
-    fn invalidate(&self, id: &String);
+    fn create(&self, id: &str, dest: &str, timeout: usize);
```
rmeissner, line 14: Why switch to `str` with this PR? Can we use `impl Into<String>` then it can handle `str` and `String`

src/utils/cache.rs
```diff
@@ -62,20 +62,65 @@ pub trait CacheExt: Cache {
...
+                        Err(ApiError { status: cached_with_code.code, message: ApiErrorMessage::SingleLine(cached_with_code.data) })
+                    };
                 }
-            };
-        Ok(data)
+            }
+            None => {
+                let response = client.get(url).send()?;
+                if response.status().is_server_error() {
+                    return Err(anyhow::anyhow!("Got server error for {}", url).into());
```
rmeissner, line 82: Why not use an ApiError here?

src/utils/cache.rs
```diff
@@ -62,20 +62,65 @@ pub trait CacheExt: Cache {
...
+                }
+                let status_code = response.status().as_u16();
+                let is_client_error = response.status().is_client_error();
+                let raw_data = response.text()?;
+                self.create(&url, CachedWithCode::join(status_code, &raw_data).as_str(), timeout);
+                if is_client_error {
+                    return if let Ok(backend_error) = serde_json::from_str::<BackendError>(&raw_data) {
```
rmeissner, line 89: We have this logic multiple time, can we extract it into a method?

src/utils/errors.rs
rmeissner: They should be serialized all in the same format and that is:
```json
{
    "code": 1,
    "message": "Checksum address validation failed",
    "arguments": [
        "some_address"
    ]
}
```

I would use the polymorphism only internally to make it easier to create the error, but if you parse the backend error I would always use this for rendering out ... as said for error happening in the client gateway I would propose status code "1337" (or something else agreed on with the client teams), and proper message + arguments.

src/utils/tests/cache.rs
```diff
@@ -0,0 +1,49 @@
+use crate::utils::cache::CachedWithCode;
```
rmeissner, line 1: We are missinga tests with multiple seperators: `404;foo;bar`. For that I would want to have
```
CachedWithCode {
  code: 404,
  data: String::from("foo;bar"),
};
```

rmeissner (changes requested):

src/utils/cache.rs
jpalvarezl: I think it's generally better to use `&str` (string slice) as it's immutable and generally we are just borrowing the value: https://stackoverflow.com/questions/24158114/what-are-the-differences-between-rusts-string-and-str#:~:text=A%20Rust%20String%20is%20like,contents%20of%20std%3A%3Astring%20.

src/utils/cache.rs
jpalvarezl: I was too lazy to initialise `status_code` earlier xD

jpalvarezl:
Thanks for spotting that test case with the `separator` in the payload, totally missed it!

Not sure if this comment: https://github.com/gnosis/safe-client-gateway/pull/137#discussion_r483590549 is immediately actionable. Is this something I should address in this PR?

rmeissner:
Yes, one of the goals of this PR is to provide better error reporting for the clients, and as the client gateway is specialized for the clients we should try to return errors in a format that is usable for them. If this is way too much effort we need to reevaluate this.

src/utils/cache.rs
```diff
@@ -66,38 +66,43 @@ pub trait CacheExt: Cache {
...
+                    deserialise_backend_error(status_code, &raw_data)
+                } else {
+                    Ok(raw_data)
+                };
             }
         }
     }
 }
 
+fn deserialise_backend_error(status_code: u16, raw_data: &str) -> ApiResult<String> {
```
rmeissner, line 98: This could be a `::new` method of the `ApiError`

src/utils/cache.rs
```diff
@@ -120,7 +125,7 @@ impl CachedWithCode {
     }
 
     pub(super) fn is_error(&self) -> bool {
-        500 > self.code && self.code >= 400
+        200 < self.code && self.code >= 400
```
rmeissner, line 128: So if `self.code` is greater than 200 then it is an error?  That would mean 201 is an error.

rmeissner: should be `self.code < 200 || self.code >= 400`

jpalvarezl: not sure in what mental state I was in when I wrote that "correction"

src/utils/cache.rs
jpalvarezl: I left `new` as private for now. And exposed a different method so we handle the inner works just in `errors.rs` for now. Not sure if that's the best, though.

src/utils/errors.rs
rmeissner: So this is still not correctly serialized. What I expect:
- Gateway error:
Status code: 500
Payload:
```json
{
    "code": 1337,
    "message": "Serde serialization error: <serde error message>"
}
``` 
- Well formatted service error:
Status code: <service status code>
Payload: <service error payload>, e.g.:
```json
{
    "code": 1,
    "message": "Checksum address validation failed",
    "arguments": [
        "some_address"
    ]
}
``` 
- Unformatted service error:
Status code: <service status code>
Payload:
```json
{
    "code": 42,
    "message": "Unknown service error: <service response>",
}
``` 

Something along these lines, this way every error returned by the client gateway is of the format:

```json
{
    "code": <error code>,
    "message": "<error message>",
    "arguments": [
        <0..n arguments>
    ]
}
```

We should communicate with the clients that we use new status code and if necessary align them with them (if they prefer something else than `1337` and `42`)

rmeissner (changes requested):

src/utils/errors.rs
```diff
@@ -0,0 +1,114 @@
...
+
+impl ApiError {
+    pub fn from_backend_error(status_code: u16, raw_error: &str) -> ApiError {
+        match serde_json::from_str::<ErrorDetails>(&raw_error) {
+            Ok(backend_error) => {
+                ApiError::new(status_code, backend_error)
+            }
+            Err(error) => {
+                let error_details = ErrorDetails {
+                    code: 1337,
```
rmeissner, line 36: I would make this an error code `42` and return the `raw_error` as the message (not the serde error)

src/utils/errors.rs
```diff
@@ -0,0 +1,114 @@
...
+pub struct ErrorDetails {
+    pub code: u64,
+    pub message: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub arguments: Option<Vec<String>>,
+}
+
+impl ApiError {
+    pub fn from_backend_error(status_code: u16, raw_error: &str) -> ApiError {
+        match serde_json::from_str::<ErrorDetails>(&raw_error) {
```
rmeissner, line 30: You could move the `ApiError::new` before the match

src/utils/errors.rs
```diff
@@ -0,0 +1,114 @@
...
+use rocket::http::{ContentType, Status};
+use serde_json;
+use serde::{Serialize, Deserialize};
+use std::io::Cursor;
+use log::debug;
+use anyhow::Result;
+
+pub type ApiResult<T, E = ApiError> = Result<T, E>;
+
+#[derive(Error, Debug, Serialize)]
```
rmeissner, line 14: We should not need Serialize here

src/utils/errors.rs
```diff
@@ -0,0 +1,114 @@
...
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "ApiError({}: {})", self.status, self.details)
+    }
+}
+
+impl<'r> Responder<'r> for ApiError {
+    fn respond_to(self, _: &Request) -> response::Result<'r> {
+        debug!("Handle ApiError");
+        Response::build()
+            .sized_body(Cursor::new(serde_json::to_string(&self).expect("Couldn't serialize error")))
```
rmeissner, line 70: we should render out `&self.details` as the body

src/utils/errors.rs
```diff
@@ -0,0 +1,114 @@
...
+            .ok()
+    }
+}
+
+impl From<anyhow::Error> for ApiError {
+    fn from(err: anyhow::Error) -> Self {
+        Self {
+            status: 500,
+            details: ErrorDetails {
+                code: 42,
```
rmeissner, line 82: I would have this as a `1337` as it is an error created in the gateway

src/utils/errors.rs
```diff
@@ -0,0 +1,114 @@
...
+        }
+    }
+}
+
+impl From<reqwest::Error> for ApiError {
+    fn from(err: reqwest::Error) -> Self {
+        Self {
+            status: 500,
+            details: ErrorDetails {
+                code: 42,
```
rmeissner, line 95: same here (code as `1337`) as this is mostly an error because we could not open the connection or have an invalid ssl ceritificate

src/utils/errors.rs
```diff
@@ -0,0 +1,114 @@
...
+            },
+        }
+    }
+}
+
+impl From<serde_json::error::Error> for ApiError {
+    fn from(err: serde_json::error::Error) -> Self {
+        Self {
+            status: 500,
+            details: ErrorDetails {
```
rmeissner, line 107: We have very smilar logic 3 times, maybe we could extract this into a method

rmeissner (changes requested):

src/utils/errors.rs
jpalvarezl: hmmm somehow I do. Why do you expect it to not bee needed? I think it's necessary for the `Responder` implementation of `ApiError`

jpalvarezl: Never mind that previous comment. You are right, once in the responder impl I use `self.details` instead, `Serialize` becomes totally unnecessary 👍

src/utils/cache.rs
```diff
@@ -62,20 +62,59 @@ pub trait CacheExt: Cache {
...
-                    // Don't cache if it is a Server error
-                    if response.status().is_server_error() { anyhow::bail!("Got server error for {}", url); };
-                    let raw_data = response.text()?;
-                    self.create(&url, &raw_data, timeout);
-                    raw_data
+    ) -> ApiResult<String> {
+        match self.fetch(&url) {
+            Some(cached) => {
+                let cached_with_code = CachedWithCode::split(&cached);
+                return if !cached_with_code.is_error() {
```
rmeissner, line 69: Theoretically we could extract this into a method (building the response based on a boolean from a code + data)

rmeissner (approved):

src/utils/cache.rs
jpalvarezl: I added a method `to_result` in the `impl CacheWithCode` block.

PR #139, opened 2020-09-03
Closes #53 Fields for sign submission

### LINKED ISSUE #53: Add information required to submit transaction
It should be possible to submit a multisig transaction for execution using the information provided by the details endpoint.

See https://docs.google.com/document/d/1Bu9-CM9Fy-9duNqb6h6MlgzTE4oxFHDyCGAfBitIF6M/edit#

jpalvarezl:
Closes #53 
- Added fields for signing transactions (in the required order) as part of the `MultsigExecutionDetails`
- Adding gas token info optionally, when available

### DIFF
--- a/src/models/converters/transactions/details.rs
+++ b/src/models/converters/transactions/details.rs
@@ -2,7 +2,7 @@ extern crate chrono;
 
 use crate::models::backend::transactions::{ModuleTransaction, MultisigTransaction};
 use crate::models::commons::Operation;
-use crate::models::service::transactions::details::{DetailedExecutionInfo, ModuleExecutionDetails, MultisigConfirmation, MultisigExecutionDetails, TransactionData, TransactionDetails, SignatureInfo};
+use crate::models::service::transactions::details::{DetailedExecutionInfo, ModuleExecutionDetails, MultisigConfirmation, MultisigExecutionDetails, TransactionData, TransactionDetails};
 use crate::models::service::transactions::TransactionStatus;
 use crate::providers::info::{InfoProvider, SafeInfo, TokenInfo};
 use anyhow::Result;
@@ -13,7 +13,17 @@ impl MultisigTransaction {
         info_provider: &mut dyn InfoProvider,
     ) -> Result<TransactionDetails> {
         let safe_info = info_provider.safe_info(&self.safe.to_string())?;
-        let gas_token = self.gas_token.as_ref().map(|token_address| info_provider.token_info(&token_address).ok()).flatten();
+        let gas_token = if let Some(gas_token) = &self.gas_token {
+            if gas_token != "0x0000000000000000000000000000000000000000" {
+                info_provider.token_info(gas_token).ok()
+            } else {
+                None
+            }
+        } else {
+            None
+        };
+
+        //self.gas_token.as_ref().map(|token_address| info_provider.token_info(&token_address).ok()).flatten();
         Ok(TransactionDetails {
             executed_at: self.execution_date.map(|data| data.timestamp_millis()),
             tx_status: self.map_status(&safe_info),
@@ -30,7 +40,7 @@ impl MultisigTransaction {
         })
     }
 
-    fn build_execution_details(&self, safe_info: SafeInfo, gas_token: Option<TokenInfo>) -> MultisigExecutionDetails {
+    fn build_execution_details(&self, safe_info: SafeInfo, gas_token_info: Option<TokenInfo>) -> MultisigExecutionDetails {
         MultisigExecutionDetails {
             submitted_at: self.submission_date.timestamp_millis(),
             nonce: self.nonce,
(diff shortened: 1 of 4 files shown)


src/models/converters/transactions/details.rs
```diff
@@ -28,31 +26,45 @@ impl MultisigTransaction {
...
+        }
+    }
+
+    fn build_signature_info(&self, gas_token: Option<TokenInfo>) -> SignatureInfo {
+        SignatureInfo {
+            refund_receiver: self.refund_receiver.to_owned(),
+            base_gas: self.base_gas,
+            safe_tx_gas: self.safe_tx_gas,
+            gas_limit: self.gas_price.to_owned(),
+            gas_token: gas_token.as_ref().map(|it| it.address.to_owned()).unwrap_or(String::from("0x0000000000000000000000000000000000000000")),
```
rmeissner, line 64: the `gas_token` field should be taken from the `MultisigTransaction` and not from the `TokenInfo` else it might be that we map an unknown token to "0x0000000000000000000000000000000000000000"

src/models/converters/transactions/details.rs
```diff
@@ -28,31 +26,45 @@ impl MultisigTransaction {
...
+                    signature: confirmation.signature.to_owned(),
+                    submitted_at: confirmation.submission_date.timestamp_millis(),
+                })
+                .collect(),
+            signature_info: Some(self.build_signature_info(gas_token)),
+        }
+    }
+
+    fn build_signature_info(&self, gas_token: Option<TokenInfo>) -> SignatureInfo {
+        SignatureInfo {
```
rmeissner, line 59: Why add another level for this and not put it on the same level as `safe_tx_hash` and `nonce`? Nonce is similar to all the info inside this struct, as it also is only used for signing.

src/models/converters/transactions/details.rs
```diff
@@ -28,31 +26,45 @@ impl MultisigTransaction {
...
+            refund_receiver: self.refund_receiver.to_owned(),
+            base_gas: self.base_gas,
+            safe_tx_gas: self.safe_tx_gas,
+            gas_limit: self.gas_price.to_owned(),
```
rmeissner, line 63: This was a typo in the doc and should be gas price. Gas limit is different to gas price ;)

src/models/service/transactions/details.rs
```diff
@@ -30,6 +31,19 @@ pub struct MultisigExecutionDetails {
...
+    pub signature_info: Option<SignatureInfo>,
+}
+
+#[derive(Serialize, Debug, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub struct SignatureInfo {
+    pub refund_receiver: Option<String>,
+    pub base_gas: Option<usize>,
+    pub safe_tx_gas: Option<usize>,
+    pub gas_limit: Option<String>,
```
rmeissner, line 43: As mentioned above this should be `gas_price`

src/models/service/transactions/details.rs
```diff
@@ -30,6 +31,19 @@ pub struct MultisigExecutionDetails {
     pub signers: Vec<String>,
     pub confirmations_required: u64,
     pub confirmations: Vec<MultisigConfirmation>,
+    pub signature_info: Option<SignatureInfo>,
+}
+
+#[derive(Serialize, Debug, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub struct SignatureInfo {
+    pub refund_receiver: Option<String>,
```
rmeissner, line 40: In this struct there should be no option

src/models/service/transactions/details.rs
```diff
@@ -30,6 +31,19 @@ pub struct MultisigExecutionDetails {
     pub signers: Vec<String>,
     pub confirmations_required: u64,
     pub confirmations: Vec<MultisigConfirmation>,
+    pub signature_info: Option<SignatureInfo>,
```
rmeissner, line 34: `SignatureInfo` should not be an option and I would have all the fields of that struct directly in the `MultisigExecutionDetails` struct (see doc, as there optional fields are mentioned, everything else should not be optional)

rmeissner (changes requested):

src/models/converters/transactions/details.rs
jpalvarezl: I wanted to keep this in isolation to prevent having to modify tests too soon, before I was sure of what was necessary. I will flatten the structure.

src/models/converters/transactions/details.rs
```diff
@@ -13,7 +13,17 @@ impl MultisigTransaction {
         info_provider: &mut dyn InfoProvider,
     ) -> Result<TransactionDetails> {
         let safe_info = info_provider.safe_info(&self.safe.to_string())?;
-        let gas_token = self.gas_token.as_ref().map(|token_address| info_provider.token_info(&token_address).ok()).flatten();
+        let gas_token = if let Some(gas_token) = &self.gas_token {
```
rmeissner, line 16: Wouldn't it make sense to have this logic in the info provider as it applies everywhere?

src/models/converters/transactions/details.rs
```diff
@@ -51,18 +61,12 @@ impl MultisigTransaction {
...
                 .collect(),
-            signature_info: Some(self.build_signature_info(gas_token)),
-        }
-    }
-
-    fn build_signature_info(&self, gas_token: Option<TokenInfo>) -> SignatureInfo {
-        SignatureInfo {
-            refund_receiver: self.refund_receiver.to_owned(),
+            refund_receiver: self.refund_receiver.as_ref().unwrap_or(&String::from("0x0000000000000000000000000000000000000000")).to_owned(),
+            gas_token: self.gas_token.as_ref().unwrap_or(&String::from("0x0000000000000000000000000000000000000000")).to_owned(),
```
rmeissner, line 65: very unimportant: could we have the order of the params similar to how it is in the contracts -> nonce, safeTxGas, baseGas, gasPrice, gasToken, refundReceiver

src/models/converters/transactions/details.rs
jpalvarezl: I will push it down then.

rmeissner (approved):

PR #140, opened 2020-09-04
Add default values for host and port to .env.sample for discoverability

biafra23:
Having the default values for host and port in the `.env.sample` make it easier for newcomers to find the place to change them if they need to. The `ROCKET_PORT` needs to be changed if there is already another service running on 8000. The `ROCKET_ADDRESS` needs to be changed if the user wants to connect to the service from a different device/host.


.env.sample
```diff
@@ -4,5 +4,7 @@ SCHEME=http
 ROCKET_SECRET_KEY=Qt6DPFUU8qO4BKTCQnKAgt9FBBJxIWAYUGyHuruVfpE=
 ROCKET_LOG=normal
 ROCKET_DATABASES={service_cache={url="redis://127.0.0.1:6379"}}
+ROCKET_PORT=8000
+ROCKET_ADDRESS=localhost
```
rmeissner, line 8: Does this break docker compose if you 1:1 copy it to .env? (In the docker file we build the prod version and there the default address is `0.0.0.0` and not localhost. If this is overwritten it might be that the service is not reachable ... not sure how the port forwarding works)

biafra23: I am totally fine with 0.0.0.0 if that solves it. I thought the default is localhost. Let me check if it is a problem.

rmeissner: It depends on your environment. Docker is configured to use the release env and local uses the dev environment, both have different default (as development should not be reachable externally)

biafra23: When running the service with `docker-compose up` it is not reachable via telnet either way. Having `0.0.0.0` or `localhost` in .env don't work. The connection is established and dropped immediately. Usually it is necessary to expose ports with docker. But I don't know how that works with `docker-compose up`.

rmeissner: The question was more if the web server is accessible either way.

rmeissner (approved):

PR #141, opened 2020-09-07
Closes #114 Method name for custom transactions in summary

### LINKED ISSUE #114: Return method for smart contract interactions
For the web UI we want to show the method that is being called through contract interactions also in the transaction overview list. 

image

For that, the gateway should return this information.

jpalvarezl:
Closes #114 

- Including `data_decoded` in Custom `TransactionInfo` to have access to method name and params

### DIFF
--- a/src/models/converters/transactions/mod.rs
+++ b/src/models/converters/transactions/mod.rs
@@ -179,6 +179,7 @@ impl MultisigTransaction {
             to: self.to.to_owned(),
             data_size: data_size(&self.data),
             value: self.value.as_ref().unwrap().into(),
+            data_decoded: self.data_decoded.to_owned(),
         }
     }
 }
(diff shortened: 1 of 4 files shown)


rmeissner:
We only require the method name, not the whole decoded data

src/models/converters/transactions/mod.rs
```diff
@@ -179,6 +179,7 @@ impl MultisigTransaction {
             to: self.to.to_owned(),
             data_size: data_size(&self.data),
             value: self.value.as_ref().unwrap().into(),
+            method_name: self.data_decoded.as_ref().map(|it| it.method.clone()),
```
rmeissner, line 182: Maybe we should at some point define how we convert Strings into owned strings. in some places we use `.to_owned()` in some `.clone()` ... not really important and not sure about the difference. (`to_owned()` works with `&str` and `String` :D)

rmeissner (approved):

src/models/converters/transactions/mod.rs
jpalvarezl: yeah, I would read on the implications of either option, I just changed it `to_owned` to keep it consistent for now :)

PR #143, opened 2020-09-09
Closes #142 Null decimals in TokenInfo handled

### LINKED ISSUE #142: Certain ERC721 tranfers can't load TokenInfo
Incoming transfers fail to load `TokenInfo` in certain cases. Examples: 

PVT token: https://safe-transaction.staging.gnosisdev.com/api/v1/tokens/0x8979D84FF2c2B797dFEc02469d3a5322cBEf4b98/

Transaction (`/v1/safes/<address>/transfers/`): https://safe-transaction.staging.gnosisdev.com/api/v1/safes/0x1230B3d59858296A31053C1b8562Ecf89A2f888b/transfers/?transaction_hash=0x6b4ddfcf19320e1edaad5bcdef3da54f463ee5cb609ba4a1e2042fbff702e718

TX details returned: 
https://safe-client-rinkeby.staging.gnosisdev.com/v1/transactions/ethereum_0x1230B3d59858296A31053C1b8562Ecf89A2f888b_0x6b4ddfcf19320e1edaad5bcdef3da54f463ee5cb609ba4a1e2042fbff702e718_0xd698f24913be771f

jpalvarezl:
Closes #142
- Default on null value deserialiser added to JSON utils
- Any other default value required for a specific type can be add by implementing the `Default` trait for said type

### DIFF
--- a/src/providers/info.rs
+++ b/src/providers/info.rs
@@ -7,6 +7,7 @@ use serde::de::DeserializeOwned;
 use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
 use mockall::automock;
+use crate::utils::json::default_if_null;
 
 #[automock]
 pub trait InfoProvider {
@@ -44,6 +45,7 @@ pub struct TokenInfo {
     #[serde(rename = "type")]
     pub token_type: TokenType,
     pub address: String,
+    #[serde(deserialize_with = "default_if_null")]
     pub decimals: u64,
     pub symbol: String,
     pub name: String,
--- a/src/providers/info.rs
+++ b/src/providers/info.rs
@@ -1,13 +1,13 @@
 use crate::utils::cache::{Cache, CacheExt};
 use crate::config::{base_transaction_service_url, info_cache_duration};
 use crate::utils::context::Context;
+use crate::utils::json::default_if_null;
 use serde_json;
 use anyhow::Result;
 use serde::de::DeserializeOwned;
 use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
 use mockall::automock;
-use crate::utils::json::default_if_null;
 
 #[automock]
 pub trait InfoProvider {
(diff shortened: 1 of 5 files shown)


jpalvarezl:
テストを書いたほうがいいかな？

rmeissner:
はい

src/providers/info.rs
```diff
@@ -7,6 +7,7 @@ use serde::de::DeserializeOwned;
 use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
 use mockall::automock;
+use crate::utils::json::default_if_null;
```
rmeissner, line 10: Not that I really mind ... but do we have any convention on the order of imports? (I try to group the `crate` ones and the others .... theoretically you could even try to sort them like it is done in android studio formatting)

rmeissner (approved):
As mentioned

src/providers/info.rs
jpalvarezl: I think I added this one manually, so that's why it was left somewhere weird. I will move it up.

src/utils/tests/json.rs
```diff
@@ -0,0 +1,88 @@
...
+    fn default() -> Self {
+        WithCustomDefault("custom default".to_string())
+    }
+}
+
+#[test]
+#[should_panic]
+fn deserialize_missing_expected_field() {
+    let test_json = json!({
+        "other_field":"other value"
```
rmeissner, line 27: Ohhh interesting wasn't aware that it will crash if the key is not present

src/utils/tests/json.rs
```diff
@@ -0,0 +1,88 @@
+use crate::utils::json::default_if_null;
+use serde::Deserialize;
+
+
+#[derive(PartialEq, Deserialize, Debug)]
+struct ExpectedStruct {
+    #[serde(deserialize_with = "default_if_null")]
+    field: String,
```
rmeissner, line 8: Maybe u64 would have been more interesting for us, but either way this is nice :)

src/utils/tests/json.rs
jpalvarezl: Yeah, I figured I wanted to make sure we are not injecting any stuff :)

Issue #145, opened 2020-09-10
Disable logging for no matching request errors and health endpoint


jpalvarezl:
Logging no matching route errors makes our alert channel very noisy and contributes nothing to our logs. It should be disabled for this specifically.

Additionally, disable logging for `/health` so that we get cleaner logs in Kibana.

PR #146, opened 2020-09-10
Closes #127 Introduction of health endpoint

### LINKED ISSUE #127: Add liveness and readiness endpoints
To have good scaling and HA (that’s necessary so traffic can be balanced) we need to provide endpoints to check liveness and readiness. This also allows us to have updates without downtime in a "semiautomated" manner.

### Notes
- https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/
- https://www.openshift.com/blog/liveness-and-readiness-probes
- Concrete implementation needs to be checked
  - Quote by @denisgranha: `you can use any endpoint, it’s better having one with some logic for it as it gives more precise status, but about endpoint it’s enough to check the API it’s up and running`

jpalvarezl:
Closes #127 
- ~~Health endpoint caches the response from `/about` endpoint of the transaction service and returns it~~
- Health endpoint caches and responds with the value from the`("/health", "")` key value pair

### DIFF
--- a/src/routes/health.rs
+++ b/src/routes/health.rs
@@ -1,9 +1,10 @@
 use crate::utils::context::Context;
-use crate::config::request_cache_duration;
 use crate::utils::cache::CacheExt;
 use crate::utils::errors::ApiResult;
+use crate::config::request_cache_duration;
+use rocket::response::content;
 
 #[get("/health")]
-pub fn health(context: Context) -> ApiResult<String> {
-    context.cache().cache_string("/health", request_cache_duration())
+pub fn health(context: Context) -> ApiResult<content::Json<String>> {
+    context.cache().cache_resp("/health", request_cache_duration(), || Ok(String::new()))
 }
--- a/src/routes/health.rs
+++ b/src/routes/health.rs
@@ -1,0 +1,10 @@
+use crate::utils::context::Context;
+use crate::config::{about_cache_duration, base_transaction_service_url};
+use crate::utils::cache::CacheExt;
+use crate::utils::errors::ApiResult;
+
+#[get("/health")]
+pub fn health(context: Context) -> ApiResult<String> {
+    let url = format!("{}/about", base_transaction_service_url());
+    context.cache().request_cached(context.client(), &url, about_cache_duration())
+}
--- a/src/routes/health.rs
+++ b/src/routes/health.rs
@@ -1,10 +1,9 @@
 use crate::utils::context::Context;
-use crate::config::{about_cache_duration, base_transaction_service_url};
+use crate::config::request_cache_duration;
 use crate::utils::cache::CacheExt;
 use crate::utils::errors::ApiResult;
 
 #[get("/health")]
 pub fn health(context: Context) -> ApiResult<String> {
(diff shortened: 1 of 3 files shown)


jpalvarezl:
One open question would be, if the transaction service returns for some reason an error here we would forward it and then I am not sure if that's good or bad in this situation. As we could still serve anything in the cache, but that would be stale and maybe not a great idea either?

src/routes/health.rs
```diff
@@ -0,0 +1,10 @@
+use crate::utils::context::Context;
+use crate::config::{about_cache_duration, base_transaction_service_url};
+use crate::utils::cache::CacheExt;
+use crate::utils::errors::ApiResult;
+
+#[get("/health")]
+pub fn health(context: Context) -> ApiResult<String> {
+    let url = format!("{}/about", base_transaction_service_url());
+    context.cache().request_cached(context.client(), &url, about_cache_duration())
```
rmeissner, line 9: I would not do this, as this would make the health of this service depend on the health of the transaction service ... which is wrong in the context of deployment (where we want to know if the service is ready to being used). To be fair the same argument can be made about the redis but I would put them on different levels ... maybe :P

I would really just return a empty response. (with or without caching)

cc @Uxio0 what is your take on this?

Uxio0: Totally agree @rmeissner. Basically this `health` endpoint is important so the services can be monitored. For example, if the service is not responding for some configured time it will get rebooted or another replica will be set up. It doesn't make sense that it depends on the transaction service `health`.

We use `about` endpoint on the transaction service as it's cached, easy to return and also it if it's not working is because there's a configuration problem.

> I would really just return a empty response. (with or without caching)
That should work, as long as it's true that if this endpoint is working the service is working

src/utils/cache.rs
```diff
@@ -85,6 +85,22 @@ pub trait CacheExt: Cache {
             }
         }
     }
+
+    fn cache_string(
```
rmeissner, line 89: This method is pretty meaningless, right? ... we will never need something like this again (it creates an empty string and caches it). Maybe it should at least take a function to create a meaninfull string (which could still be empty for the health case). And can't you use `cache_resp` with a `string`?

jpalvarezl: 🤦  don't know what I was thinking with that method.

rmeissner (approved):

rmeissner:
Can you add the endpoint to the deployment doc and let devops know that we made changes (not sure if something else needs to be adjusted, maybe @Uxio0 knows)

Uxio0:
> Can you add the endpoint to the deployment doc and let devops know that we made changes (not sure if something else needs to be adjusted, maybe @Uxio0 knows)

This should be enough, let devops know so deployment can be adjusted and that's all. I would recommend also removing all logging from that endpoint, as it will show in Kibana every few seconds

jpalvarezl:
I am not sure how to do the latter :D

I will update the doc and look into it!

jpalvarezl:
I have added the disabling of logs for `/health` to #145

PR #149, opened 2020-09-18
Closes #92 Refactored MultiSig Transaction classification

### LINKED ISSUE #92: Refactor Multisig Transaction classification logic
Multisig Transaction classification is a bit confusing and could be refactored. Specifically, there is a `token_info` request on every classification, which doesn't make sense in 60% of the cases (in the unit tests we just mock it as an error reponse). These cases are `Transfer::Ether`, `SettingsChange` and `Custom` transaction.

The solution should prevent unnecessary requests to the `InfoProvider` even though we have a cache, we are not able to distinguish an error that matters from an error that happens when the request doesn't make sense given the context.

Switching the order of checks so that we only request token information when we strictly need it. However, the order in which we check could matter. First, we should cover with unit tests before commencing this refactor.

jpalvarezl:
Closes #92 

- Cleaned up methods no longer used ~~(got rid entirely of `transfer_type` tests)~~
- Implemented the heuristic described in the issue
- Adjusted mocks to fit the fewer calls to `InfoProvider` for `TokenInfo`

- Added tests for transfer and settings method names `crate::utils::mod` methods
- Added tests for `check_sender_or_receiver`
- Not sure what else, this got a bit out of hand


src/models/converters/transactions/tests/mod.rs
```diff
@@ -1,4 +1,45 @@
+use crate::models::converters::transactions::data_size;
+
 mod data_size_calculation;
 mod summary;
 mod details;
-mod transfer_type_checks;
```
rmeissner, line 4: Aren't these the same tests as above?

Also not sure if I would add tests the mod.rs

src/models/converters/transactions/mod.rs
```diff
@@ -45,72 +45,34 @@ impl MultisigTransaction {
...
+        } else if self.value_as_uint() > 0 && data_size(&self.data) == 0 {
             TransactionInfo::Transfer(self.to_ether_transfer())
+        } else if self.value_as_uint() == 0 && data_size(&self.data) > 0 && self.safe == self.to {
+            TransactionInfo::SettingsChange(self.to_settings_change())
+        } else if self.data_decoded.as_ref().map(|it| it.is_erc20_transfer_method() || it.is_erc721_transfer_method()).unwrap_or(false) {
+            if !check_sender_or_receiver(&self.data_decoded, &self.safe) {
+                TransactionInfo::Custom(self.to_custom())
+            } else {
+                if let Ok(token) = info_provider.token_info(&self.to) {
+                    match token.token_type {
```
rmeissner, line 59: Before we also checked, that the token type of the loaded token and the token transfer are the same (this was for some edge cases where weird things happen ... I guess)

src/models/converters/transactions/tests/summary.rs
```diff
@@ -11,12 +11,12 @@ use crate::models::commons::ParamValue::SingleValue;
 
 #[test]
 fn data_size_calculation() {
-    assert_eq!(data_size(&None), "0");
```
rmeissner, line 14: How often do we test this method?

src/models/converters/transactions/tests/transfer_type_checks.rs
```diff
@@ -1,213 +0,0 @@
-use crate::models::backend::transactions::{MultisigTransaction, Confirmation};
```
rmeissner, line 1: Do we have similar tests for the new logic (with the same input)?

src/models/converters/transactions/mod.rs
```diff
@@ -45,72 +45,34 @@ impl MultisigTransaction {
...
-        if self.is_erc20_transfer(&token) {
-            TransactionInfo::Transfer(self.to_erc20_transfer(&token.as_ref().unwrap()))
-        } else if self.is_erc721_transfer(&token) {
-            TransactionInfo::Transfer(self.to_erc721_transfer(&token.as_ref().unwrap()))
-        } else if self.is_ether_transfer() {
+        if (self.value_as_uint() > 0 && data_size(&self.data) > 0) || !self.operation.contains(&Operation::CALL) {
+            TransactionInfo::Custom(self.to_custom())
+        } else if self.value_as_uint() > 0 && data_size(&self.data) == 0 {
             TransactionInfo::Transfer(self.to_ether_transfer())
+        } else if self.value_as_uint() == 0 && data_size(&self.data) > 0 && self.safe == self.to {
```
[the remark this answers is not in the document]
jpalvarezl: So far this is just the pseudo-code from the issue written in rust. We should probably still check for `SettingsChange` method names in `data_decoded` as we used to.

src/models/converters/transactions/tests/summary.rs
jpalvarezl: I think too often. But it is usually tested indirectly. I wanted to have something where we asserted against `None`, `0`, `0x` and happy path. It could be a little too paranoid.

rmeissner: We test this directly at least 3 times (visible in this pr)

jpalvarezl: You are right, I removed the redundant ones.

src/models/converters/transactions/mod.rs
jpalvarezl: Right now, that check is happening in the `if` whether we check for erc20 or erc721 transfer methods. I should mix it? Only when we identify the method we proceed to fetch the `tokenInfo` and match the `tokenType`

jpalvarezl:
```rust
// ...
} else if self.data_decoded.as_ref().map(|data_decoded|
            data_decoded.is_erc20_transfer_method() || data_decoded.is_erc721_transfer_method()).unwrap_or(false) &&
            check_sender_or_receiver(&self.data_decoded, &self.safe) {
            info_provider.token_info(&self.to)
                .map_or(TransactionInfo::Custom(self.to_custom()),
                        |token|
                            match token.token_type {
                                TokenType::Erc20 => TransactionInfo::Transfer(self.to_erc20_transfer(&token)),
                                TokenType::Erc721 => TransactionInfo::Transfer(self.to_erc721_transfer(&token)),
                                _ => TransactionInfo::Custom(self.to_custom()),
                            },
                )
        } else {
// ...
```

rmeissner: nvm ... not sure what I wanted to point at (maybe I was thinking about transfers in ethereum transaction objects). I think this should be fine ... there might be some edge cases, but until liliya finds them I would not optimize for that :P

src/models/converters/transactions/tests/mod.rs
jpalvarezl: I removed the tests from mod. `transfer_type_checks` were the methods checking `is_settings_change` , `is_ether_transfer`, `is_erc20_transfer` and `is_erc721_transfer` which in the current implementation don't exist. I will add more tests trying to cover the types, but yeah, there is also fewer methods performing the checks.

src/models/converters/transactions/tests/transfer_type_checks.rs
jpalvarezl: Not yet 😓  working on it

src/models/converters/transactions/mod.rs
```diff
@@ -45,72 +45,35 @@ impl MultisigTransaction {
     }
 
     fn transaction_info(&self, info_provider: &mut dyn InfoProvider) -> TransactionInfo {
-        if self.is_settings_change() {
-            // Early exit if it is a setting change
-            return TransactionInfo::SettingsChange(self.to_settings_change());
-        }
-        let token = info_provider.token_info(&self.to).ok();
```
jpalvarezl, line 52: It seems like we were always trying to fetch the `token_info` regardless whether it was available in the `MultisigTransaction`. Should we keep this behaviour? Or only fetch when there is missing info? @rmeissner

src/models/converters/transactions/tests/transfer_type_checks.rs
jpalvarezl: Just restored the methods and adjusted them to the new methods like we talked

src/models/converters/transactions/mod.rs
rmeissner: I would always load it .... it is simpler for now

src/models/converters/transactions/tests/check_sender_or_receiver.rs
```diff
@@ -0,0 +1,112 @@
+use crate::models::commons::{DataDecoded, Parameter};
+use crate::models::converters::transactions::check_sender_or_receiver;
+use crate::models::commons::ParamValue::SingleValue;
+
+#[test]
+fn check_sender_or_receiver_safe_sender() {
+    let data_decoded = Some(DataDecoded {
+        method: "transfer".to_string(),
```
rmeissner, line 8: Not sure what we test here, as for a `transfer` method the params are not important

src/models/converters/transactions/tests/check_sender_or_receiver.rs
```diff
@@ -0,0 +1,112 @@
...
+
+    let actual = check_sender_or_receiver(&data_decoded, "0x1230B3d59858296A31053C1b8562Ecf89A2f888b");
+    assert!(actual);
+}
+
+#[test]
+fn check_sender_or_receiver_data_decoded_none() {
+    let data_decoded: Option<DataDecoded> = Option::None;
+
+    check_sender_or_receiver(&data_decoded, "");
```
rmeissner, line 94: We don't assert anything here ,do we just expect it to now throw?

src/models/converters/transactions/tests/check_sender_or_receiver.rs
```diff
@@ -0,0 +1,112 @@
...
+                value_decoded: None,
+            }]),
+    });
+
+    let actual = check_sender_or_receiver(&data_decoded, "0x1230B3d59858296A31053C1b8562Ecf89A2f888b");
+    assert!(actual);
+}
+
+#[test]
+fn check_sender_or_receiver_safe_wrong_method() {
```
rmeissner, line 40: I would expect another test with the wrong method, but with the correct `to`

rmeissner (approved):
Wow sooo many tests :D 

I didn't look at all tests in detail yet, but looks good now, no unnecessary network calls :)

src/models/converters/transactions/tests/check_sender_or_receiver.rs
jpalvarezl: So this is just one of the 3 happy paths. I don't understand very well what this method is supposed to do, but it seems that it checks that data decoded contains a certain address value in a parameter that's either called `from` or `to` and also that the `method` field is `transfer`, any of these 3 is true then this method returns true, which from a business logic perspective, I don't understand exactly what that means.

src/models/converters/transactions/tests/check_sender_or_receiver.rs
jpalvarezl: no no, I messed the assert. This should return `false`

PR #152, opened 2020-09-23
Closes #65 Generation and Publishing of test coverage reports

### LINKED ISSUE #65: Integrate test coverage tool
- [x] Generate test coverage
- [x] Upload test coverage

### Open questions
- What service to use

### References
- https://github.com/gnosis/dex-services/blob/master/.travis.yml#L62
  - https://coveralls.io/github/gnosis/dex-services
  - Note from @e00E: I think there are some places where the coverage report wrongly reports lines or branches not being covered but overall it works if the goal isn't to get 100% coverage.
- https://github.com/gnosis/dex-services/issues/688

jpalvarezl:
Closes #65 
- Link to coveralls: https://coveralls.io/github/gnosis/safe-client-gateway

I will fork a new issue for adding the badge for master once this PR is merged as the current markdown generated by coveralls points to this feature branch. Also, we should bump up the version to `v1.2.0` or at least `v1.1.0` in the `Cargo.toml` file as it seems to be still stuck in `v0.2.0`

Edit: the aforementioned issue is #153

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,5 +40,5 @@ jobs:
       # message which made the build fail (probably travis related).
       - env CARGO_INCREMENTAL=0 RUSTFLAGS="-Zprofile -Ccodegen-units=1 -Copt-level=0 -Clink-dead-code -Coverflow-checks=off -Awarnings" cargo +nightly test
       # Ignore untested cargo files in travis root and auto-generated eth-contract code
-      - ./grcov --branch --ignore-not-existing --llvm --excl-start "test" --excl-line "#\[cfg_attr\(test, mockall::automock\)\]|#\[derive" --ignore "/*" --ignore "target/debug/build/**" target/debug/ --output-path coveralls.json --output-type "coveralls+" --source-dir . --service-name travis-ci --service-job-id $TRAVIS_JOB_ID
+      - ./grcov --branch --ignore-not-existing --llvm --ignore "*test*" --excl-line "#\[cfg_attr\(test, mockall::automock\)\]|#\[derive" --ignore "/*" --ignore "target/debug/build/**" target/debug/ --output-path coveralls.json --output-type "coveralls+" --source-dir . --service-name travis-ci --service-job-id $TRAVIS_JOB_ID
       - curl --form json_file=@coveralls.json https://coveralls.io/api/v1/jobs
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,7 @@ script:
   - cargo build --verbose --all --locked
   - cargo test --verbose --all --locked
   # Ignore untested cargo files in travis root and auto-generated eth-contract code
-  - ./grcov --branch --ignore-not-existing --llvm --excl-line "#\[derive|#\[serde|#\[automock|#\[database" --ignore "/*" --ignore "target/debug/build/**" target/debug/ --output-path coveralls.json --output-type "coveralls+" --source-dir . --service-name travis-ci --service-job-id $TRAVIS_JOB_ID
+  - ./grcov --branch --ignore-not-existing --llvm --ignore "/*" --ignore "target/debug/build/**" target/debug/ --output-path coveralls.json --output-type "coveralls+" --source-dir . --service-name travis-ci --service-job-id $TRAVIS_JOB_ID
   - curl --form json_file=@coveralls.json https://coveralls.io/api/v1/jobs
 
 deploy:
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,7 +28,7 @@ script:
   - cargo build --verbose --all --locked
   - cargo test --verbose --all --locked
   # Ignore untested cargo files in travis root and auto-generated eth-contract code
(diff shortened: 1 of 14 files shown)


.travis.yml
```diff
@@ -1,11 +1,11 @@
-language: rust
+anguage: rust
```
rmeissner, line 1: :thinking:

.travis.yml
```diff
@@ -1,11 +1,11 @@
...
 cache: cargo
 env:
   global:
-  - DOCKERHUB_ORG=gnosispm
-  - DOCKERHUB_PROJECT=safe-client-gateway
+    - DOCKERHUB_ORG=gnosispm
+    - DOCKERHUB_PROJECT=safe-client-gateway
 services:
   - docker
 script:
```
rmeissner, line 11: Looking at the logs of https://travis-ci.com/github/gnosis/safe-client-gateway/builds/186205255 I cannot find the logs for `cargo build --verbose --all --locked` and `cargo test --verbose --all --locked` are they still executed?

.travis.yml
```diff
@@ -24,4 +24,21 @@ deploy:
     script: bash scripts/deploy_docker.sh $TRAVIS_TAG
     on:
       tags: true
-      branch: main
\ No newline at end of file
+      branch: main
+
+jobs:
```
rmeissner, line 29: Currently I would include this in our normal build script (the `script` tag that already exists in the root of this file). So the `curl` and environment settings go before the `cargo build` and the `./grcov` goes after it.

We would only need multiple jobs if we want different compilers or os settings for different builds (which is not the case for us)

Alternatively we can remove the `script` tag from the root and inline the commands in the job `script`. The main change here is that we use `--verbose --all --locked` for `cargo test` (and `cargo build`)

.travis.yml
```diff
@@ -24,4 +24,21 @@ deploy:
...
+    script:
+      - curl --location https://github.com/mozilla/grcov/releases/latest/download/grcov-linux-x86_64.tar.bz2 | tar jxf -
+      # These flags are recommended by https://github.com/mozilla/grcov#grcov-with-travis .
+      # vk: I had to remove `-Zpanic_abort_tests -Cpanic=abort` because this would cause compilation
+      # to fail but I haven't investigated more into why.
+      # I added `-Awarnings` which allows all warnings. This was necessary because nightly can
+      # introduce some new warnings and when testing one such warning appeared in the ethcontract
+      # generated contract code which caused the whole file to be printed as part of the warning
+      # message which made the build fail (probably travis related).
+      - env CARGO_INCREMENTAL=0 RUSTFLAGS="-Zprofile -Ccodegen-units=1 -Copt-level=0 -Clink-dead-code -Coverflow-checks=off -Awarnings" cargo +nightly test
```
rmeissner, line 41: I would prefer if we multi-line this: each env var becomes a `export XYZ=foo` and the execution of the tests would also be on a separate line (see  https://github.com/mozilla/grcov#grcov-with-travis)

jpalvarezl:
I was just how it was shown in the link in your other comment:
>I would prefer if we multi-line this: each env var becomes a export XYZ=foo and the execution of the tests would also be on a separate line (see https://github.com/mozilla/grcov#grcov-with-travis)

.travis.yml
jpalvarezl: They appear to be executed now. I still don't get why we get 24% coverage reported when it also shows, 78% for `/src` . I am wondering if this could be an issue with the `deploy` step that runs the script with the line `RUN cargo build --release --locked` a couple more times from the `Dockerfile`

.travis.yml
```diff
@@ -1,16 +1,34 @@
 language: rust
+
+before-install:
+  - curl --location https://github.com/mozilla/grcov/releases/latest/download/grcov-linux-x86_64.tar.bz2 | tar jxf -
+
 rust:
   - nightly
-cache: cargo
```
rmeissner, line 4: Why are we removing this? (This potentially makes the build quite a lot slower)

.travis.yml
```diff
@@ -1,16 +1,34 @@
 language: rust
+
+before-install:
+  - curl --location https://github.com/mozilla/grcov/releases/latest/download/grcov-linux-x86_64.tar.bz2 | tar jxf -
```
rmeissner, line 4: I am still not sure why this is in `before-install`

.travis.yml
jpalvarezl: Nick recommended it for the subgraph project sometime ago, I forgot the exact reason. I will put it back in. I think it has no effect in this case.

.travis.yml
jpalvarezl: I will move this to be a step in the script

jpalvarezl: I think that was actually somehow messing up what percentage was reported as the full coverage. 🤔  no idea to be honest, glad that now we get a % that makes sense.

rmeissner: yeah, sorry didn't see that they did that in the grcov example there ... not sure why they would do that 🤔 but yeah if it works now :P

rmeissner (approved):

PR #158, opened 2020-10-01
Closes # 133 Eur and Usd fiat balances

### LINKED ISSUE #133: Add balances endpoint with support for more fiat than USD for balances
- fetch usd prices from safe-transaction service 
- use https://exchangeratesapi.io/ to fetch other exchange rates
- combine to return requested exchange rate

jpalvarezl:
Closes #133
- Added endpoint `/v1/safes/<address>/balances/<fiat>/`
- Introduced new `struct` for both backend and service
- Introduced method for fetching exchanged rated from the [ECB](https://exchangeratesapi.io/)

Pending
- ~~Better implementation for exchange rate fetching in the `DefaultInfoProvider`~~
- ~~Tests for converters~~
- ~~Maybe it is time to add tests for the `DefaultInfoProvider`~~ #159

### DIFF
--- a/src/json/balances/balance_compound_ether.json
+++ b/src/json/balances/balance_compound_ether.json
@@ -1,0 +1,15 @@
+{
+  "tokenAddress": "0xd6801a1DfFCd0a410336Ef88DeF4320D6DF1883e",
+  "token": {
+    "name": "Compound Ether 📈",
+    "symbol": "cETH",
+    "decimals": 8,
+    "logoUri": "https://gnosis-safe-token-logos.s3.amazonaws.com/0xd6801a1DfFCd0a410336Ef88DeF4320D6DF1883e.png"
+  },
+  "balance": "5002",
+  "balanceUsd": "0.0014",
+  "usdConversion": "28.5462",
+  "fiatBalance": "0.0014",
+  "fiatConversion": "28.5462",
+  "fiatCode": "USD"
+}
--- a/src/models/backend/balances.rs
+++ b/src/models/backend/balances.rs
@@ -6,8 +6,8 @@ pub struct Balance {
     pub token_address: Option<String>,
     pub token: Option<BalanceToken>,
     pub balance: String,
-    pub balance_usd: String,
-    pub usd_conversion: String,
+    pub fiat_balance: String,
+    pub fiat_conversion: String,
 }
 
 #[derive(Deserialize, Debug, Hash)]
--- a/src/models/backend/balances.rs
+++ b/src/models/backend/balances.rs
@@ -1,0 +1,21 @@
+use crate::providers::info::TokenInfo;
+use serde::Deserialize;
+
+#[derive(Deserialize, Debug, Hash)]
+#[serde(rename_all = "camelCase")]
+pub struct Balance {
+    pub token_address: String,
+    pub token: BalanceToken,
+    pub balance: String,
+    pub balance_usd: String,
+    pub usd_conversion: String,
+}
+
+#[derive(Deserialize, Debug, Hash)]
+#[serde(rename_all = "camelCase")]
+pub struct BalanceToken {
+    pub name: String,
+    pub symbol: String,
+    pub decimals: u64,
+    pub logo_uri: String,
+}
--- a/src/models/backend/balances.rs
+++ b/src/models/backend/balances.rs
@@ -1,11 +1,10 @@
-use crate::providers::info::TokenInfo;
 use serde::Deserialize;
 
 #[derive(Deserialize, Debug, Hash)]
 #[serde(rename_all = "camelCase")]
 pub struct Balance {
-    pub token_address: String,
-    pub token: BalanceToken,
+    pub token_address: Option<String>,
+    pub token: Option<BalanceToken>,
     pub balance: String,
     pub balance_usd: String,
     pub usd_conversion: String,
--- a/src/models/converters/balances.rs
+++ b/src/models/converters/balances.rs
@@ -1,0 +1,21 @@
+use crate::utils::errors::ApiResult;
+use crate::models::service::balances::Balance;
+use crate::models::backend::balances::Balance as BalanceDto;
+use crate::providers::info::{TokenInfo, TokenType};
+
+impl BalanceDto {
+    pub fn to_balance(&self, conversion_rate: u64) -> ApiResult<Balance> {
+        Ok(Balance {
+            token_info: TokenInfo {
+                token_type: TokenType::Erc20,
+                address: self.token_address.to_owned(),
+                decimals: self.token.decimals,
+                symbol: self.token.symbol.to_owned(),
+                name: self.token.name.to_owned(),
+                logo_uri: Some(self.token.logo_uri.to_owned()),
+            },
+            amount: self.balance.to_owned(),
+            fiat_amount: self.balance_usd.to_string(),
+        })
+    }
+}
--- a/src/models/converters/balances.rs
+++ b/src/models/converters/balances.rs
@@ -1,21 +1,22 @@
-use crate::utils::errors::ApiResult;
 use crate::models::service::balances::Balance;
 use crate::models::backend::balances::Balance as BalanceDto;
 use crate::providers::info::{TokenInfo, TokenType};
 
 impl BalanceDto {
-    pub fn to_balance(&self, conversion_rate: u64) -> ApiResult<Balance> {
-        Ok(Balance {
+    pub fn to_balance(&self, conversion_rate: &f64) -> Balance {
+        let fiat_amount = self.balance_usd.parse::<f64>().unwrap_or(0.0) * conversion_rate;
+        let token_type = self.token_address.as_ref().map(|_| TokenType::Erc20).unwrap_or(TokenType::Ether);
+        Balance {
             token_info: TokenInfo {
-                token_type: TokenType::Erc20,
-                address: self.token_address.to_owned(),
-                decimals: self.token.decimals,
-                symbol: self.token.symbol.to_owned(),
-                name: self.token.name.to_owned(),
-                logo_uri: Some(self.token.logo_uri.to_owned()),
+                token_type,
+                address: self.token_address.to_owned().unwrap_or(String::from("0x0000000000000000000000000000000000000000")),
+                decimals: self.token.as_ref().map(|it| it.decimals).unwrap_or(0),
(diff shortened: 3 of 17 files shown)


src/providers/info.rs
```diff
@@ -68,6 +75,16 @@ impl InfoProvider for DefaultInfoProvider<'_> {
     }
 }
 
+impl DefaultInfoProvider<'_> {
+    pub fn exchange_usd_to_eur(&self) -> Result<f64> {
+        let url = String::from("https://api.exchangeratesapi.io/latest?base=USD&symbols=EUR");
```
rmeissner, line 80: As mentioned above I would not limit this to EUR ... I am not sure where this requirement that we only support `EUR` is comming from. We want to support all currencies that we have an exchange rate for (see how it is currently in the web interface)

src/routes/balances.rs
```diff
@@ -0,0 +1,18 @@
...
+use crate::config::request_cache_duration;
+use rocket::response::content;
+use crate::utils::cache::CacheExt;
+use crate::services::balances::*;
+use crate::models::service::balances::*;
+use crate::providers::info::DefaultInfoProvider;
+
+#[get("/v1/safes/<safe_address>/balances/<fiat>")]
+pub fn get_balances(context: Context, safe_address: String, fiat: String) -> ApiResult<content::Json<String>> {
+    let info_provider = DefaultInfoProvider::new(&context);
```
rmeissner, line 12: In our transaction endpoint we build the info provider and so on in the `services` module. No strong opinion on that as for some endpoints without much logic we also do it in the route and it should not make any difference ... I think :P )

src/models/service/balances.rs
```diff
@@ -0,0 +1,26 @@
...
+    pub token_info: TokenInfo,
+    pub amount: String,
+    pub fiat_amount: String,
+}
+
+impl Fiat {
+    pub fn from_dasd(fiat: &str, info_provider: &DefaultInfoProvider) -> Self {
+        match fiat {
+            "usd" => Fiat::Usd(1.0),
+            "eur" => Fiat::Eur(info_provider.exchange_usd_to_eur().unwrap_or(1.0)),
```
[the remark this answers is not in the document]
jpalvarezl: Yeah, I was not sure about the spec, I thought it was just Euro and Usd. That's why I decided to wrap stuff in our own enum in case someday we decided to change exchange rate values provider. I will just pass on the currency code then, it should simplify things.

src/providers/info.rs
jpalvarezl: From nowhere. I think I just misunderstood the ticket and assumed we just wanted that.

src/routes/balances.rs
jpalvarezl: It might be easier to test though 🤔  in this case we provide the exchange rate instead of having to use the mock...

jpalvarezl: Not sure what's best

rmeissner: If we go from the test perspective you could say more code is tested when you do it in the services (as we do not test the routes module, right?) :D

jpalvarezl: well yeaaaah, but a counter argument could be made that mocked behaviour is synthetic and therefore having a more minimalist interface is more reliable. As in, passing the single float value we use instead of a full InfoProvider :)

rmeissner: But then you still need to test how this single float value is calculated ... this is basically just pushing responsibilities. So I would agree with you if we test both methods, but we don't currently. Also the mocked behaviour is internally synthetic, not to the code calling it.

jpalvarezl: Ah good point, If we'd eventually add tests for routes then yeah, we are just pushing that upwards. For the sake of consistency then I don't really mind pushing that down when I finally get to testing.

src/providers/info.rs
```diff
@@ -68,6 +76,21 @@ impl InfoProvider for DefaultInfoProvider<'_> {
     }
 }
 
+impl DefaultInfoProvider<'_> {
+    pub fn exchange_usd_to(&self, currency_code: &str) -> Result<f64> {
+        let currency_code = currency_code.to_uppercase();
+        let url = format!("https://api.exchangeratesapi.io/latest?base=USD&symbols={}", &currency_code);
```
rmeissner, line 82: Loading all exchange rates at once (e.g. no symbol param in the query to the exchange rate api) will allow a better caching

src/models/converters/tests/mod.rs
```diff
@@ -3,4 +3,5 @@ mod get_transfer_direction;
 mod transfers;
 mod transfer_erc20;
 mod transfer_erc721;
-mod transfer_ether;
\ No newline at end of file
+mod transfer_ether;
+pub (super) mod balances;
```
rmeissner, line 7: when do we use `pub (super)`?

src/providers/info.rs
```diff
@@ -68,6 +76,20 @@ impl InfoProvider for DefaultInfoProvider<'_> {
     }
 }
 
+impl DefaultInfoProvider<'_> {
+    pub fn exchange_usd_to(&self, currency_code: &str) -> Result<f64> {
+        let currency_code = currency_code.to_uppercase();
+        let url = format!("https://api.exchangeratesapi.io/latest?base=USD");
+        debug!("exchange url: {}", &url);
```
rmeissner, line 83: can we remove this?

src/json/balances/balance_compound_ether.json
```diff
@@ -0,0 +1,15 @@
+{
+  "tokenAddress": "0xd6801a1DfFCd0a410336Ef88DeF4320D6DF1883e",
+  "token": {
+    "name": "Compound Ether 📈",
+    "symbol": "cETH",
+    "decimals": 8,
+    "logoUri": "https://gnosis-safe-token-logos.s3.amazonaws.com/0xd6801a1DfFCd0a410336Ef88DeF4320D6DF1883e.png"
+  },
+  "balance": "5002",
```
rmeissner, line 9: while they are correct, I would remove the `balanceUsd` and `usdConversion` in favor of the `fiatX` methods and also use them in the backend model

src/models/backend/balances.rs
```diff
@@ -0,0 +1,20 @@
+use serde::Deserialize;
+
+#[derive(Deserialize, Debug, Hash)]
+#[serde(rename_all = "camelCase")]
+pub struct Balance {
+    pub token_address: Option<String>,
+    pub token: Option<BalanceToken>,
+    pub balance: String,
+    pub balance_usd: String,
```
rmeissner, line 9: Replace with `fiat_balance`

See above

src/models/backend/balances.rs
```diff
@@ -0,0 +1,20 @@
...
+    pub usd_conversion: String,
```
rmeissner, line 10: replace with `fiat_conversion`

See above

src/models/service/balances.rs
```diff
@@ -0,0 +1,10 @@
+use crate::providers::info::TokenInfo;
+use serde::Serialize;
+
+#[derive(Serialize, Debug, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub struct Balance {
+    pub token_info: TokenInfo,
+    pub amount: String,
```
rmeissner, line 8: No strong preference, but would it make sense to use `balance` instead of `amount` to be in line with the transaction service? 

(same for `fiat_amount` vs `fiat_balance`)

src/models/service/balances.rs
```diff
@@ -0,0 +1,10 @@
+use crate::providers::info::TokenInfo;
+use serde::Serialize;
+
+#[derive(Serialize, Debug, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub struct Balance {
```
rmeissner, line 6: Afaik the web interface requires the `fiat_conversion` field

rmeissner (changes requested):

src/models/converters/tests/mod.rs
jpalvarezl: For tests in general I would say, like you did for `missing_signers` I think it allows you to have visibility as if the tests were at the same crate the method you are testing (as long as you are only 1 crate deep, which I think is the case for tests)

src/providers/info.rs
jpalvarezl: Sure

src/models/converters/balances.rs
```diff
@@ -4,7 +4,7 @@ use crate::providers::info::{TokenInfo, TokenType};
 
 impl BalanceDto {
     pub fn to_balance(&self, usd_to_fiat: f64) -> Balance {
-        let fiat_amount = self.balance_usd.parse::<f64>().unwrap_or(0.0) * usd_to_fiat;
+        let fiat_amount = self.fiat_balance.parse::<f64>().unwrap_or(0.0) * usd_to_fiat;
```
rmeissner, line 8: the var could be reanmed too :P

rmeissner (approved):

Agupane:
Hey @jpalvarezl ! just have a question, as far I understand we should be able to ask the balances in another currency than USD right?, in that case I think something is missing because I cannot fetch the balances in EUR or another fiat different than USD:

Example:
`https://safe-transaction.staging.gnosisdev.com/api/v1/safes/0x9913B9180C20C6b0F21B6480c84422F6ebc4B808/balances/eur/?exclude_spam=true&limit=3000`

Let me know if something else should be done on my side :) thanks

jpalvarezl:
Hi @Agupane, I think you are mixing services there. This is a gateway that maps what the transaction service returns. You can find the repo for the transaction service API here: https://github.com/gnosis/safe-transaction-service

If you want the balances for the safe you mention in your example, you can find it here: https://safe-client-rinkeby.staging.gnosisdev.com/v1/safes/0x9913B9180C20C6b0F21B6480c84422F6ebc4B808/balances/eur/

The fiat part of the path is dynamic, just see that the fiat is returned by https://exchangeratesapi.io/ (This is what we use for exchange references from USD, to whatever comes in the request)

We haven't yet implemented pagination or a filter for spam tokens, just FYI.

jpalvarezl:
Also, you can expect the spam filter implemented within the next 2 weeks: https://github.com/gnosis/safe-client-gateway/issues/163 as it is part of our current milestone.

You don't need to worry about setting a limit as this endpoint, as well as the one in the transaction service API, is not paginated.

Agupane:
Perfect thanks for the clarification! no problem about the pagination by now, we are still working on it too :)

PR #160, opened 2020-10-08
Perf logs 

### LINKED ISSUE #147: Add performance logs
We want to be able to understand the usage of the client gateway, therefore we want to be able to measure how many requests are performed against the client gateway and how long the response time is.

For this we can use kibana with pre-processing.

jpalvarezl:
Closes #147 

- Using request's `local_cache` to store the timestamp of when the request was received

### DIFF
--- a/src/monitoring/request_timer.rs
+++ b/src/monitoring/request_timer.rs
@@ -21,10 +21,10 @@ impl Fairing for RequestTimer {
     }
 
     fn on_response(&self, request: &Request, _response: &mut Response) {
-        let h = request.headers().get("requested_timestamp").next().unwrap().parse::<i64>().unwrap();
-        let instant = Utc::now().timestamp_millis();
-
-        let delta = instant - h;
-        log::debug!("For endpoint {} the request processing duration was {} [ms]", request.uri().path(), delta)
+        if let Some(requested_timestamp) = request.headers().get("requested_timestamp").next().map(|it| it.parse::<i64>().ok()).flatten() {
+            let instant = Utc::now().timestamp_millis();
+            let delta = instant - requested_timestamp;
+            log::debug!("For endpoint {} the request processing duration was {} [ms]", request.uri().path(), delta)
+        }
     }
 }
--- a/src/monitoring/request_timer.rs
+++ b/src/monitoring/request_timer.rs
@@ -1,7 +1,6 @@
 use rocket::fairing::{Fairing, Info, Kind};
 use chrono::Utc;
 use rocket::{Request, Response, Data};
-use rocket::http::Header;
 
 pub struct RequestTimer();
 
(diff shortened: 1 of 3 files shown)


src/monitoring/request_timer.rs
```diff
@@ -0,0 +1,30 @@
...
+        let h = Header::new("requested_timestamp", instant.to_string());
+        request.add_header(h)
+    }
+
+    fn on_response(&self, request: &Request, _response: &mut Response) {
+        let h = request.headers().get("requested_timestamp").next().unwrap().parse::<i64>().unwrap();
+        let instant = Utc::now().timestamp_millis();
+
+        let delta = instant - h;
+        log::debug!("For endpoint {} the request processing duration was {} [ms]", request.uri().path(), delta)
```
rmeissner: I would probably use `info` and not debug (but that is only important for the final version)

src/monitoring/request_timer.rs
```diff
@@ -0,0 +1,30 @@
...
+        Info {
+            name: "RequestTimer",
+            kind: Kind::Request | Kind::Response,
+        }
+    }
+
+    fn on_request(&self, request: &mut Request, _data: &Data) {
+        let instant = Utc::now().timestamp_millis();
+        log::debug!("on_request: {}", &instant);
+        let h = Header::new("requested_timestamp", instant.to_string());
```
rmeissner, line 19: Hmmm do we have a better alternative than the header? (it is fine if everything else is way more complicated)

Some notes:
- I would make the header key a constant
- Maybe look at `local_cache` from the `Request` struct

src/monitoring/request_timer.rs
jpalvarezl: @rmeissner is there a json structure and/or endpoint I need to submit this data for kibana?

src/monitoring/request_timer.rs
```diff
@@ -0,0 +1,28 @@
...
+    fn info(&self) -> Info {
+        Info {
+            name: "RequestTimer",
+            kind: Kind::Request | Kind::Response,
+        }
+    }
+
+    fn on_request(&self, request: &mut Request, _data: &Data) {
+        let instant = Utc::now();
+        log::info!("on_request: {}", &instant.to_rfc3339());
```
rmeissner, line 17: sorry I was not clear here ... I would remove the `on_request` and `on_response` for the final version (or make them `debug`) and only log the final information to `info`.

src/monitoring/request_timer.rs
rmeissner: @Uxio0 needs to help here :D I have no idea ... so this log should already show up but we want to have a format that it is easy to define an Regex (?!?) that can extract the information, so that kibana can handle them and we can build dashboards with the. .... I just have no clue how this works :P

Uxio0: Usually we do it using a regex configured in fluentd (the service that's gathering all the logs and providing to Elastic Search), I think that there's the option of providing the JSON to fluentd, but devops are not using it at all as far as I know

rmeissner (approved):

PR #167, opened 2020-10-14
Submit confirmation through gateway

### LINKED ISSUE #162: Submit transaction confirmation
Currently the clients have to submit confirmations to the transaction service API. 

When the apps are in the transaction details, they re-fetch the transactions details after submission (from the gateway) in order to obtain the new details with the new state. 

However, the web hook triggered cache invalidation comes too late; therefore there is a small time window in which the apps re-fetch the same stale details.

The gateway should provide an endpoint that can forward the request for the confirmation to the transaction service API. This should invalidate the cache for the transaction the same way that the web hook does. 

The endpoint should be `POST -> /transactions/<tx_id>/confirmations` and the `safeTxHash` should be passed as part of the body of the request.

**TBD**
Returned object by the gateway. Maybe it is not necessary to supply the full Tx details as a response.
Update:
Full object of transaction details will be returned

jpalvarezl:
Closes #162 

Changes:
- ~~Defined `ConfirmationRequest` for backend and service~~
- Using new endpoint requiring only the `safeTxHash` and the `signedSafeTxHash` in the body
- Clears cache on **successful** request, re-fetches details, and supplies updated details to clients
- Forwarding back and forth of request/response/error objects

~~NOTE: still not tested against a real safe~~


src/services/transactions_details.rs
```diff
@@ -99,36 +98,56 @@ pub fn get_transactions_details(
...
+            transaction_hash,
+            details_hash
+        } => get_module_transaction_details(context, &safe_address, &transaction_hash, &details_hash),
+        TransactionIdParts::Multisig { safe_tx_hash, .. } => get_multisig_transaction_details(context, &safe_tx_hash),
+        TransactionIdParts::TransactionHash(safe_tx_hash) => get_multisig_transaction_details(context, &safe_tx_hash),
+        _ => Err(ApiError::new_from_message(String::from("Bad transaction id"))),
+    }
+}
+
+pub(super) fn parse_id(details_id: &str) -> Result<TransactionIdParts> {
```
rmeissner, line 120: Maybe that should be moved to a more general module, as this might be used in multiple places (e.g. submit confirmations, right?

src/services/transactions_details.rs
```diff
@@ -99,36 +98,56 @@ pub fn get_transactions_details(
...
+            details_hash
+        } => get_ethereum_transaction_details(context, &safe_address, &transaction_hash, &details_hash),
+        TransactionIdParts::Module {
+            safe_address,
+            transaction_hash,
+            details_hash
+        } => get_module_transaction_details(context, &safe_address, &transaction_hash, &details_hash),
+        TransactionIdParts::Multisig { safe_tx_hash, .. } => get_multisig_transaction_details(context, &safe_tx_hash),
+        TransactionIdParts::TransactionHash(safe_tx_hash) => get_multisig_transaction_details(context, &safe_tx_hash),
+        _ => Err(ApiError::new_from_message(String::from("Bad transaction id"))),
```
rmeissner, line 116: The only type missing in this match is creation, right? Should we be more explicit in the error? E.g. "Creation transactions are not supported for transaction details"

src/services/tx_confirmation.rs
```diff
@@ -0,0 +1,27 @@
...
+        context.cache().invalidate_pattern(&format!("*{}*", &safe_tx_hash));
+
+        let url = format!("{}/v1/safes/{}/multisig-transactions/", base_transaction_service_url(), &safe_address);
+        context.client().post(&url)
+            .body(serde_json::to_string::<BackendConfirmationRequest>(&backend_confirmation_request)?)
+            .send()?;
+
+        get_multisig_transaction_details(context, &safe_tx_hash)
+    } else {
+        Err(ApiError::new_from_message(String::from("Bad id for confirmation submission")))
```
rmeissner, line 25: I would also here be more concrete in the error message: "Can only submit confirmations for multisig transactions"

rmeissner: Also this should not be a 500, but a 400 status code

src/services/transactions_details.rs
rmeissner: Here this should probably be a `400` status code, as it is an invalud argument. Or a `422` as we could not process the id.

src/services/tests/mod.rs
```diff
@@ -1 +1,2 @@
-mod invalidate_caches;
\ No newline at end of file
+mod invalidate_caches;
```
rmeissner, line 1: we are quite inconssitent on our module visibility :/

src/routes/transactions.rs
```diff
@@ -19,3 +22,10 @@ pub fn details(context: Context, details_id: String) -> ApiResult<content::Json<
         transactions_details::get_transactions_details(&context, &details_id)
     })
 }
+
+#[post("/v1/transactions/<tx_id>/confirmations", format = "json", data = "<confirmation_request>")]
+pub fn submit_confirmation(context: Context, tx_id: String, confirmation_request: Json<ConfirmationRequest>) -> ApiResult<content::Json<String>> {
+    context.cache().cache_resp(&context.uri(), request_cache_duration(), || {
```
rmeissner, line 28: post endpoints should probably not be cached ;)

src/services/tx_confirmation.rs
jpalvarezl: This is outdated. Now the error from the tx service is forwareded.

src/services/tests/mod.rs
jpalvarezl: I think that since we are testing "java-like" but the rust way is a but more clumped up this could maybe worthy of a larger scale refactor/code re-org?

rmeissner: The inconsistency in module visibility has nothing to do with tests. We include similar modules sometimes with `super` visibility and sometimes without visibility (in many places, not just tests). This is probably due to auto imports in the ide ... Would still expect that this is consistent.

rmeissner: Regarding test structure refactoring, I see no reason why we should do this.

rmeissner: I think this comment is outdated I will resolve it

jpalvarezl:
WTH happened to coverage O.O

src/services/transactions_details.rs
jpalvarezl: In the end I just ask for the `safeTxHash` as it is a more "common" way between the backend and the gateway to identify transactions. I was thinking that if we ever want to implement confirmation through notifications (not saying it's a good idea, just fantasising here :) ) It would be better to keep things flexible.

src/services/tx_confirmation.rs
```diff
@@ -0,0 +1,37 @@
...
+use crate::utils::context::Context;
+use crate::utils::errors::{ApiError, ApiResult};
+use std::collections::HashMap;
+
+pub fn submit_confirmation(
+    context: &Context,
+    safe_tx_hash: &str,
+    signature: &str,
+) -> ApiResult<()> {
+    context
```
rmeissner, line 12: I would only invalidate the cache after a successful post

src/services/tx_confirmation.rs
```diff
@@ -0,0 +1,37 @@
...
+    let mut json = HashMap::new();
+    json.insert("signature", signature);
+
+    let result = context.client().post(&url).json(&json).send()?;
+
+    if result.status().is_success() {
+        Ok(())
+    } else {
+        Err(ApiError::from_backend_error(
+            result.status().as_u16(),
```
rmeissner, line 30: Would it be helpful to add a helper to `ApiError` that takes the result?

rmeissner (approved):
Minor comments, but functionality should be good to go

src/services/tx_confirmation.rs
jpalvarezl: Ah, that's not really a `Result` but a `Response` from reqwest. The var name is a bit confusing. But I think it's a good idea to have a default mapper like that

PR #170, opened 2020-10-21
"trusted" and "exclude_spam" filters added to balance endpoint

### LINKED ISSUE #163: Allow to use spam and trusted filter in balance endpoint
The transaction service supports to filter out spam token or to only show trusted tokens. The client gateway should support the same.

jpalvarezl:
Closes #163 

Changes:
- Added `trusted` and `exclude_spam` filters, both with `false` default value

### DIFF
--- a/src/routes/balances.rs
+++ b/src/routes/balances.rs
@@ -5,15 +5,25 @@ use crate::utils::context::Context;
 use crate::utils::errors::ApiResult;
 use rocket::response::content;
 
-#[get("/v1/safes/<safe_address>/balances/<fiat>")]
+#[get("/v1/safes/<safe_address>/balances/<fiat>?<trusted>&<exclude_spam>")]
 pub fn get_balances(
     context: Context,
     safe_address: String,
     fiat: String,
+    trusted: Option<bool>,
+    exclude_spam: Option<bool>,
 ) -> ApiResult<content::Json<String>> {
+    let trusted = trusted.unwrap_or(false);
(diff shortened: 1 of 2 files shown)


src/routes/balances.rs
```diff
@@ -5,15 +5,25 @@ use crate::utils::context::Context;
...
+#[get("/v1/safes/<safe_address>/balances/<fiat>?<trusted>&<exclude_spam>")]
 pub fn get_balances(
     context: Context,
     safe_address: String,
     fiat: String,
+    trusted: Option<bool>,
+    exclude_spam: Option<bool>,
 ) -> ApiResult<content::Json<String>> {
+    let trusted = trusted.unwrap_or(false);
+    let exclude_spam = exclude_spam.unwrap_or(false);
```
rmeissner, line 17: Not sure if we should default to `true` here

rmeissner (approved):

src/routes/balances.rs
jpalvarezl: I matched the values in the tx service for the `/safes/{address}/balances/usd/` endpoint

rmeissner: Yes, but the clients should not show spam tokens by default, therefore it would make more sense to have `true` as a default for the client gateway.

jpalvarezl: Np, I can change it to `true`, I will create an issue, just for documentation's sake.

PR #176, opened 2020-10-28
Included the fiat total as part of the balance response

### LINKED ISSUE #175: Implement balance endpoint with total fiat
The aforementioned endpoint would return an object containing the total in the desired fiat and a vector with each individual balance:

```rust
pub enum Balances {
    fiat_total: String, // total balance in the requested fiat
    Items: Vec<Balance>
}
```

`fiat_total` will contain the aggregated sum of all the `items.fiat_balance`.

jpalvarezl:
Closes #175 

- Included `fiat_total` in the root of the response object
- All the individual balances are included now under `items`

### DIFF
--- a/src/models/service/balances.rs
+++ b/src/models/service/balances.rs
@@ -9,3 +9,10 @@ pub struct Balance {
     pub fiat_balance: String,
     pub fiat_conversion: String,
 }
+
+#[derive(Serialize, Debug, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub struct Balances {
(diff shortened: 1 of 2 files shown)


rmeissner (approved):
Looks good (and simple) ... a test would have been nice, but I know that this is outside of our current testing strategy, but maybe it would be a nice first tested method of the `services` module (but as a separate issue, this one looks good :) )

jpalvarezl:
Thanks! Will do, I will open an issue for testing the `service` module as soon as we can turn context into a trait :)

PR #182, opened 2020-11-02
Missing decimal for Erc721TokenInfo type was replaced with TokenInfo

### LINKED ISSUE #181: Service crash when inconsistent token types are served in transaction
Offending transaction: https://safe-transaction.mainnet.gnosis.io/api/v1/safes/0xBc79855178842FDBA0c353494895DEEf509E26bB/all-transactions/?limit=1&offset=1

Reference: https://github.com/gnosis/safe-transaction-service/issues/133

- In the case of a `Erc20TransferDto` with a token with type `TokenInfo::Erc721`: the `value` field` will contain the `tokenId`
- In the case of a `Erc721TransferDto` with a token with type `TokenInfo::Erc20`: the `tokenId` will contain the `value` field  

Map the crossed fields to the corresponding values and return the `txInfo` object with the correct `TransactionInfo::Transfer`

jpalvarezl:
Closes #181 

- Used `default_if_null` custom deserialiser for decimals in `Erc20TokenInfo`

Note: If the backend will serve an `Erc721Transfer` with an `Erc20TokenInfo` we map the same inconsistency. As I understood, we just shouldn't crash, right? Otherwise this fix would have to be a bit more involved.


src/models/converters/tests/transfers.rs
```diff
@@ -165,3 +165,27 @@ fn transfer_dto_to_transaction_details() {
...
+    mock_info_provider.expect_safe_info().times(0);
+    mock_info_provider.expect_token_info().times(0);
+    let tx = serde_json::from_str::<Erc20TransferDto>(
+        crate::json::ERC_20_TRANSFER_WITH_ERC721_TOKEN_INFO,
+    )
+    .unwrap();
+
+    let actual = tx.to_transfer_info(&mut mock_info_provider);
+
+    let expected = TransferInfo::Erc20(Erc20Transfer {
```
rmeissner, line 181: This is not the result I would expect.

Either we mark it as what it is, an ERC721 transfer, as we are transfering a erc721 token (the id is the value returned by the backend).
Or we just ignore this for now and don't convert it.

But this way we display it wrongly and in the end it says you transfered a large amount of this token.

rmeissner (changes requested):

src/models/converters/tests/transfers.rs
jpalvarezl: The `tokenId` is also `null` in this particular case. The deserialiser is already rendering this as an `Erc20Transfer`. the `type` field coming from the backend is of `"Erc20Transfer"` type so it really complicates things for us.

A solution I see is to refactor these deserialisers to be more generic and use `TokenInfo` instead?

jpalvarezl: So `TokenInfo` defines the type of transfer, rather than the transfer itself. Although big changes here are definitely breaking changes.

jpalvarezl: Scratch what I said, if we don't change anything in the `models::service` module there is no breaking change to be made.

rmeissner: >  type so it really complicates things for us.

Then handle the throw, so that it doesn't crash the view rendering and skip this value when converting the transfer array. Proper fix can be done later (actually @Uxio0 will do that in the core services).

Uxio0: I will deploy the fix tomorrow morning with some others bugfixes

jpalvarezl:
It seems to be possible to use `TokenInfo` for `Transfers`, in that case we get a token type against which we can assert whether it matches that of the transfer. If assertion fails, the `txType` is returned as `Unknown`:
```json
[
   {
      "id": "ethereum_0xBc79855178842FDBA0c353494895DEEf509E26bB_0x2f920d8b75eff1857266643aedcfcf677b834ca164b6684a23a2c8e4574432ec_0x271d310e29f92da0",
      "timestamp": 1604284676000,
      "txStatus": "SUCCESS",
      "txInfo": {
        "type": "Unknown"
      },
      "executionInfo": null
    },
    {
      "id": "ethereum_0xBc79855178842FDBA0c353494895DEEf509E26bB_0x317db9d079e46fef2f758e37bd20efb14d5c83e2510307079207bc6f04cdee48_0xda87a3cf48c093c7",
      "timestamp": 1604136681000,
      "txStatus": "SUCCESS",
      "txInfo": {
        "type": "Unknown"
      },
      "executionInfo": null
    }
]
```

rmeissner (commented):
I feel like this is quite a big change. So if we start this I would not do it half way.

E.g. 
- Remove expected type from `token_info_with_fallback` and make it a global function (instead of a function on the struct.
- Change `to_transfer_info` of Erc721TransferDto Erc20TransferDto to be able to handle any token type to build the transfer info. This could maybe also be a method that is not specific to the structs that takes (address, value, token). Where value could also be an id (in case of an ERC721). Token would be `info_ref`.

src/models/converters/transfers.rs
```diff
@@ -36,7 +49,7 @@ impl TransferDto {
         Ok(TransactionDetails {
             executed_at: self.get_execution_time(),
             tx_status: TransactionStatus::Success,
-            tx_info: self.to_transfer(info_provider, safe),
+            tx_info: self.to_transfer(info_provider, safe)?,
```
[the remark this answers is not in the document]
jpalvarezl: I removed that change :)

jpalvarezl:
There is still a `panic!` for the else branch of a `match` block. It's just for when a transfer is not either ERC20 or ERC721. 

There was minimal changes to tests I think, so the old behaviour was preserved as much as possible, by just adding that the `TokenType` takes priority over the decision as to what sort of transfer transaction this corresponds to.

src/models/converters/transfers.rs
```diff
@@ -170,24 +130,39 @@ impl EtherTransferDto {
...
+            logo_uri: token_info.map(|it| it.logo_uri.to_owned()).flatten(),
+            decimals: token_info.map(|it| it.decimals.to_owned()),
+            value: element.to_owned(),
+        }),
+        TokenType::Erc721 => TransferInfo::Erc721(Erc721Transfer {
+            token_address: token_address.to_owned(),
+            token_id: element.to_owned(),
+            token_name: token_info.map(|it| it.name.to_owned()),
+            token_symbol: token_info.map(|it| it.symbol.to_owned()),
+            logo_uri: token_info.map(|it| it.logo_uri.to_owned()).flatten(),
```
rmeissner, line 156: Why do we have to flatten this?

rmeissner (approved):

src/models/converters/transfers.rs
jpalvarezl: Because here we have `Option<TokenInfo>` and within `token_info` we have `Option<String>` for `logo_uri`, so we are in a "option of an option" situation, with flatten you get rid of all the nested options.

PR #202, opened 2020-12-04
Queued and History transaction endpoints

### LINKED ISSUE #188: DRAFT transaction summary list v2
```rust
pub enum ListItem {
    label{ section_type: SectionType },
    transaction(Transaction),
    date_label{ timestamp_millis: u64 }, 
    
    group_start{ group_nonce: String }, // use transaction endpoint filter by nonce and get page count
    group_end{ group_nonce: String },
    
    //alternatively
    
    conflict_header{ nonce: String, count: String },
    conflict_transaction{ transaction: Transaction, position: Position },
}

pub enum Position { start, mid, end }

pub enum SectionType { Next, Queued, History }
```

jpalvarezl:
Closes #188 

Changes: 
- Adds `history` and `queued` transaction endpoints
- Server side rendering for lists and labels


src/services/transactions_queued.rs
```diff
@@ -0,0 +1,215 @@
...
+        display_trusted_only
+    );
+
+    let body = context
+        .cache()
+        .request_cached(&context.client(), &url, request_cache_duration())?;
+    let mut backend_transactions: Page<MultisigTransaction> = serde_json::from_str(&body)?;
+
+    // If there is a next url we remove the last item for information on the next page
+    let after_tx = if backend_transactions.next.is_some() {
```
jpalvarezl, line 47: Here I would use `and_then` : https://doc.rust-lang.org/std/option/enum.Option.html#method.and_then

src/models/converters/page_metadata.rs
```diff
@@ -0,0 +1,31 @@
...
+
+        let chunked: Vec<Vec<&str>> = encoded
+            .split("&")
+            .map(|it| it.split("=").collect())
+            .collect();
+
+        chunked.into_iter().for_each(|it| {
+            let first = it.first().unwrap_or(&"").deref();
+            if first == "limit" {
+                output.limit = it.get(1).unwrap_or(&"0").parse::<u64>().unwrap_or(0);
```
rmeissner, line 23: we should default to `20` here too

src/services/transactions_queued.rs
```diff
@@ -0,0 +1,215 @@
...
+
+    let mut service_transactions: Vec<TransactionListItem> = Vec::new();
+    for (group_nonce, transaction_group) in
+        &tx_iter.group_by(|transaction| transaction.nonce as i64)
+    {
+        // Check if we need to add section headers
+        if last_proccessed_nonce < safe_nonce && group_nonce == safe_nonce {
+            // If the last nonce processed was the initial nonce (-1) and this group nonce is the current Safe nonce then we start the Next section
+            service_transactions.push(TransactionListItem::Label { label: Label::Next })
+        } else if last_proccessed_nonce <= safe_nonce && group_nonce > safe_nonce {
```
jpalvarezl, line 74: The first part of this `else if` statement is always `true` , right? What id you check `last_proccessed_nonce == safe_nonce`?

rmeissner (approved):

src/services/transactions_queued.rs
```diff
@@ -0,0 +1,215 @@
...
+        let conflict_from_previous_page = previous_page_nonce == group_nonce;
+        // If there is more than 1 item in this group or we are in an edge group then we have a conflict
+        let has_conflicts = group_iter.peek().is_some() || is_edge_group;
+        // If we start a new conflict group then we should render the conflict header
+        if has_conflicts && !conflict_from_previous_page {
+            service_transactions.push(TransactionListItem::ConflictHeader {
+                nonce: group_nonce as u64,
+            })
+        }
+        // Add the one transaction that is always present
```
jpalvarezl, line 100: This is the first one in the group. If this one doesn't exist `group_by` wouldn't have caught it. 🤔  What if we check `transaction_group.len() > 1` ? Would that simplify this?

rmeissner: transaction_group and iterator doesn'T have a len property

src/services/transactions_queued.rs
rmeissner: yes would make this a little nicer

src/services/transactions_queued.rs
rmeissner: no it is not always true ... if you don't have a next transaction on the first page for example

src/services/transactions_queued.rs
rmeissner: Also not sure what you simplify here :D

jpalvarezl: hahaha I see, right iterator... Adding the conflict header ,maybe? But also not a lot of value would be added, I guess

PR #209, opened 2020-12-09
Closes #206: Refactor and test queue endpoint

### LINKED ISSUE #206: Refactor and test queued endpoint for better testability
- Extract logic into method that can be easily tested without dependencies on Context
- Test extracted logic

rmeissner:


src/services/transactions_queued.rs
```diff
@@ -43,26 +43,79 @@ pub fn get_queued_transactions(
...
+            1, // Direction forward
+        ),
+        previous: build_page_url(
+            context,
+            &safe_address,
+            &page_meta,
+            timezone_offset,
+            display_trusted_only,
+            backend_transactions.previous,
+            -1, // Direction backwards
```
jpalvarezl, line 80: We could use an enum like `pub enum LinkDirection { Previous, Next }`, also usable in the `history` endpoint.

src/services/transactions_queued.rs
```diff
@@ -43,26 +43,79 @@ pub fn get_queued_transactions(
...
+        ),
+        results: service_transactions,
+    })
+}
+
+// Nonce of the first item in the next page (-1 if not present)
+fn get_edge_nonce(backend_transactions: &mut Page<MultisigTransaction>) -> i64 {
     // If there is a next url we remove the last item for information on the next page
-    let after_tx = if backend_transactions.next.is_some() {
+    if backend_transactions.next.is_some() {
```
jpalvarezl, line 89: We could replace from line 89 to 92 with `backend_transactions.next.and_then(|_| backend_transactions.results.pop()).map_or(-1, |tx| tx.nonce as i64`

src/services/transactions_queued.rs
```diff
@@ -43,26 +43,79 @@ pub fn get_queued_transactions(
...
+}
+
+// Nonce of the last item in the previous page (-1 if not present)
+fn get_previous_page_nonce(
+    page_meta: &PageMetadata,
+    tx_iter: &mut dyn Iterator<Item = MultisigTransaction>,
+) -> i64 {
     // If we are not on the first page then we take the first item to get information on the previous page
-    let before_tx = if page_meta.offset == 0 {
+    if page_meta.offset == 0 {
```
jpalvarezl, line 103: same as above, maybe we should use `and_then` here too.

src/services/transactions_queued.rs
```diff
@@ -43,26 +43,79 @@ pub fn get_queued_transactions(
...
-    // Nonce of the first item in the next page (-1 if not present)
-    let edge_nonce = after_tx.map_or(-1, |tx| tx.nonce as i64);
-    // Nonce of the last item in the previous page (-1 if not present)
-    let previous_page_nonce = before_tx.map_or(-1, |tx| tx.nonce as i64);
-    let mut last_proccessed_nonce = previous_page_nonce;
+    }
+    .map_or(-1, |tx| tx.nonce as i64)
+}
 
+fn process_transactions(
```
jpalvarezl, line 111: Generally for me, `service_transactions` correspond to the types from the gateway as opposed to the `backend_transaction` and in this case `tx_list_item` as this refers to a screen on the app rather than a domain type. If you agree with this naming, I would change the name of `service_transactions` to `tx_list_items`. `tx_iter` is fine though, as the `MultisigTransaction` type is right next to it and then we know those are backend txs.

src/services/transactions_queued.rs
rmeissner: make calculation down the road harder, as I need a `-1` and `1` ;)

src/services/transactions_queued.rs
rmeissner: yes I would do this after adding the test :)

src/services/transactions_queued.rs
rmeissner: no strong preferences here

src/services/transactions_queued.rs
rmeissner: Ahhh maybe just setting the values to the enum variants would be nice :) 

Again I would refactor this after adding tests :)

src/services/tests/transactions_queue.rs
```diff
@@ -0,0 +1,614 @@
+use crate::json::{
```
rmeissner, line 1: The name of the file should be `_queued`

jpalvarezl (approved):

PR #213, opened 2020-12-16
Added default value to 0 for ERC20 transfer

jpalvarezl:

### DIFF
--- a/src/models/backend/transfers.rs
+++ b/src/models/backend/transfers.rs
@@ -51,21 +51,6 @@ pub struct Erc20Transfer {
 #[derive(Derivative, Deserialize, Debug, Clone)]
 #[derivative(Hash)]
 #[serde(rename_all = "camelCase")]
-pub struct Erc20TransferMalformed {
-    pub execution_date: DateTime<Utc>,
-    pub block_number: u64,
-    pub transaction_hash: String,
-    pub to: String,
-    pub token_id: String,
-    pub token_address: String,
-    #[derivative(Hash = "ignore")]
-    pub token_info: Option<TokenInfo>,
-    pub from: String,
-}
-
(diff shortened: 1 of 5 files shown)


src/models/backend/transfers.rs
```diff
@@ -39,13 +40,29 @@ pub struct Erc20Transfer {
     pub block_number: u64,
     pub transaction_hash: String,
     pub to: String,
+    #[serde(deserialize_with = "default_if_null")]
```
rmeissner, line 43: Should we do the same for tokenId in an erc721 transfer?

src/models/backend/transfers.rs
```diff
@@ -39,13 +40,29 @@ pub struct Erc20Transfer {
...
     pub token_address: String,
     #[derivative(Hash = "ignore")]
     pub token_info: Option<TokenInfo>,
     pub from: String,
 }
 
+#[derive(Derivative, Deserialize, Debug, Clone)]
+#[derivative(Hash)]
+#[serde(rename_all = "camelCase")]
+pub struct Erc20TransferMalformed {
```
rmeissner, line 54: Not a big fan of this :/

jpalvarezl: Oh wait, that's totally unnecessary actually, just forgot to clean up.

src/models/backend/transfers.rs
jpalvarezl: I am not sure if `tokenId = 0` has a special meaning. I checked the standard quickly and seems to be ok. If you can think of a situation where `tokenId = 0` may be misunderstood let me know.

rmeissner: it would be similarly confusing to a user as a wrong value amount, imo

jpalvarezl:
Won't fix for now

PR #215, opened 2020-12-17
Filtered only transactions that aren't queued and were in fact executed

### LINKED ISSUE #210: DATE_LABEL not always decreasing
**Describe the bug**
DATE_LABEL not always decreasing. Sometimes a higher date is below a lower date and sometimes dates are repeated.

**To Reproduce**
- Open tx list in poc for safe: 0x1230B3d59858296A31053C1b8562Ecf89A2f888b

**Expected behavior**
DATE_LABELS should be continuously decreasing. See screenshots.

**Desktop (please complete the following information):**
 - OS: Android

**Additional context**
1 | 2 
-- | -- 
image | Screenshot_20201211-183230

jpalvarezl:
Closes #210 

- Fetching only executed transactions for the `all-transactions` endpoint

### DIFF
--- a/src/services/transactions_history.rs
+++ b/src/services/transactions_history.rs
@@ -107,7 +107,7 @@ fn fetch_backend_paged_txs(
 ) -> Result<Page<Transaction>> {
     let page_metadata = PageMetadata::from_url_string(page_url.as_ref().unwrap_or(&"".to_string()));
     let url = format!(
-        "{}/v1/safes/{}/all-transactions/?{}&queued=false",
+        "{}/v1/safes/{}/all-transactions/?{}&queued=false&executed=true",
         base_transaction_service_url(),
         safe_address,
         page_metadata.to_url_string()


src/services/transactions_history.rs
```diff
@@ -107,7 +107,7 @@ fn fetch_backend_paged_txs(
 ) -> Result<Page<Transaction>> {
     let page_metadata = PageMetadata::from_url_string(page_url.as_ref().unwrap_or(&"".to_string()));
     let url = format!(
-        "{}/v1/safes/{}/all-transactions/?{}&queued=false",
+        "{}/v1/safes/{}/all-transactions/?{}&queued=false&executed=true",
```
rmeissner, line 110: Do we need both? shouldn't `executed=true` be enough? 🤔 but should also not harm

rmeissner (approved):

src/services/transactions_history.rs
jpalvarezl: You are right. I think `queued=false` transactions should be a subset of `executed=true` but just to be extra safe 🤔 maybe instead of being safe it is just confusing :P I can change it if you prefer

PR #221, opened 2021-01-13
Origin => Safe App Info

### LINKED ISSUE #220: [Safe Client Gateway] Return safe app information (core service's `origin` field)
## The feature

Deserialise field in `MultisigTransactions.origin` which contains a serialized json object and is nullable. 

Examples:
- https://safe-transaction.staging.gnosisdev.com/api/v1/safes/0x1230B3d59858296A31053C1b8562Ecf89A2f888b/all-transactions/?limit=1&offset=24

- https://safe-transaction.gnosis.io/api/v1/safes/0xBc79855178842FDBA0c353494895DEEf509E26bB/all-transactions/?limit=1&offset=31

The json can be truncated (particularly the name, but it's probably best to not assume anything)

Use the manifest of the app to provide `name`, `url` and `icon`. Example manifest: https://apps.gnosis-safe.io/walletConnect/manifest.json

Every app must contain a `manifest.json`. This is fetched using the url from the `orgin` field when possible.

Proposals:

- The object struture: (nullability of the fields is defined by https://docs.gnosis.io/safe/docs/sdks_safe_apps/#get-your-safe-app-listed-in-the-safe-multisig)
```rust
struct SafeAppInfo {
  name: String,
  url: String,
  logo_url: String,
}
```
 - This object will be part of the base transaction summary object like so:
```rust 
struct TransactionSummary {
  id: String,
  timestamp: i64,
  tx_status: TransactionStatus,
  tx_info: TransactionInfo,
  execution_info: Option<ExecutionInfo>,
  safe_app_info: Option<SafeAppInfo>, // new nullable field
}
```

Also in `TransactionDetails`:

```rust
struct TransactionDetails {
  executed_at: Option<i64>,
  tx_status: TransactionStatus,
  tx_info: TransactionInfo,
  tx_data: Option<TransactionData>,
  detailed_execution_info: Option<DetailedExecutionInfo>,
  tx_hash: Option<String>,
  safe_app_info: Option<SafeAppInfo>, // new nullable field
}
```

- Introduce new cache duration value for this (should be longer than the default)

## (Optional)
Check for urls starting with “ipfs.org” :
- Might timeout
-  “Ipfs.org” replace with “ipfs.gnosis.io” (replace only for internal request of the manifest, expose the original URL)

Always try to resolve with the url and the manifest of the safe app

jpalvarezl:
Closes #220 

Changes introduced:
- Added to the `TransactionDetails` and `TransactionSummary` structs an optional field `safe_app_info`
- Deserialization of `MultisigTransaction.origin` defaults to `None` 
- Optional part still missing


src/config/mod.rs
```diff
@@ -35,6 +35,10 @@ pub fn about_cache_duration() -> usize {
     usize_with_default("ABOUT_CACHE_DURATION", request_cache_duration())
 }
 
+pub fn long_cache() -> usize {
```
rmeissner, line 38: Any specific reason why here we use a duration based env var, while the others are more use case based (long cache vs about cache). Personally I would stick to use case, but there might be good reasons to have it duration based.

src/models/converters/transactions/safe_app_info.rs
```diff
@@ -0,0 +1,41 @@
...
+    pub(super) description: String,
+    #[serde(rename(deserialize = "iconPath"))]
+    pub(super) icon_path: String,
+}
+
+impl OriginInternal {
+    fn to_safe_app_info(&self, info_provider: &mut dyn InfoProvider) -> Option<SafeAppInfo> {
+        let manifest_url = format!("{}/manifest.json", self.url);
+        info_provider
+            .raw_request(&manifest_url)
```
rmeissner, line 32: should we add a `load_safe_app_info` to the InfoProvider?

src/models/converters/transactions/safe_app_info.rs
```diff
@@ -0,0 +1,41 @@
...
+#[derive(Deserialize, Debug, PartialEq)]
+pub(super) struct Manifest {
+    pub(super) name: String,
+    pub(super) description: String,
+    #[serde(rename(deserialize = "iconPath"))]
+    pub(super) icon_path: String,
+}
+
+impl OriginInternal {
+    fn to_safe_app_info(&self, info_provider: &mut dyn InfoProvider) -> Option<SafeAppInfo> {
```
rmeissner, line 29: Wouldn't it be nicer to return a Result here that we then convert to an Option in the `safe_app_info_from` method? (would allow it to use the `?` to unwrap)

src/models/service/transactions/summary.rs
```diff
@@ -21,6 +22,14 @@ pub struct ExecutionInfo {
     pub missing_signers: Option<Vec<String>>,
 }
 
+#[derive(Serialize, Debug, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub struct SafeAppInfo {
```
rmeissner, line 27: maybe have this in the `mod.rs`?

src/providers/info.rs
```diff
@@ -76,6 +77,10 @@ impl InfoProvider for DefaultInfoProvider<'_> {
             anyhow::bail!("Token Address is 0x0")
         }
     }
+
+    fn raw_request(&mut self, url: &str) -> ApiResult<String> {
```
rmeissner, line 81: I am not sure about this method .... I would prefer if the InfoProvider is very use case specific (so returns deserialized data)

src/models/service/transactions/summary.rs
jpalvarezl: I think I will move it to the `info.rs` with the rest of the stuff.

src/models/converters/transactions/safe_app_info.rs
jpalvarezl: that's a good a idea 👍

src/config/mod.rs
jpalvarezl: It was because of what you mentioned the other day regarding having caches based in duration in the future. If you prefer we can stick to use case for now and then do a full migration of everything when we decide a concrete naming scheme for this.

rmeissner (approved):

PR #230, opened 2021-01-19
Consumption of timezone offset and adjustment of datelabels

### LINKED ISSUE #225: Transaction History endpoint time offset param
## Enhancement

The `v1/safes/<safe_address>/transactions/history` endpoint returns date labels  foreach day in UTC. This may not reflect the situation of the location where the data is being requested from. The date labels should be added in concordance to the time zone of the requesting client. 

Use the query parameter `timezone_offset` to offset the date label calculation and conditional addition to the list.

jpalvarezl:
Closes #225 

- `timezone_offset` sent by the clients is used to adjust the calculation of date labels
- Added tests for UTC +1:00 and UTC+11:00

### DIFF
--- a/src/services/tests/transactions_history.rs
+++ b/src/services/tests/transactions_history.rs
@@ -228,8 +228,9 @@ fn backend_txs_to_summary_txs_with_values() {
 #[test]
 fn service_txs_to_tx_list_items_empty() {
     let service_tx: Vec<TransactionSummary> = vec![];
+    let berlin_timezone_offset = 3600;
 
-    let actual = service_txs_to_tx_list_items(service_tx, -1).unwrap();
+    let actual = service_txs_to_tx_list_items(service_tx, -1, berlin_timezone_offset).unwrap();
 
     assert_eq!(actual.is_empty(), true);
 }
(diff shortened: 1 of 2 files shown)


src/services/tests/transactions_history.rs
```diff
@@ -372,13 +379,120 @@ fn service_txs_to_tx_list_items_last_timestamp_defined_and_same() {
...
+
+#[test]
+fn service_txs_to_tx_list_items_date_label_someplace_else_timezone() {
+    let mut mock_info_provider = MockInfoProvider::new();
+    mock_info_provider.expect_safe_info().times(0);
+    mock_info_provider.expect_token_info().times(0);
+
+    let service_txs = get_service_txs(&mut mock_info_provider);
+    let service_txs_copy = get_service_txs(&mut mock_info_provider);
+    let berlin_timezone_offset = 39600; // + 11 hours Melbourne/Australia
```
rmeissner, line 446: Berlin in Australia?

rmeissner (approved):
Looks very nice ... really like the tests :)

PR #235, opened 2021-01-21
Known Addresses

### LINKED ISSUE #198: Return names of known addresses
# What is needed?
To provide human-readable names for known / trusted addresses (exchanges, popular smart contracts etc.), there should be an external source that the clients can query where known addresses are mapped to a name. This database can be manually curated in a similar manner as the "trusted" / "spam" tokens.

This relates to this frontend epic: https://github.com/gnosis/safe/issues/337

# Background
There are a range of addresses, where their names are "public knowledge". Yet if the user does not add them to their address book manually, the user would only see the non-human-readable Ethereum address when interacting with those addresses.

# Technical details

** update by @jpalvarezl **
** update by @jpalvarezl 20/01/2021 **
** update by @jpalvarezl 25/01/2021 **

## Feature: technical side

It is necessary to include additional information on the addresses for all transaction types. `TransactionInfo` is the best place to do this, as it is available in both the `TransactionSummary` and `TransactionDetails` structs.

The priority order of data sources for enriching address is `TokenInfo > KnownAddressInfo` 

This should be included for all the variants of `TransactionInfo`, namely `Transfer`, `SettingsChange` and `Custom`.

For `TransactionInfo::SettingsChange` we include this information within `SettingsInfo` in the following structure: (note that only variants with an address field can contain these changes, `ChangeThreshold` doesn't have any address and it's therefore excluded)

```rust
pub enum SettingsInfo {
    SetFallbackHandler {
        handler: String,
        handler_info: Option<AddressInfo>,
    },
    AddOwner {
        owner: String,
        owner_info: Option<AddressInfo>,
        threshold: u64,
    },
    RemoveOwner {
        owner: String,
        owner_info: Option<AddressInfo>,
        threshold: u64,
    },
    SwapOwner {
        old_owner: String,
        old_owner_info: Option<AddressInfo>,
        new_owner: String,
        new_owner_info: Option<AddressInfo>,
    },
    // no changes
    ChangeThreshold {
        threshold: u64,
    },
    ChangeImplementation {
        implementation: String,
        implementation_info: Option<AddressInfo>,
    },
    EnableModule {
        module: String,
        module_info: Option<AddressInfo>,
    },
    DisableModule {
        module: String,
        module_info: Option<AddressInfo>,
    },
}

pub struct Custom {
    pub to: String,
    pub data_size: String,
    pub value: String,
    pub method_name: Option<String>,
    pub action_count: Option<usize>,
    pub to_info: Option<AddressInfo>
}

pub struct Transfer {
    pub sender: String,
    pub sender_info: Option<AddressInfo>,
    pub recipient: String,
    pub recipient_info: Option<AddressInfo>,
    pub direction: TransferDirection,
    pub transfer_info: TransferInfo,
}
```

`AddressInfo` is polymorphic as specified in the `Entities` section of the issue.

## Endpoints

Known address resource: Pending merge of https://github.com/gnosis/safe-transaction-service/issues/208

## Entities

(Note for clients: all names will be serialised **camelCased**) 

The proposed structure for `AddressInfo` is as follows:

```rust
pub enum AddressInfo {
  pub name: String,
  pub logo_uri: Option<String>,
}
```

Cache in a long term cache as this will result in many requests in the worst case scenario and data is not expected to change much.

jpalvarezl:
Closes #198 

TODO's:
- ~~Implement source of known addresses~~
- ~~Tests~~
- ~~Module transactions~~

Proposed changes:
- Added method `InfoProvider::address_info` fetching a `token_info` and then `contract_info`
- Added `address_info` for `SettingsChange` transactions for each available address field
- Added `address_info` for `Custom` transactions for `to` address
- Added `sender_info` and `recipient_info` for `Transfers` checking the direction of the transfer to avoid poluting the cache
- Small optimization, will leave a comment below to bring it to your attention


src/models/converters/transactions/mod.rs
```diff
@@ -89,16 +89,16 @@ impl MultisigTransaction {
             .unwrap_or(false)
             && check_sender_or_receiver(&self.data_decoded, &self.safe)
         {
-            info_provider.token_info(&self.to).map_or(
-                TransactionInfo::Custom(self.to_custom()),
-                |token| match token.token_type {
+            match info_provider.token_info(&self.to) {
```
jpalvarezl, line 92: While setting up mocks, I noticed that `map_or` used `Transaction::Custom(self.to_custom)` as the default **value**, so it was always evaluated. Even if it turned out to be a transfer. There is a version of `map_or_else` that uses a callback to evaluate the default if needed, but because `info_provider` is passed to 2 closures the compiler doesn't like it.

Long story short: I unfolded this stuff into a couple of match blocks to prevent the eager calling of `self.to_custom` which resulted into a nice little optimization. Maybe I could use `if let Ok(token) {  match token.type // ... } else { //.... }` if it's more readable.

src/models/converters/mod.rs
```diff
@@ -18,3 +20,15 @@ pub(super) fn get_transfer_direction(safe: &str, from: &str, to: &str) -> Transf
...
+pub(super) fn get_address_info(
+    safe: &str,
+    address: &str,
+    info_provide: &mut dyn InfoProvider,
+) -> Option<AddressInfo> {
+    if safe != address {
+        info_provide.address_info(address).ok()
+    } else {
+        None
+    }
```
jpalvarezl, line 33: Used only in transfers

jpalvarezl:
I am not sure what's causing the massive dip in coverage 🤔 But I would also like to avoid duplicating too many tests as this PR is quite noisy as it is. Wdyt?

src/providers/info.rs
```diff
@@ -107,6 +109,33 @@ impl InfoProvider for DefaultInfoProvider<'_> {
...
+                let contract_info = serde_json::from_str::<ContractInfo>(&contract_info_json)?;
+                if contract_info.display_name.trim().is_empty() {
+                    anyhow::bail!("No display name")
+                } else {
+                    Ok(AddressInfo {
+                        name: contract_info.display_name.to_owned(),
+                        logo_uri: contract_info.logo_uri.to_owned(),
+                    })
+                }
+            })
```
jpalvarezl, line 137: Just marking where the data is fetched for easier review

src/models/converters/mod.rs
```diff
@@ -18,3 +20,15 @@ pub(super) fn get_transfer_direction(safe: &str, from: &str, to: &str) -> Transf
         TransferDirection::Unknown
     }
 }
+
+pub(super) fn get_address_info(
+    safe: &str,
+    address: &str,
+    info_provide: &mut dyn InfoProvider,
+) -> Option<AddressInfo> {
+    if safe != address {
```
rmeissner, line 29: Could we add a comment for this? 

Just to avoid that we look at this in 1 week and have no idea why we did this.

src/models/converters/transactions/mod.rs
```diff
@@ -150,9 +167,11 @@ impl MultisigTransaction {
         }
     }
 
-    fn to_ether_transfer(&self) -> Transfer {
+    fn to_ether_transfer(&self, info_provider: &mut dyn InfoProvider) -> Transfer {
         Transfer {
             sender: self.safe.to_owned(),
+            sender_info: None,
```
rmeissner, line 173: Little very unimportant details but it triggers my ocd: can we have `sender`, `sender_info`, `recipient` and `recipient_info` always defined in the same order (currently the token transfers have a different order than the ether transfer) 😄

src/models/converters/transactions/mod.rs
```diff
@@ -181,6 +200,7 @@ impl MultisigTransaction {
                 .data_decoded
                 .as_ref()
                 .and_then(|it| it.get_action_count()),
+            to_info: info_provider.address_info(&self.to).ok(), //TODO test
```
rmeissner, line 203: similar here, could we move it right next to the `to` similar to how it is in the `ModuleTransaction`?

src/models/service/transactions/summary.rs
```diff
@@ -10,6 +10,7 @@ pub struct TransactionSummary {
     pub tx_status: TransactionStatus,
     pub tx_info: TransactionInfo,
     pub execution_info: Option<ExecutionInfo>,
+    #[serde(skip_serializing_if = "Option::is_none")]
```
rmeissner, line 13: why only for `safe_app_info` and not for `execution_info` ... should we be consistent for this and always do it (or never), Might not be important for this issue, but worth aligning on.

src/providers/info.rs
```diff
@@ -107,6 +109,33 @@ impl InfoProvider for DefaultInfoProvider<'_> {
...
+            .or_else(|_| {
+                let url = format!(
+                    "{}/v1/contracts/{}/",
+                    base_transaction_service_url(),
+                    address
+                );
+                let contract_info_json =
+                    self.cache
+                        .request_cached(self.client, &url, safe_app_manifest_cache())?;
```
rmeissner, line 127: not sure if we should use this cache timeout here, very different use cases 🤔

rmeissner (approved):
A couple minor things, nothing that should stop a merge.

Only thing we should address is to use a different (maybe even new) env var for the cache timeout for the contract info.

src/models/service/transactions/summary.rs
jpalvarezl: I just added it to `safe_app_info` because it is more often null, I can do it for `execution_info` too 👍 I don't really oppose to this, but it would be nice to be able to do it at a struct level rather than a field level :(

src/providers/info.rs
jpalvarezl: It was just the only available super long cache duration, I was thinking we could clean this up with the task you mentioned regarding having cache durations named after their duration rather than the functionality in which they are used.

Issue #241, opened 2021-02-02
Migrate to use github action. Restore caching as it was disabled in the last PR.


Uxio0:
https://github.com/actions-rs/meta
https://levans.fr/rust_travis_cache.html

https://github.com/actions-rs/grcov
https://github.com/actions-rs/example/blob/master/.github/workflows/grcov.yml

https://svartalf.info/posts/2020-04-10-github-action-for-binary-crates-installation/

jpalvarezl:
Done with #248

PR #248, opened 2021-02-03
Github action migration

jpalvarezl:
Reference #241 

Changes: 
- Added concurrent jobs `rustfmt` and `build` as requirement for merging
- Deployment is still in test face, I put an `echo` instead just to be sure we don't polute docker before we know the setup works
- remove remnants of travis.yml
- Apply the prior configuration of `grcov` to the `action-rs/grcov.yml` file
- cargo deps are cached
- grcov cached


jpalvarezl:
I was thinking to merge this, I replace the deployment script run with an `echo` just to test this first if you think that's a good idea.

rmeissner:
I would just change the repo setting and remove all traces of travis :D

jpalvarezl:
On the plus side, submission to coveralls works, on the downside (or maybe upside) why the hell we got 5.7% using the same `exlc-*` statements.

jpalvarezl:
Just to document this here. There is `403` being thrown when trying to recover `grcov` from the cache. Not sure what it is associated, but unbeliveable, fetching and building `grcov` is our current build choke point. Cargo cache works like a charm.

rmeissner (approved):

PR #252, opened 2021-02-08
Restored deployment script and adjusted to use GitHub env

jpalvarezl:
Changes:
- Restored to prior deployment script
- Adjusted variables to obtain the info from github action's env

### DIFF
--- a/scripts/deploy_docker.sh
+++ b/scripts/deploy_docker.sh
@@ -2,15 +2,27 @@
 
 set -euo pipefail
 
-if [ "$1" = "develop" -o "$1" = "main" ]; then
-    # If image does not exist, don't use cache
-    docker pull gnosispm/$DOCKERHUB_PROJECT:$1 && \
-    docker build -t $DOCKERHUB_PROJECT -f Dockerfile . --cache-from gnosispm/$DOCKERHUB_PROJECT:$1 || \
-    docker build -t $DOCKERHUB_PROJECT -f Dockerfile .
-else
-    docker pull gnosispm/$DOCKERHUB_PROJECT:staging && \
-    docker build -t $DOCKERHUB_PROJECT -f Dockerfile . --cache-from gnosispm/$DOCKERHUB_PROJECT:staging || \
-    docker build -t $DOCKERHUB_PROJECT -f Dockerfile .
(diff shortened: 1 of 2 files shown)


scripts/deploy_docker.sh
```diff
@@ -2,15 +2,27 @@
...
+
+    # Workflow run number
+    export BUILD_NUMBER=$GITHUB_RUN_NUMBER
+    # strip the first char as that should always be "v" (as tags should be in the format "vX.X.X")
+    description="$(git describe --tags --always)"
+    export VERSION=${description:1}
+
+    echo "Trigger docker build and upload for version $VERSION ($BUILD_NUMBER)"
+
+    echo "$DOCKER_PASSWORD" | docker login -u "$DOCKER_USERNAME" --password-stdin
```
rmeissner, line 15: Don't we have a github actions step to log in already?

scripts/deploy_docker.sh
```diff
@@ -2,15 +2,27 @@
...
-if [ "$1" = "develop" -o "$1" = "main" ]; then
-    # If image does not exist, don't use cache
-    docker pull gnosispm/$DOCKERHUB_PROJECT:$1 && \
-    docker build -t $DOCKERHUB_PROJECT -f Dockerfile . --cache-from gnosispm/$DOCKERHUB_PROJECT:$1 || \
-    docker build -t $DOCKERHUB_PROJECT -f Dockerfile .
-else
-    docker pull gnosispm/$DOCKERHUB_PROJECT:staging && \
-    docker build -t $DOCKERHUB_PROJECT -f Dockerfile . --cache-from gnosispm/$DOCKERHUB_PROJECT:staging || \
-    docker build -t $DOCKERHUB_PROJECT -f Dockerfile .
+if [ "$PULL_REQUEST_NUMBER" = "" ]; then
```
rmeissner, line 5: Why do we need this?

rmeissner (changes requested):

scripts/deploy_docker.sh
jpalvarezl: I just restored the old script, but I can get rid of it. I thought it was just double checking that this was comming from a PR.

rmeissner (approved):

PR #254, opened 2021-02-09
Added error logging

### LINKED ISSUE #197: Properly configure logging
Things that need to be improved:
- Log errors
  - Make sure errors are notified in Slack
- Disable rocket logging
  - Make sure that normal logging still works

Use Case
- It should be possible to find error logs on kibana with the error code and partial url
- There should be notifications to the team when an error orrurs (e.g. on slack)

jpalvarezl:
Closes #197 

Changes:
- `Responder` impl for `ApiError` logs the error too. Example output:
`[2021-02-09T13:23:21Z ERROR safe_client_gateway::utils::errors] ERR::500::/about::ApiErrorMessage: code:1337; message:Some("Error(\"missing field `transactionServiceBaseUrl`\", line: 1, column: 29)"); arguments:None`

- To disable rocket logs we should set `.env::ROCKET_LOG = off`

### DIFF
--- a/src/utils/errors.rs
+++ b/src/utils/errors.rs
@@ -92,7 +92,14 @@ impl fmt::Display for ApiError {
 }
 
 impl<'r> Responder<'r> for ApiError {
-    fn respond_to(self, _: &Request) -> response::Result<'r> {
+    fn respond_to(self, request: &Request) -> response::Result<'r> {
+        log::error!(
+            "ERR::{}::{}::{}",
+            self.status,
+            request.uri().to_string(),
+            self.details
+        );
+
         Response::build()
             .sized_body(Cursor::new(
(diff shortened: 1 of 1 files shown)


rmeissner (approved):
Did you check that the combination of ROCKET_LOG=off and RUST_LOG=debug works so that our logs are still seen?

jpalvarezl:
Yes, with `RUST_LOG=off` there is nothing visible. Using `ROCKET_LOG=off` and `RUST_LOG=debug` shows the logs as usual. There is a bunch of logs that come from the dependencies. We could up the `RUST_LOG=info` , if you wanted to have less noise there, but they are kind of useful, I would keep them if they do no harm:

Screenshot 2021-02-10 at 08 58 54

rmeissner:
I would set it for the services to info or even critical. Or did you ever look at the debug logs on kibana?

PR #256, opened 2021-02-09
POST MultiSig TXs

### LINKED ISSUE #244: Post MultisigTransaction endopint
## Feature

Create an endpoint for submitting transactions in the gateway. 

The endpoint simply forwards the request to the core services.

## Endpoint

`POST /safes/{address}/multisig-transactions/`

jpalvarezl:
Closes #244


src/services/transactions_proposal.rs
```diff
@@ -31,3 +33,31 @@ pub fn submit_confirmation(
...
+    );
+    let response = context
+        .client()
+        .post(&url)
+        .json(&transaction_request)
+        .send()?;
+
+    if response.status().is_success() {
+        let pattern = format!("*{}_{}*", &ID_PREFIX_MULTISIG_TX, &safe_address);
+        context.cache().invalidate_pattern(&pattern);
```
jpalvarezl, line 55: This is a little dodgy, but also not really :/ Maybe it would be better to request the `transaction_id` instead, that way we know which details to invalidate in the cache? Although if this is a whole purpose endpoint, there might not be a `safe_tx_hash` to use for invalidation.

jpalvarezl: Sorry that was poorly worded. I think that having a `tx_details_id` would make this invalidation pattern more accurate. But if this endpoint is meant to be used outside the `tx_details` eventually, then there might not be a `tx_details_id` to send at all.

rmeissner: Why not use the same pattern as for the `submit_confirmation`? the `contract_transaction_hash` is the `safe_tx_hash`

rmeissner: actually it might make sense to invalidate just by `safe_address`, as we want the list to be invalidated too (as we added a new tx)

rmeissner (commented):
I would adjust the caching, besides that looks good

jpalvarezl:
I thought it would be a bit too much to go with the safe but I think you are right, then the clients get a fresh list and everything gets nicely invalidated.

rmeissner (approved):

Issue #246, opened 2021-02-02
Cache error duration


jpalvarezl:
We should define a short duration for errors cached in the gateway.

PR #259, opened 2021-02-11
Warm up cache with TokenInfo

rmeissner:
This PR just shows what changes are required to load all tokens at once into the cache. This will allow us to run some tests on the performace impact.

Findings:
- Lots of small requests to core services result in quite some overhead right now
  - Keep Alive does not solve this
  - Http2 is not available for services right now (we are coordinating with devops to evaluate this)
- Request for unknown addresses that don't hit a known token take longer
  - Information is pulled from on-chain (@Uxio0 maybe we could add a query param, that disables the on-chain lookup?)
- Using `/v1/tokens/?limit=10000` to load all tokens into cache provides improvements
  - Cold load request time goes down by 50% (~15s to ~8s) - Note: I did NOT do extensive benchmarking 
  - Uses ~1.5mb of redis cache (not really an issue) - We should probably check our [REDIS eviction policy](https://redis.io/topics/lru-cache)
  - If we want to do the same for contract info, we should filter to `​/contracts​/` to filter out contract without a display name

Future ideas:
- Predict what caches to warm/ pre-populate caches
  - Keep stats on what Safes interact with the gateway and have task that keps the results for these in cache up to date (e.g. when we get a webhook to invalidate them we would also reload them directly, so that they are up to date)
  - Create a cron-job that updates the caches (e.g. for tokens, for contracts, for safe apps, for most used safes)
    - Related: Add cache admin endpoint
-  Error cache duration: https://github.com/gnosis/safe-client-gateway/issues/246

Changes to caching:
- Use cache prefixes to cleary separate cache entries for tokens from other cache entries (also to avoid that they get invalidated by webhooks)
- load all tokens into cache instead of 1 by 1

Unrelated changes:
- Update pre-commit hook to not format child files

### DIFF
--- a/src/providers/info.rs
+++ b/src/providers/info.rs
@@ -82,89 +82,6 @@ pub struct DefaultInfoProvider<'p> {
     token_cache: HashMap<String, Option<TokenInfo>>,
 }
 
-// TODO: we have 3 time `impl DefaultInfoProvider` maybe combine them
-impl DefaultInfoProvider<'_> {
-    fn load_safe_info(&mut self, safe: &String) -> ApiResult<Option<SafeInfo>> {
-        let url = format!("{}/v1/safes/{}/", base_transaction_service_url(), safe);
-        let data: String = self
-            .cache
-            .request_cached(self.client, &url, info_cache_duration())?;
-        Ok(serde_json::from_str(&data).unwrap_or(None))
-    }
-
-    fn populate_token_cache(&mut self) -> ApiResult<()> {
-        let url = format!("{}/v1/tokens/?limit=10000", base_transaction_service_url());
-        let response = self.client.get(&url).send()?;
-        let data: Page<TokenInfo> = response.json()?;
-        for token in data.results.iter() {
-            self.cache.create(
-                &format!("dip_ti_{}", token.address),
-                &serde_json::to_string(&token)?,
-                info_cache_duration(),
-            )
-        }
-        Ok(())
-    }
-
-    fn check_token_cache(&mut self) -> ApiResult<()> {
-        if self.cache.fetch("dip_tcl").is_some() {
-            // Cache is still up to data
-            return Ok(());
-        }
-        let result = self.populate_token_cache();
-        // If error we use a shorter cache timeout (we do not want to DoS our service in case of an error)
-        self.cache.create(
-            "dip_tcl",
-            "",
-            if result.is_ok() {
-                info_cache_duration()
-            } else {
-                10000
-            },
-        );
-        result
-    }
-
-    // TODO: Check Eviction policies: https://redis.io/topics/lru-cache
-    fn load_token_info(&mut self, token: &String) -> ApiResult<Option<TokenInfo>> {
-        self.check_token_cache()?;
-        match self.cache.fetch(&format!("dip_ti_{}", token)) {
-            Some(cached) => Ok(Some(serde_json::from_str::<TokenInfo>(&cached)?)),
-            None => Ok(None),
-        }
-    }
-
-    pub fn exchange_usd_to(&self, currency_code: &str) -> ApiResult<f64> {
-        let currency_code = currency_code.to_uppercase();
-        let exchange = self.fetch_exchange()?;
-        match exchange.rates {
-            Some(rates) => rates
-                .get(&currency_code)
-                .cloned()
-                .ok_or(api_error!("Currency not found")),
-            None => Err(bail!("Currency not found")),
-        }
-    }
-
-    pub fn available_currency_codes(&self) -> ApiResult<Vec<String>> {
-        let exchange = self.fetch_exchange()?;
-        Ok(exchange
-            .rates
-            .map_or(vec![], |s| s.keys().cloned().collect::<Vec<_>>()))
-    }
-
-    fn fetch_exchange(&self) -> ApiResult<Exchange> {
-        let url = format!("https://api.exchangeratesapi.io/latest?base=USD");
-        let body = self.cache.request_cached(
-            self.client,
-            &url,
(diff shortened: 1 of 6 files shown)


src/providers/info.rs
```diff
@@ -82,16 +68,90 @@ pub struct TokenInfo {
...
+
+    fn populate_token_cache(&mut self) -> Result<()> {
+        let url = format!("{}/v1/tokens/?limit=10000", base_transaction_service_url());
+        let response = self.client.get(&url).send()?;
+        let data: Page<TokenInfo> = response.json()?;
+        for token in data.results.iter() {
+            self.cache.create(
+                &format!("dip_ti_{}", token.address),
+                &serde_json::to_string(&token)?,
+                info_cache_duration(),
```
jpalvarezl, line 104: Very nice unfolding of the response into individual entries.

src/utils/cache.rs
```diff
@@ -38,6 +38,11 @@ impl Cache for ServiceCache {
 }
 
 pub trait CacheExt: Cache {
+    fn invalidate_caches(&self, key: &str) {
+        self.invalidate_pattern(&format!("c_resp_*{}*", &key));
+        self.invalidate_pattern(&format!("c_reqs_*{}*", &key));
```
jpalvarezl, line 43: Just as an idea, maybe we could have a bit of a bigger cache manager struct, that keeps track of all the constants the cache handles, then we could have more specific methods for invalidating specific parts of the cache.

jpalvarezl: I like this direction though, having a prefix that would allow distinguishing between cached request and responses.

Uxio0:
> Information is pulled from on-chain (@Uxio0 maybe we could add a query param, that disables the on-chain lookup?)

That could be a solution. I opened a ticket to track this https://github.com/gnosis/safe-transaction-service/issues/217

jpalvarezl:
https://redis.io/topics/lru-cache

@rmeissner The docs recommend `allkeys-lru` for most cases, they state: 

>In general as a rule of thumb:
> - Use the allkeys-lru policy when you expect a power-law distribution in the popularity of your requests, that is, you expect that a subset of elements will be accessed far more often than the rest. This is a good pick if you are unsure.
> - Use the allkeys-random if you have a cyclic access where all the keys are scanned continuously, or when you expect the distribution to be uniform (all elements likely accessed with the same probability).
> - Use the volatile-ttl if you want to be able to provide hints to Redis about what are good candidate for expiration by using different TTL values when you create your cache objects.


We should also get away from `volatile-*` as it may throw an error as stated here: 

> The policies volatile-lru, volatile-random and volatile-ttl behave like noeviction if there are no keys to evict matching the prerequisites.

`noeviction` behaves like so:
> noeviction: return errors when the memory limit was reached and the client is trying to execute commands that could result in more memory to be used (most write commands, but DEL and a few more exceptions).

jpalvarezl:
Additional information regarding configuring redis:

- We should set the `maxmemory` value to whatever devOps would think reasonable for the max to be used by our service.  (`CONFIG GET maxmemory` returns `0` in my local instance, which means unlimited) 

- We should set `maxmemory-policy` to probably `allkeys-random` (the default appears to be `noeviction` in my local machine)

src/providers/info.rs
```diff
@@ -139,6 +146,87 @@ impl InfoProvider for DefaultInfoProvider<'_> {
...
+        self.cache.create(
+            "dip_tcl",
+            "",
+            if result.is_ok() {
+                info_cache_duration()
+            } else {
+                10000
+            },
+        );
+        result
```
jpalvarezl, line 218: @rmeissner , I am not sure what's the purpose of `dip_tcl`, why not cache the result of `populate_token_Cache` request with the usual timeout, like we do? Is this so that we check regularly for the full list, but the individual items are checked rarely? 

I was thinking we could add a new payload to trigger this from the `/v1/hook/update/<token>` endpoint. Then we could have a crontab somewhere issuing the request.

jpalvarezl:
See the proposal for the new timeouts. I would notify devOps for the deployment and update the spec doc for the gw.

https://github.com/gnosis/safe-client-gateway/issues/278

jpalvarezl:
We would also have to agree in a cache configuration. Regarding `maxmemory` and `maxmemory-policy`.

src/providers/info.rs
rmeissner: because this would mean we cache the whole data twice and as it is quite some data I would avoid this.

src/providers/info.rs
```diff
@@ -138,6 +145,87 @@ impl InfoProvider for DefaultInfoProvider<'_> {
...
+        }
+        let result = self.populate_token_cache();
+        // If error we use a shorter cache timeout (we do not want to DoS our service in case of an error)
+        self.cache.create(
+            "dip_tcl",
+            "",
+            if result.is_ok() {
+                info_cache_duration()
+            } else {
+                10000
```
jpalvarezl, line 214: There is this hardcoded timeout that I am not sure what it would be. This would be the error timeout ? Maybe we can merge the other PR before continuing with this one. https://github.com/gnosis/safe-client-gateway/pull/282

rmeissner: yes this is the error timeout ... can be hardcoded for a staging version but as you mentioned once #282 We should update this (maybe add a TODO commennt?)

jpalvarezl: I've merged the error timeout issue, so I will just add it. We can adjust the values according to #278

src/utils/cache.rs
rmeissner: I tried to keep these keys already in the cache module. Should we extract them to constants?

jpalvarezl (approved):

PR #261, opened 2021-02-11
Funtionatlity-wise complete, needs unit tests

### LINKED ISSUE #255: Adding CancelTransaction variant for transaction related endpoints
## Feature

Currently the web app classifies cancelation transactions on its own. It would be good to push this logic to the gateway, so clients can benefit of not having to implemente this themselves, while at the same time having a more uniform data representation across apps.

## Proposal 1

```rust
pub enum TransactionInfo {
    Transfer(Transfer),
    SettingsChange(SettingsChange),
    Custom(Custom),
    Creation(Creation),
    Cancellation(Cancellation), // new
    Unknown,
}

pub struct Cancellation {
  // TBD: web does not require fields for the UI but the clients might
}
```

## Proposal 2

```rust
pub struct Custom {
    pub to: String,
    pub data_size: String,
    pub value: String,
    pub method_name: Option<String>,
    pub action_count: Option<usize>,
    pub to_info: Option<AddressInfo>,
    pub is_cancellation: bool, // new
}
```
## Endpoints

Any endpoint returning transactions. That would be:
- `v1/safes/<address>/transactions`
- `v1/safes/<address>/transactions/history`
- `v1/safes/<address>/transactions/queued`

## Resources

Currently the web app identifies cancellation transactions using the following criteria:
https://github.com/gnosis/safe-react/blob/d17da5489b595385544fea426fcabadb40e7a168/src/logic/safe/store/actions/transactions/utils/transactionHelpers.ts#L52-L86

In `Suggested Fix` section the classification heuristic is defined in steps: https://github.com/gnosis/safe-react/issues/1800

Using fields from the transactions served by the gw:

```js
export const isCancelTransaction = ({ txInfo, safeAddress }: { txInfo: Custom; safeAddress: string }): boolean =>
  sameAddress(txInfo.to, safeAddress) &&
  sameString(txInfo.dataSize, '0') &&
  sameString(txInfo.value, '0') &&
  txInfo.methodName === null
```

jpalvarezl:
Closes #255 

Changes:
- Added `is_cancellation` flag to `Custom` transaction variant
- Unit tests


src/models/converters/transactions/mod.rs
```diff
@@ -191,12 +191,19 @@ impl MultisigTransaction {
...
+        let data_size = data_size(&self.data).to_string();
+        let value = self.value.as_ref().unwrap().to_owned();
+        let method_name = self.data_decoded.as_ref().map(|it| it.method.to_owned());
         Custom {
             to: self.to.to_owned(),
             to_info: info_provider.address_info(&self.to).ok(),
-            data_size: data_size(&self.data).to_string(),
-            value: self.value.as_ref().unwrap().into(),
-            method_name: self.data_decoded.as_ref().map(|it| it.method.to_owned()),
+            is_cancellation: &self.to == &self.safe
```
rmeissner, line 200: We need to check all params of thebsafe tx. See current implementation from web linked in the issue

rmeissner (changes requested):

src/models/converters/transactions/mod.rs
```diff
@@ -225,6 +232,7 @@ impl ModuleTransaction {
                 .data_decoded
                 .as_ref()
                 .and_then(|it| it.get_action_count()),
+            is_cancellation: false, // assumption: Module transactions can't be cancellation txs
```
rmeissner, line 235: That is correct (as module tx don't adjust the nonce). I would remove the comment or add this

src/models/converters/transactions/tests/summary.rs
```diff
@@ -84,6 +84,7 @@ fn module_tx_to_summary_transaction() {
             method_name: None,
             action_count: None,
             to_info: None,
+            is_cancellation: false, //module transactions are assumed to never be cancellation transactions
```
rmeissner, line 87: See above

src/models/converters/transactions/mod.rs
```diff
@@ -211,6 +212,34 @@ impl MultisigTransaction {
...
+                .map_or(true, |base_gas| base_gas.eq(&0))
+            && self
+                .gas_price
+                .as_ref()
+                .map_or(true, |gas_price| gas_price == "0")
+            && self.gas_token.as_ref().map_or(true, |gas_token| {
+                gas_token.len() < 2 || gas_token == "0x0000000000000000000000000000000000000000"
+            })
+            && self.refund_receiver.as_ref().map_or(true, |gas_token| {
+                gas_token.len() < 2 || gas_token == "0x0000000000000000000000000000000000000000"
```
rmeissner, line 235: I would reanme `gas_token` to `refund_receiver` in the closure

src/models/converters/transactions/mod.rs
```diff
@@ -211,6 +212,34 @@ impl MultisigTransaction {
...
+            && self
+                .base_gas
+                .as_ref()
+                .map_or(true, |base_gas| base_gas.eq(&0))
+            && self
+                .gas_price
+                .as_ref()
+                .map_or(true, |gas_price| gas_price == "0")
+            && self.gas_token.as_ref().map_or(true, |gas_token| {
+                gas_token.len() < 2 || gas_token == "0x0000000000000000000000000000000000000000"
```
rmeissner, line 232: why `< 2`? Actually I would say the equals is enough. I know web regards `0x` as a valid address ... but it is actually not :P

src/models/converters/transactions/mod.rs
```diff
@@ -211,6 +212,34 @@ impl MultisigTransaction {
...
+            })
+            && self.refund_receiver.as_ref().map_or(true, |gas_token| {
+                gas_token.len() < 2 || gas_token == "0x0000000000000000000000000000000000000000"
+            })
+            && self
```
rmeissner, line 237: I would remove this ... actually we don't care about the data decode (it has no data ... so it should never have a data_decoded in the first placE)

src/models/converters/transactions/mod.rs
```diff
@@ -211,6 +212,34 @@ impl MultisigTransaction {
             .flatten()
             .unwrap_or(0)
     }
+
+    fn is_cancellation(&self) -> bool {
```
rmeissner, line 216: Should we add a follow up issue for unit testing this method?

rmeissner (approved):

src/models/converters/transactions/mod.rs
jpalvarezl: ah yeah, that's what I was afraid. Cool I will only the equals then :)

src/models/converters/transactions/mod.rs
jpalvarezl: Yeah, I will do that, it would be nice to have this deployed now and write the tests througout the day :)

https://github.com/gnosis/safe-client-gateway/issues/264

PR #267, opened 2021-02-11
Implemented suggested changes by team

### LINKED ISSUE #266: Improvements to transaction propose endpoint
We should double check which fields are mandatory and optional. 

Initial feedback:
- [x] rename `contractTransactionHash` to `safeTxHash`
- [x] make `data`, `refundReceiver`  and `origin` optional

jpalvarezl:
Closes #266 

Changes:
- `data`, `refundReceiver` and `origin` are nullable
- renamed `contract_transaction_hash` to `safe_tx_hash`


src/models/service/transactions/requests.rs
```diff
@@ -12,15 +12,16 @@ pub struct ConfirmationRequest {
...
     pub nonce: String,
     pub operation: Operation,
     pub safe_tx_gas: String,
     pub base_gas: String,
     pub gas_price: String,
     pub gas_token: String,
-    pub refund_receiver: String,
-    pub contract_transaction_hash: String,
+    pub refund_receiver: Option<String>,
+    pub safe_tx_hash: String,
```
rmeissner, line 23: Do the core service accept this?  else you need to annotate with serde that for serializing it, it should use a different key

jpalvarezl: Good catch

rmeissner: cc @Uxio0

jpalvarezl: I verified it using the kotlin example we wrote some time ago: https://safe-team-rinkeby.staging.gnosisdev.com/app/#/safes/0xd6f5Bef6bb4acD235CF85c0ce196316d10785d67/transactions

rmeissner (approved):

PR #274, opened 2021-02-17
Rejection list for transactions

### LINKED ISSUE #258: Add "Already rejected" flag to transaction details
## Feature

In the proper rejection feature, we need to know that the transaction was already rejected in order to disable rejection feature. otherwise user will be able to submit multiple cancelation for one tx.

- First option: Add "isRejected" flag to transaction details response 
- Second option: Return "rejections" as list of signatures same as "confirmations", this will be helpful later if we wanted to show who rejected the tx in timeline box. 

**Edit by @jpalvarezl **

## Proposal

Modify transaction details to contain the list of rejections in the `MultisigExecutionDetails` struct, like so:

```rust
pub struct MultisigExecutionDetails {
    pub submitted_at: i64,
    pub nonce: u64,
    // .... 
   pub confirmations: Vec<MultisigConfirmation>,
    #[serde(skip_serializing_if = "Option::is_none")]
    pub rejectors: Option<Vec<String>>, //new
   //...
}
```

jpalvarezl:
Closes #258 

Changes:
- `MultisigExecutionDetails` now include a list of `rejectors`
- `SafeTxHash` calculation exclussively for cancellation works (Tests assume `safe_tx_gas == 0`, therefore test data generated with the current web app will not match the resulting hash)
- Made extensive use of `ethabi::encode` where possible (ERC191 bytes were padded when added there, so they had to be pushed directly into the vector, is this what encoded `packed` would have solved?)
- Added `to_hex_string` macro, I would like add `to_checksummed_hex_string` at some point. WDYT, @rmeissner ?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,6 +16,8 @@ reqwest = { version = "0.11.0", features = ["blocking", "json"] }
 
 # Ethereum types does not support checksummed addresses
 # ethereum-types= { version = "0.8.0", features = ["serialize"]}
+ethcontract-common = "0.11.1"
+ethabi = "12.0.0"
 
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,8 +14,6 @@ rocket = "0.4.5"
 rocket_codegen = "0.4.5"
 reqwest = { version = "0.11.0", features = ["blocking", "json"] }
 
-# Ethereum types does not support checksummed addresses
-# ethereum-types= { version = "0.8.0", features = ["serialize"]}
 ethcontract-common = "0.11.1"
 ethereum-types = { version = "0.9.2", features = ["serialize"]}
 ethabi = "12.0.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,6 +17,7 @@ reqwest = { version = "0.11.0", features = ["blocking", "json"] }
 # Ethereum types does not support checksummed addresses
 # ethereum-types= { version = "0.8.0", features = ["serialize"]}
 ethcontract-common = "0.11.1"
+ethereum-types = { version = "0.9.2", features = ["serialize"]}
 ethabi = "12.0.0"
 
 serde = { version = "1.0", features = ["derive"] }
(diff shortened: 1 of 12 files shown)


jpalvarezl:
Not sure if all the information regarding the confirmations for a cancellation is necessary. Just left it in, but it could very well just be a `Vec<String>`

rmeissner:
A rejection as a predetermined safetxhash for a specific nonce (as all parameters are defined), so there is no need to fetch all transactions at that nonce. You should calculate the safetxhash try to load that specific transaction and use the information of this transaction to render out the rejection information.

src/utils/transactions.rs
```diff
@@ -0,0 +1,106 @@
...
+
+pub const ERC191_BYTE: &'static str = "19";
+pub const ERC191_VERSION: &'static str = "01";
+
+pub fn fetch_rejections(context: &Context, safe_address: &str, nonce: u64) -> Option<Vec<String>> {
+    let safe_address: Address =
+        serde_json::from_value(serde_json::value::Value::String(safe_address.to_string())).unwrap();
+
+    let safe_tx_hash = hash(safe_address, nonce).to_vec();
+    let safe_tx_hash = to_hex_string!(safe_tx_hash);
```
rmeissner, line 24: this shadowing feels weird (especially since the variable is redeclared)

Cargo.toml
```diff
@@ -16,6 +16,9 @@ reqwest = { version = "0.11.0", features = ["blocking", "json"] }
 
 # Ethereum types does not support checksummed addresses
 # ethereum-types= { version = "0.8.0", features = ["serialize"]}
```
rmeissner, line 18: Should we remove these comments?

src/utils/transactions.rs
```diff
@@ -0,0 +1,106 @@
...
+        ethabi::Token::Uint(Uint::zero()),                     //gas_price
+        ethabi::Token::Address(Address::zero()),               //gas_token
+        ethabi::Token::Address(Address::zero()),               //refund_receiver
+        ethabi::Token::Uint(Uint::from(nonce)),                //base_gas
+    ]);
+
+    keccak256(encoded_parts)
+}
+
+fn fetch_cancellation_tx(context: &Context, safe_tx_hash: String) -> Option<MultisigTransaction> {
```
rmeissner, line 88: I would comment somewhere that we don't throw on error, but that we silently fail with `None`

rmeissner (approved):

Issue #278, opened 2021-02-18
Evaluate and increase timeout for requests


jpalvarezl:
- [x] Compile cache durations and check the current state. 

**edit**
`SAFE_APP_MANIFEST_CACHE_DURATION` is not defined in the spec doc of the gateway, it is therefore using the default value.

- Defaults: (as per the spec doc and the code defaults)
```
INFO_CACHE_DURATION = 60 * 15 (15 minutes)
EXCHANGE_API_CACHE_DURATION = 60 * 60 * 12 (12 hours)
REQUEST_CACHE_DURATION" = 60 * 15 (15 minutes)
ABOUT_CACHE_DURATION =  REQUEST_CACHE_DURATION
SAFE_APP_MANIFEST_CACHE_DURATION =  60 * 60 * 24 * 7 (1 week)
```

- `Staging` is using the default values written about ☝️ 

- `Production` is using the default for `EXCHANGE_API_CACHE_DURATION` and `SAFE_APP_MANIFEST_CACHE_DURATION`. The rest of the values are defined as following: (as per the spec doc)
 
```
INFO_CACHE_DURATION =  =  60 * 60 * 24 * 7 (1 week)
REQUEST_CACHE_DURATION" =  60 * 60 * 24 * 7 (1 week)
ABOUT_CACHE_DURATION =  60 * 10 (10 minutes) 
```

- [x] Draft a proposal for new timeouts

Get rid of `SAFE_APP_MANIFEST_CACHE_DURATION` and use `INFO_CACHE_DURATION` in its place. Proposed values:

```
EXCHANGE_API_CACHE_DURATION = 60 * 60 * 12 (12 hours)
REQUEST_CACHE_DURATION" = 60 * 60 * 24 * 7 (1 week)
INFO_CACHE_DURATION = 60 * 60 * 24 * 7 (1 week)
ABOUT_CACHE_DURATION =  60 * 10
REQS_ERROR_CACHE_DURATION = 60 (1 minute)
INFO_ERROR_CACHE_DURATION = 60 * 60 * 24 (1 day) # it used to be 10000 in the TokenInfo PR, maybe we should reduce to ~ 3 hours to match?
```

- [ ] Notify devOps requesting the changes

jpalvarezl:
[...]

PR #282, opened 2021-02-22
Info and Request error cache timeout added

### LINKED ISSUE #246: Cache error duration

### LINKED ISSUE #281: Noisy error alerts for services
Relates to https://github.com/gnosis/safe-client-gateway/issues/269

Issue #269 triggers a 500 in the gateway even though the core services report it as a 404. 

We should:
- Pipe 404s (maybe all 4xx) through to the clients
- Adjust our custom responder to not log 4xx

jpalvarezl:
Closes #246 

Changes:
- Added `REQS_ERROR_CACHE_DURATION` and `INFO_ERROR_CACHE_DURATION` env vars
- Default is 1 minute and 24 hours respectively
- Only 4xx errors are cached (old behaviour is preserved)

### DIFF
--- a/src/config/mod.rs
+++ b/src/config/mod.rs
@@ -58,7 +58,7 @@ pub fn internal_client_connect_timeout() -> u64 {
 }
 
 pub fn client_error_cache_duration() -> usize {
-    usize_with_default("CLIENT_ERROR_CACHE_DURATION", 3 * 10)
+    usize_with_default("CLIENT_ERROR_CACHE_DURATION", 60 * 3)
 }
 
 pub fn log_all_error_responses() -> bool {
--- a/src/config/mod.rs
+++ b/src/config/mod.rs
@@ -57,6 +57,10 @@ pub fn internal_client_connect_timeout() -> u64 {
     u64_with_default("INTERNAL_CLIENT_CONNECT_TIMEOUT", 1000)
 }
 
+pub fn client_error_cache_duration() -> usize {
+    usize_with_default("CLIENT_ERROR_CACHE_DURATION", 3 * 10)
+}
+
 pub fn log_all_error_responses() -> bool {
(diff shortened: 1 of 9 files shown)


src/utils/cache.rs
```diff
@@ -64,31 +64,40 @@ pub trait CacheExt: Cache {
...
+
                 let is_client_error = response.status().is_client_error();
                 let raw_data = response.text()?;
-                self.create(
-                    &url,
-                    CachedWithCode::join(status_code, &raw_data).as_str(),
-                    timeout,
-                );
-                return if is_client_error {
+
```
rmeissner, line 85: would it be nicer to put the `if` just around the timeout?

rmeissner (approved):

src/config/mod.rs
```diff
@@ -50,6 +50,14 @@ pub fn internal_client_connect_timeout() -> u64 {
     u64_with_default("INTERNAL_CLIENT_CONNECT_TIMEOUT", 1000)
 }
 
+pub fn request_error_cache_timeout() -> usize {
+    usize_with_default("REQS_ERROR_CACHE_DURATION", 60)
+}
+
+pub fn info_error_cache_timeout() -> usize {
+    usize_with_default("INFO_ERROR_CACHE_DURATION", 60 * 60 * 24)
```
rmeissner, line 58: Should we use the respective non-error timeout as a default? ... hmm maybe not ... not sure :D

jpalvarezl: We can tune this later I suppose :) I will document this timeouts in #278

src/utils/cache.rs
jpalvarezl: This is also necessary for the return value either `Err` or `Ok`, if I understood you correctly

PR #289, opened 2021-03-03
SafeInfo endpoint

### LINKED ISSUE #242: `SafeInfo` endpoint in the gateway including`AddressInfo`
## Feature 
`SafeInfo` endpoint in the gateway including`AddressInfo` for:
- `master_copy`
- `owners`
- `default fallback handler`
- `modules`

## Proposal

I suffixed the `struct` names with `Ex` as they are an expanded view of the data supplied by the backend. I am open to suggestions for alternatives.

```rust
pub struct SafeInfoEx {
    pub address: AddressEx, // or String if we don't think it's necessary
    pub nonce: u64,
    pub threshold: u64,
    pub owners: Vec<AddressEx>, // or Vec<String>, if deemed necessary
    pub implementation: AddressEx,
    pub modules: Option<Vec<AddressEx>>,
    pub fallback_handler: Option<AddressEx>,
    pub version: String,
}

pub struct AddressEx {
    pub value: String,
    pub name: Option<String>,
    pub logo_url: Option<String>,
}

```

All field names will be `camelCased` in the `json` response.
Network request for `AddressInfo` will not be issued for `address` and `owners` as usually this will not be a hit and it would be overridden by local data of the apps, however in the future we may want this.

jpalvarezl:
Closes #242 

Changes: 
- Added all the fields available in the core services to `SafeInfo`
- Added `SafeInfoEx` with extended `AddressInfo` for all the available fields
- Core business logic is there, we need to talk in refinement what we do with `owners`
- `AddressInfo` not fetched for `owners` and `address` fields


rmeissner (approved):
Not sure about the `ex` postfix :P (for the non Address stuff) ... but not really important

jpalvarezl:
It's very localised, we can changing it anytime without affecting the apps :) I am not sure what to name it tbh :/

PR #297, opened 2021-03-09
URL rules for `origin`

### LINKED ISSUE #275: `origin` can be used to trigger a call from the client gateway to anywhere
- Check if this is a security issue
  - Could you create a loop where the origin points to the safe transaction itself and therefore results in the interface calling itself? 
- In some cases this caused a blocking call (as we did not have a connection timeout) which at some point blocked the whole service. 
  - Switch to non blocking version of `reqwest`  

Solutions:
- Disallow `localhost` as source
- Disallow static IPs as source (aka require a dns entry)
- Only allow http protcols

jpalvarezl:
Closes #275 

Changes:
- Added dependency to `lazy_static` and `regex`
- Added rules to prevent injection of malicious URLs through origin field. The rules are
    - Only `http` or `https`
    - no `localhost`
    - no usage of IP addresses directly 

Open questions:
- Are we sure we want to disallow ipfs URLs ? (see the rules in the issue linked)
- Async `reqwest` would require using async `rocket`, right? Then I would prefer forking that effort for when we look into that for `celery` too.


src/utils/tests/urls.rs
```diff
@@ -0,0 +1,59 @@
...
+#[test]
+fn valid_url_no_trailing_slash() {
+    let input_url = "https://happy.path";
+
+    let actual = build_manifest_url(input_url).unwrap();
+    assert_eq!(actual, "https://happy.path/manifest.json")
+}
+
+#[test]
+fn valid_url_with_trailing_slash() {
```
rmeissner, line 54: Maybe a test with a port?

PR #302, opened 2021-03-10
Used correct invalidation patterns

### LINKED ISSUE #299: Cache invalidation via webhook takes too long and crashes service
After the last update the cache webhook time increased from ~20 seconds (what is already super bad) to 60+ seconds.

- On webhook we iterate all cache keys up to 4 times
- We have a lot of additional keys because of the token information

### Optimisation (ordered by complexity)
- Use 1 pattern that covers `CACHE_RESP_PREFIX` and `CACHE_REQS_PREFIX` (if required change the prefix)
- Use `HSET` for token information -> All token information are tied to 1 key

jpalvarezl:
Closes #299

### DIFF
--- a/src/providers/info.rs
+++ b/src/providers/info.rs
@@ -5,7 +5,7 @@ use crate::config::{
 };
 use crate::models::commons::Page;
 use crate::providers::address_info::{AddressInfo, ContractInfo};
-use crate::utils::cache::{Cache, CacheExt};
+use crate::utils::cache::{Cache, CacheExt, TOKEN_HASH};
 use crate::utils::context::Context;
 use crate::utils::errors::ApiResult;
 use crate::utils::json::default_if_null;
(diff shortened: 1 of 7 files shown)


src/utils/cache.rs
```diff
@@ -42,8 +42,7 @@ impl Cache for ServiceCache {
 
 pub trait CacheExt: Cache {
     fn invalidate_caches(&self, key: &str) {
-        self.invalidate_pattern(&format!("{}_*{}*", CACHE_RESP_PREFIX, &key));
-        self.invalidate_pattern(&format!("{}_*{}*", CACHE_REQS_PREFIX, &key));
+        self.invalidate_pattern(&format!("c*{}*", &key));
```
rmeissner, line 45: I would use `c_re??`

src/utils/cache.rs
```diff
@@ -160,3 +170,15 @@ fn pipeline_delete(con: &redis::Connection, keys: Iter<String>) {
     }
     pipeline.execute(con);
 }
+
+fn execute_hset(con: &redis::Connection, hash: &str, field: &str, value: &str) {
```
rmeissner, line 174: no need to write this yourself, they already exist: https://github.com/mitsuhiko/redis-rs/blob/master/src/commands.rs#L499

src/providers/info.rs
```diff
@@ -235,7 +232,7 @@ impl DefaultInfoProvider<'_> {
     // TODO: Check Eviction policies: https://redis.io/topics/lru-cache
     fn load_token_info(&mut self, token: &String) -> ApiResult<Option<TokenInfo>> {
         self.check_token_cache()?;
```
rmeissner, line 234: In the check function we don't need a separate key. We can just use `exists` https://github.com/mitsuhiko/redis-rs/blob/master/src/commands.rs#L361

rmeissner (approved):

PR #306, opened 2021-03-11
Use custom scan count

rmeissner:
- Allow to set the count for scan -> new `scan_match_count` function. 
- Introduce env variable for scan count -> `REDIS_SCAN_COUNT` (default `300`)

More info https://docs.keydb.dev/blog/2020/08/10/blog-post/

### DIFF
--- a/src/utils/cache.rs
+++ b/src/utils/cache.rs
@@ -1,7 +1,10 @@
+use crate::config::redis_scan_count;
 use crate::utils::errors::{ApiError, ApiResult};
 use mockall::automock;
 use rocket::response::content;
-use rocket_contrib::databases::redis::{self, pipe, Commands, Iter, PipelineCommands};
+use rocket_contrib::databases::redis::{
+    self, pipe, Commands, FromRedisValue, Iter, PipelineCommands, ToRedisArgs,
+};
 use serde::ser::Serialize;
 use serde_json;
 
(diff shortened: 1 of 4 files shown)


src/utils/cache.rs
```diff
@@ -186,7 +192,20 @@ fn pipeline_delete(con: &redis::Connection, keys: Iter<String>) {
...
+    pattern: P,
+    count: C,
+) -> redis::RedisResult<redis::Iter<'a, RV>> {
+    redis::cmd("INFO")
+        .cursor_arg(0)
+        .arg("MATCH")
+        .arg(pattern)
+        .arg("COUNT")
+        .arg(count)
+        .iter(con)
```
jpalvarezl, line 206: you could assign this to `let count : usize = redis::cmd("INFO")....unwrap()` so you don't have to use the internal result type of redis. You can go as far as coercing the type to `Option<usize>` if you end the comand with `ok()`

src/utils/cache.rs
```diff
@@ -54,7 +57,10 @@ impl Cache for ServiceCache {
     }
 
     fn invalidate_pattern(&self, pattern: &str) {
-        pipeline_delete(self, self.scan_match(pattern).unwrap());
+        pipeline_delete(
+            self,
+            scan_match_count(self, pattern, redis_scan_count()).unwrap(),
```
jpalvarezl, line 62: The `unwrap` could be moved to `scan_match_count` or not... I was thinking yesterday if it was good to panic here (I mean in this module in general). Probably yes, right?

jpalvarezl (approved):
I approve but I would really prefer if you didn't use `redis::RedisResult<...>` for `scan_match_count`

PR #317, opened 2021-03-22
Feature/issue 292 decouple address info

### LINKED ISSUE #292: AddressInfo: decouple TokenInfo and ContractInfo request
## Context

Currently the `address_info(address: &str)` fetches `TokenInfo` first and upon failure it fetches the `ContractInfo`. 

Certain cases (`confirmations` , `rejector`, `SafeInfo::address` or `SafeInfo::owners` addresses) are almost certainly not tokens and in rare cases could be found in `ContractInfo` endpoint

## Feature

Separate inside `address_info(...)` method the logic for fetching `TokenInfo` and `ContractInfo`  and create convenience methods for `full_search` (looking in both sources) or `shallow_search` (only for checking `ContractInfo`).

jpalvarezl:
Closes #292 

Changes:
- Added a default implementation method in `InfoProvider` that looks up the `tokenInfo` and `contractInfo` 
- `address_info()` function now only checks the `contracts` endpoint
- Adjusted tests and mocks accordingly

### DIFF
--- a/src/providers/info.rs
+++ b/src/providers/info.rs
@@ -82,6 +82,17 @@ pub trait InfoProvider {
     fn token_info(&mut self, token: &str) -> ApiResult<TokenInfo>;
     fn safe_app_info(&mut self, url: &str) -> ApiResult<SafeAppInfo>;
     fn address_info(&mut self, address: &str) -> ApiResult<AddressInfo>;
+
+    fn full_address_info_search(&mut self, address: &str) -> ApiResult<AddressInfo> {
+        self.token_info(&address)
+            .map(|it| AddressInfo {
+                name: it.name,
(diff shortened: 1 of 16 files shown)


src/providers/info.rs
```diff
@@ -129,33 +140,26 @@ impl InfoProvider for DefaultInfoProvider<'_> {
     }
 
     fn address_info(&mut self, address: &str) -> ApiResult<AddressInfo> {
```
rmeissner, line 142: should we call this `contract_info` to be more explicit?

rmeissner (approved):

src/providers/info.rs
jpalvarezl: will do :+1:

PR #320, opened 2021-03-24
Script to populate cache

### LINKED ISSUE #313: Add script put some load on staging
- Load top 300 safes from /analytics/multisig-transactions/by-safe/
- Trigger loading for assets, queued and history for them

jpalvarezl:
Closes #313 

Includes a script that fetches top 300 safes and fetches `balances`, `collectibles` , `transactions/history` and `transactions/queued` to warm up the cache.


scripts/load_tester/drill_config.yml
```diff
@@ -0,0 +1,24 @@
+concurrency: 4
+base: 'http://localhost:8000'
+iterations: 5
```
rmeissner, line 3: I would do iteration 1 (as we do not want to do load testing, just fill the cache)

scripts/load_tester/drill_config.yml
```diff
@@ -0,0 +1,24 @@
...
+
+plan:
+  - name: Fetch balances
+    request:
+      url: /v1/safes/{{ item.safe_address }}/balances/USD
```
rmeissner, line 8: I would load the transaction lists too, as these are the ones that really put entries into the cache

rmeissner (commented):
The current implementation is ok, but I should have been more specific on the purpose here.

So this is not primarily meant to test the local instance, but to fill the cache of the staging instance of the service. We want to do that to see how the backend behaves when the redis has more keys, as this caused issues in the past.

Therefore it is also not required to actually run all kind of requests in parallel (as the purpose is cache population, not load testing).

jpalvarezl:
Agreed. I was just surprised to see this error a lot:

```
Error connecting 'http://localhost:8000/v1/safes/...../transactions/history': reqwest::Error { kind: Request, url: Url { scheme: "http", username: "", password: None, host: Some(Domain("localhost")), port: Some(8000), path: "/v1/safes/....../transactions/history", query: None, fragment: None }, source: hyper::Error(IncompleteMessage) }
```

I am wondering there is something we can do or is it simply the script dropping the connection too quickly?

scripts/load_tester/start.py
```diff
@@ -0,0 +1,47 @@
...
+
+check_service()
+safes = load_safes()
+print("Top 300 safes:")
+print("\n\t" + "\n\t".join(safes))
+print("Safes ready ... ")
+print("Populating cache ...")
+
+for safe in safes:
+    balance_response = requests.get("http://localhost:8000/v1/safes/" + safe + "/balances/USD")
```
rmeissner, line 34: As the purpose is to populate our **staging** cache we shouldnot use localhost by default

rmeissner (changes requested):

scripts/load_tester/start.py
jpalvarezl: I thought it was meant for local testing. I will change this.

jpalvarezl: I will add more logic to determine the environment, as the top 300 safes are fetched from the `.env` core services url

jpalvarezl: meaning I will `sys.exit` if you try to fetch from production envs

rmeissner: I would not limit it to any specific env, but the default should be staging (at the minimum it should not be hardcoded)

jpalvarezl: > As the purpose is to populate our staging cache we shouldnot use localhost by default

I thought it was just for staging. I will leave it to work on any network or environment.

README.md
```diff
@@ -58,3 +58,14 @@ Additional NOTE: the `structs` defined in this project match those in staging. T
 To run all tests use the `cargo test` command. If you want to run a specific subset of tests, then add additionally any info regarding the path of the tests and `cargo` will match it.
 
 Example: `cargo test converters` will run every tests under the `converters` module. Matching occurs also at a test name level, so by writing the full name of a test, that single test can be run.
+
+Additionally, for cache testing, we have included a script that fills up the cache as it would happen in production. You can find the script in `./scripts/load_tester/start.py`. To run the script, use the following commands: 
+
+```shell
+$ alias python=python3
```
rmeissner, line 65: I would simplify these steps:
- Remove `$` (so that I can copy paste full line)
- Skip python alias ... not really required
- Skip `stat` step, just directly call `python3 -m venv venv`

scripts/load_tester/start.py
```diff
@@ -0,0 +1,48 @@
...
+
+base_gateway_url = get_base_url()
+safes = load_safes()
+print("Top 300 safes:")
+print("\n\t" + "\n\t".join(safes))
+print("Safes ready ... ")
+print("Populating cache ...")
+
+for safe in safes:
+    balance_response = requests.get(base_gateway_url + "/v1/safes/" + safe + "/balances/USD")
```
rmeissner, line 35: https://realpython.com/python-string-formatting/

scripts/load_tester/start.py
```diff
@@ -0,0 +1,48 @@
...
+    top_safes_url = TX_SERVICE_URL + "/api/v1/analytics/multisig-transactions/by-safe/?limit=300"
+    print(top_safes_url)
+    response = requests.get(top_safes_url)
+    return list(map(lambda safe: safe['safe'], response.json()['results']))
+
+
+base_gateway_url = get_base_url()
+safes = load_safes()
+print("Top 300 safes:")
+print("\n\t" + "\n\t".join(safes))
```
rmeissner, line 30: Is it really helpful to pring all 300 safe addresses? I would then probably rather have 1 log statement when we start requesting the data (e.g. first line in the for loop)

scripts/load_tester/start.py
```diff
@@ -0,0 +1,48 @@
...
+print("Safes ready ... ")
+print("Populating cache ...")
+
+for safe in safes:
```
rmeissner, line 34: A little bit inconsistent that this logic is not in a method while all the other logic is in methods

rmeissner (approved):
I works and could be merged (as it fulfills its purpose), but I would appreciate some cleanup ;)

scripts/load_tester/start.py
jpalvarezl: nah, I just thought it was nice. I will get rid of it

scripts/load_tester/start.py
jpalvarezl: do you mean the logic inside the `for`? Is that what you would have extracted, or do you want me to have a method called `main` with all the lose function calls ?

rmeissner: I mean the whole for loop ;) ... not really important, it was more that the other to steps where in method so I thought it might be more consistent to have something like:
```python
base_gateway_url = get_base_url()
safes = load_safes()
load_information_for_safes(safes)
```

scripts/cache_warmer/start.py
```diff
@@ -0,0 +1,47 @@
...
+        return "https://safe-client-rinkeby.staging.gnosisdev.com" if "rinkeby" in TX_SERVICE_URL \
+            else "https://safe-client-mainnet.staging.gnosisdev.com"
+
+
+def load_safes() -> list[str]:
+    print("Loading remote ...")
+    top_safes_url = TX_SERVICE_URL + "/api/v1/analytics/multisig-transactions/by-safe/?limit=300"
+    print(top_safes_url)
+    response = requests.get(top_safes_url)
+    return list(map(lambda safe: safe['safe'], response.json()['results']))
```
Uxio0, line 24: `return [safe['safe'] for safe in response.json()['results']]` would be the preferred way to do that (list comprehension)

scripts/cache_warmer/start.py
```diff
@@ -0,0 +1,47 @@
...
+def get_base_url() -> str:
+    if "staging" not in TX_SERVICE_URL:
+        return "https://safe-client.rinkeby.gnosis.io" if "rinkeby" in TX_SERVICE_URL \
+            else "https://safe-client.mainnet.gnosis.io"
+    else:
+        return "https://safe-client-rinkeby.staging.gnosisdev.com" if "rinkeby" in TX_SERVICE_URL \
+            else "https://safe-client-mainnet.staging.gnosisdev.com"
+
+
+def load_safes() -> list[str]:
```
Uxio0, line 19: If you use `list[str]` it will require Python 3.9. If you want to support Python 3.8 (more extended), you should better use `List[str]` (and `List` should be imported from `typing`)

scripts/cache_warmer/start.py
```diff
@@ -0,0 +1,47 @@
...
+        print(PRINT_FORMAT.format(str(collectibles_response.elapsed.total_seconds()), str(
+            collectibles_response.status_code), collectibles_response.url))
+        print(PRINT_FORMAT.format(str(tx_queued_response.elapsed.total_seconds()), str(
+            tx_queued_response.status_code), tx_queued_response.url))
+        print(PRINT_FORMAT.format(str(tx_history_response.elapsed.total_seconds()), str(
+            tx_history_response.status_code), tx_history_response.url))
+        print()
+
+
+url = get_base_url()
```
Uxio0, line 45: This lines should be inside an `if __name__ == '__main__':` block, so if you want to import any function from this script it doesn't get executed

PR #323, opened 2021-03-29
Added sort by fiat balance to endpoint

### LINKED ISSUE #318: Sort assets by fiat value
- On safe-react we will disable being able to manually sort the asset list with the next release, purely relying on the client-gateway sorting.
- We currently sort the list alphabetically
- To evaluate whether this is the best approach, I ran a quick poll on Twitter, showing a rather clear picture
image

- Sort asset list in client-gateway by fiat value to show bigger portfolio positions at the top.

- We might at some point re-introduce manual sorting in some form: https://github.com/gnosis/safe/issues/381 but the default sorting should stick with what would be the best solution for most users. (which according to the poll above seems to be sorting by fiat value)

jpalvarezl:
Closes #318 

https://stackoverflow.com/a/60916195/7721253 => Benchmark of possible ways to solve this

Sorting by fiat balance for the `balances` endpoint.

### DIFF
--- a/src/services/balances.rs
+++ b/src/services/balances.rs
@@ -36,7 +36,7 @@ pub fn balances(
 
     let mut total_fiat = 0.0;
 
-    let service_balances: Vec<Balance> = backend_balances
+    let mut service_balances: Vec<Balance> = backend_balances
         .into_iter()
         .map(|it| {
             let balance = it.to_balance(usd_to_fiat);
@@ -45,6 +45,14 @@ pub fn balances(
         })
         .collect();
 
+    service_balances.sort_by(|a, b| {
+        a.fiat_balance
+            .parse::<f64>()
+            .unwrap_or(0.0)
+            .partial_cmp(&b.fiat_balance.parse::<f64>().unwrap_or(0.0))
+            .unwrap()
+    });
+    service_balances.reverse();
     Ok(Balances {
         fiat_total: total_fiat.to_string(),
         items: service_balances,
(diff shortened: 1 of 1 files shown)


src/services/balances.rs
```diff
@@ -45,6 +46,14 @@ pub fn balances(
         })
         .collect();
 
+    service_balances.sort_by(|a, b| {
+        a.fiat_balance
+            .parse::<f64>()
```
rmeissner, line 51: Any issues here, that value might be to large or have too many decimals?

jpalvarezl: I would have to add unit tests for this parse method. As we use it in this same file too in line 44.

rmeissner: I see we fallback to 0 ... so doesn't matter for now :)

jpalvarezl: Not sure what the alternative would be however.

src/services/balances.rs
```diff
@@ -45,6 +46,14 @@ pub fn balances(
...
+            .unwrap_or(0.0)
+            .partial_cmp(&b.fiat_balance.parse::<f64>().unwrap_or(0.0))
+            .unwrap_or(Ordering::Equal)
+    });
+    service_balances.reverse();
```
rmeissner, line 56: Wouldn't it be easier to just swap `a` and `b` in the `sort_by`?

src/services/balances.rs
jpalvarezl: Should we introduce something like this: https://docs.rs/rug/0.6.0/rug/ ?

src/services/balances.rs
jpalvarezl: absolutely, well spotted.

rmeissner (approved):

PR #326, opened 2021-03-30
Added env for native coin

### LINKED ISSUE #179: Balances endpoint in non ETH networks returns ETH for native currency balance
## Feature

When the `tokenAddress` is `None` for `mainnet` and `rinkeby` is assumed to be ETH however for xDai and other networks it is not. The gateway sends along `Balance` items the `TokenInfo` for which the `name` and the `symbol` are not nullable and therefore unwrapped as ETH. 

## Solution
We will add the symbol, name and decimals to the env and load them from there when mapping
https://github.com/gnosis/safe-client-gateway/blob/main/src/models/converters/balances.rs#L21-L31

ENV var names:

- `NATIVE_COIN_DECIMALS` : `usize` (Default: `18`) 
- `NATIVE_COIN_SYMBOL` : `String` (Default: `ETH`) 
- `NATIVE_COIN_NAME`: `String` (Default: `Ether`)

jpalvarezl:
Closes #179 

Changes:
- Added `NATIVE_COIN_NAME`, `NATIVE_COIN_SYMBOL` and `NATIVE_COIN_DECIMALS` for balances
- **caveat:**  there is still `TokenType::Ether`

### DIFF
--- a/src/models/converters/balances.rs
+++ b/src/models/converters/balances.rs
@@ -1,9 +1,11 @@
 use crate::models::backend::balances::Balance as BalanceDto;
 use crate::models::service::balances::Balance;
 use crate::providers::info::{TokenInfo, TokenType};
+use crate::config::{native_coin_decimals, native_coin_symbol, native_coin_name};
 
 impl BalanceDto {
     pub fn to_balance(&self, usd_to_fiat: f64) -> Balance {
+        log::error!("AM I even called?");
         let fiat_conversion = self.fiat_conversion.parse::<f64>().unwrap_or(0.0) * usd_to_fiat;
         let fiat_balance = self.fiat_balance.parse::<f64>().unwrap_or(0.0) * usd_to_fiat;
         let token_type = self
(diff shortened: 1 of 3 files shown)


src/models/converters/balances.rs
```diff
@@ -1,9 +1,11 @@
 use crate::models::backend::balances::Balance as BalanceDto;
 use crate::models::service::balances::Balance;
 use crate::providers::info::{TokenInfo, TokenType};
+use crate::config::{native_coin_decimals, native_coin_symbol, native_coin_name};
 
 impl BalanceDto {
     pub fn to_balance(&self, usd_to_fiat: f64) -> Balance {
+        log::error!("AM I even called?");
```
rmeissner, line 8: please remove this :)

rmeissner (changes requested):

src/models/converters/balances.rs
jpalvarezl: oops forgot to get rid of it before pushing :)

Issue #327, opened 2021-03-30
Timeout during loading txs list ( both Queue and History)


liliya-soroka:
**Describe the bug**
The timeout happens very when the Queue or History endpoint is called by the clients

**To Reproduce**
- add safe = 0x8675B754342754A30A2AeF474D114d8460bca19b
- open Queue or History on any client ( web app, desktop app, mobile apps)
Current result: Timeout error on mobile app
Long status or lists updating on web and desktop

**Expected behavior**
The lists should be loaded without timeout

**Desktop (please complete the following information):**
 - OS: both mobile clients
 - Browser chrome
 - Version 3.3.0/3.2.0

**Additional context**

PR #328, opened 2021-03-30
Added timeout and error caching

rmeissner:
Closes to #327

### DIFF
--- a/src/utils/cache.rs
+++ b/src/utils/cache.rs
@@ -124,7 +124,6 @@ pub trait CacheExt: Cache {
                     );
                 }
                 let response = client_response?;
-                log::debug!("response {}", response.status());
                 let status_code = response.status().as_u16();
 
                 // Early return and no caching if the error is a 500 or greater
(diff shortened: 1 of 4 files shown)


src/utils/cache.rs
```diff
@@ -96,22 +97,38 @@ pub trait CacheExt: Cache {
...
-                let response = client.get(url).send()?;
+                let mut request = client.get(url);
+                if request_timeout > 0 {
+                    request = request.timeout(Duration::from_millis(request_timeout));
+                }
+                let client_response = request.send();
+                if cache_all_errors && client_response.is_err() {
+                    self.create(
+                        &cache_key,
+                        CachedWithCode::join(500, "").as_str(),
```
jpalvarezl, line 122: Not sure what the empty string represents here. Is there something more meaningful we could put here for easier debugging in the future?

rmeissner: Probably the err object could be serialized

PR #337, opened 2021-04-01
USD fiat skip check against exchange API

### LINKED ISSUE #335: Optimise balance requests in USD
Currently the core services already return the fiat balances in USD, there is no need to interact with the exchange API in this case.

In such case set the exchange rate to `1.0` and avoid intereacting with the exchange API.

jpalvarezl:
Closes #335

### DIFF
--- a/src/services/balances.rs
+++ b/src/services/balances.rs
@@ -33,11 +33,7 @@ pub fn balances(
     let backend_balances: Vec<BalanceDto> = serde_json::from_str(&body)?;
 
     let info_provider = DefaultInfoProvider::new(&context);
-    let usd_to_fiat = if fiat.to_lowercase() == "usd" {
-        1.0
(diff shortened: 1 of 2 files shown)


src/services/balances.rs
```diff
@@ -33,7 +33,11 @@ pub fn balances(
     let backend_balances: Vec<BalanceDto> = serde_json::from_str(&body)?;
 
     let info_provider = DefaultInfoProvider::new(&context);
-    let usd_to_fiat = info_provider.exchange_usd_to(fiat).unwrap_or(0.0);
+    let usd_to_fiat = if fiat.to_lowercase() == "usd" {
```
rmeissner, line 36: I would put this in the info provider

Uxio0 (approved):

PR #341, opened 2021-04-08
Introduce DB ids for redis and builder pattern cache operations

### LINKED ISSUE #304: Use multiple redis databases
- Use different databases for general requests and information provider related requests
- Reponses should use the same database as gerneral requests

### Definitions
- Requests: gateway -> core
- Response: gateway-> client

### LINKED ISSUE #340: Add request timeouts to core service requests
There is already a `Connection` timeout for the `reqwest::client` and specifically for IPFS requests reads, as well.

Should we set the timeout [here](https://github.com/gnosis/safe-client-gateway/blob/0393127d68f59daf19d51fe2d4aa0f421f4dd516/src/main.rs#L39) not only for `Connection`? 

No, different endpoints require different timeouts.

## Proposal

- Transaction related endpoints: 30 [s]
- Safe Info: 10 [s]
- Token Info: 5 [s]
- Safe App Info: no change
- Default: 5[s] (for balances, collectibles, contract info, etc)

jpalvarezl:
Closes #304
Closes #340 

Next steps (not necessarily in order of priority):
- Use the `Database` field #344 
- Enforce definition of `RequestCached::request_timeout` for every usage #343  
- Move the `new` method of each builder to `Cache` and keep a reference to the cache within each `cache_operation` to be able to call execute without having to pass `Cache` #343


src/cache/cache_operations.rs
```diff
@@ -0,0 +1,131 @@
+use crate::config::{request_cache_duration, request_error_cache_timeout};
+use crate::utils::errors::ApiResult;
+use serde::Serialize;
+
+pub enum Database {
```
rmeissner, line 5: Why not use an integer?

rmeissner (commented):
I like the builder approach a lot, look way cleaner. 

Not sure about the enum for the database (I would have just used an index). Maybe just defining the index together with the enum would improve this.

src/cache/cache_operations.rs
jpalvarezl: I wanted to put a name with text in it so we remember what the integer stands anywhere we use it. I think we can assign constants to enums in rust like so: 
```rust
pub enum Database {
    Info = 1,
    Default = 0,
}
```
I just need to double check how it behaves

rmeissner: Yeah I see the value of that. And I think with the explicit integer definition it would be very nice

jpalvarezl:
There are still a few warnings issued. That's because of the new database enum and fields that we will put to use once we merge the task forked for that.

jpalvarezl:
I removed `c_re` from the `c_re*{}*` invalidation pattern as I didn't see it adding any value :thinking: but I might be wrong. I don't remember exactly why it was there in the first place. Was it future proofing?

src/cache/cache_operations.rs
```diff
@@ -0,0 +1,151 @@
...
+    pub resp_generator: Box<dyn Fn() -> ApiResult<R> + 'a>,
+}
+
+impl<'a, R> CacheResponse<'a, R>
+where
+    R: Serialize,
+{
+    pub fn new() -> Self {
+        CacheResponse {
+            database: Database::Default,
```
rmeissner, line 57: I am a big hesitant about these defaults. I would probably work with `Option` and have them as default set to `None` and enforce on `execute` that all the required params are `Some`

src/providers/info.rs
```diff
@@ -1,3 +1,5 @@
+use crate::cache::cache::Cache;
```
rmeissner, line 1: cache cache cache :P ... would love a better naming, but not critical :P

src/cache/cache_operations.rs
```diff
@@ -0,0 +1,151 @@
...
+        self.cache_duration = cache_duration;
+        self
+    }
+
+    pub fn error_cache_duration(&mut self, error_cache_duration: usize) -> &mut Self {
+        self.error_cache_duration = error_cache_duration;
+        self
+    }
+
+    pub fn cache_all_errors(&mut self, cache_all_errors: bool) -> &mut Self {
```
rmeissner, line 139: Wouldn't it be nicer to have this without a param. E.g. you default to `false` and when you call this then it will change to `true` ... no way to toggle it back, because that is not a realistic use case

src/cache/cache_op_executors.rs
```diff
@@ -0,0 +1,105 @@
...
+        client,
+        &operation.url,
+        operation.cache_duration,
+        operation.error_cache_duration,
+        operation.cache_all_errors,
+        operation.request_timeout,
+    )
+}
+
+fn request_cached_priv(
```
rmeissner, line 47: Not really a fan of this name, what about renaming the the other method to `request_from_operation` or as this file is call `op_executor` maybe `execute_request_cached`? Could we inline this method? (not sure if this is nicer, so more this is not required ;) )

src/cache/cache_op_executors.rs
```diff
@@ -0,0 +1,105 @@
+use crate::cache::cache::Cache;
```
rmeissner, line 1: The name of this file ☝️ 🙈

src/cache/cache_operations.rs
```diff
@@ -0,0 +1,151 @@
...
+        self.cache_all_errors = cache_all_errors;
+        self
+    }
+
+    pub fn execute(
+        &self,
+        client: &reqwest::blocking::Client,
+        cache: &dyn Cache,
```
rmeissner, line 147: Hmm the `dyn` feels weird ... I know it is required because how it is stored in the `InfoProvider` struct. Primarily it feels weird, because we don't use dyn in the other case. I would love to know when to use what (https://doc.rust-lang.org/std/keyword.dyn.html is very short :D )

rmeissner (commented):
Looks good overall. 

I would like to handle the default and required values a little bit differently, but besides that (and some unimportant naming stuff) I think it looks good

src/services/hooks.rs
```diff
@@ -1,18 +1,27 @@
+use crate::cache::cache::Cache;
+use crate::cache::cache_operations::Invalidate;
 use crate::models::backend::webhooks::{Payload, PayloadDetails};
-use crate::utils::cache::{Cache, CacheExt};
 use crate::utils::errors::ApiResult;
 
 pub fn invalidate_caches(cache: &impl Cache, payload: &Payload) -> ApiResult<()> {
-    cache.invalidate_caches(&payload.address);
+    Invalidate::new()
+        .pattern(format!("c_re*{}*", &payload.address))
```
rmeissner, line 8: I don't like to much, that we have to expose this prefix. Could we have this somehow as part of the builder?

jpalvarezl: I was thinking the same. Would you oppose to having an enum for this? Something like 

```rust
pub enum InvalidationPatterns{
   requests_only, responses_only, requests_and_response, tokens, flush_all
}
```

I would add variants as we need them, but for now we would only need `requests_and_responses` and `flush_all`

jpalvarezl: We could also have a variant `Custom(String)` that allows passing a value or a pattern

src/cache/cache_operations.rs
jpalvarezl: They save up a lot of lines of code, I would say. I see your point though, but also in the case of the database `Database::Info` is for a very small subset of requests coming from a specific source file. I personally find it ok in this case.

src/providers/info.rs
jpalvarezl: Totally, I would rename to `redis`  as this is the redis specfiic impl of this cache, if you agree.

jpalvarezl: so it looks like `cache:;redis::Cache`

src/cache/cache_operations.rs
jpalvarezl: ah maybe I was being too strict with the builder pattern, I think that would be nice actually.

src/cache/cache_op_executors.rs
jpalvarezl: Yeah, I think that makes sense, as it's private anyway adds no value.

src/cache/cache_op_executors.rs
jpalvarezl: :see_no_evil:

src/cache/cache_operations.rs
jpalvarezl: So we agreed to keep it :)

src/cache/cache_operations.rs
```diff
@@ -0,0 +1,151 @@
...
+    pub error_cache_duration: usize,
+    pub cache_all_errors: bool,
+}
+
+impl RequestCached {
+    pub fn new() -> Self {
+        RequestCached {
+            database: Database::Default,
+            url: String::new(),
+            request_timeout: 10000,
```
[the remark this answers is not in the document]
jpalvarezl: I would solve this in #340 if that's ok with you? I can lower this value if you prefer.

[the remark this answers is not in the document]
rmeissner: I would have the proper handling of default values for the builders (see https://github.com/gnosis/safe-client-gateway/pull/341#discussion_r611530099)

src/cache/cache_operations.rs
rmeissner: but this will cause that you forget to set it and use the wrong one. While for the database at least right now we can obviously make it a default (as no other one is used) I would not set defaults for the other params (e.g. `key`, `duration` and `resp_generator`).

src/providers/info.rs
rmeissner: but the `Cache` trait is not redis specifc. I can implement that trait also with a `SQLite`.

I do agree that the `impl Cache for ServiceCache` is redis specific.

src/cache/cache_operations.rs
rmeissner: I think builder patterns are not that strict. E.g. OkHttp does this too afaik ... but yeah I am fine either way

src/services/hooks.rs
rmeissner: Makes sense (please upper case the enums :D). I would probably start out only with the ones we need ;) but yea long run we can then add more

src/cache/cache_operations.rs
rmeissner: As mentioned in our call on friday, for me the perfect builder would not compile if the required params are not set.

jpalvarezl: Agreed, I am not sure if I can do a compile time check yet. I remember agreeing on forking into issues, that i linked in the description. #343 

I will restore the original behaviour, which was setting the timeout to 0 and having a check in the method that would use the internal default of reqwest, as I thought that defining this values would be the focus of #340

src/providers/info.rs
jpalvarezl: I would ask you then to provide the suggestion for a better name as I can't think of a better option.

jpalvarezl: the only thing that isn't redis specific is the `Cache` trait, which I suppose I could move to `mod.rs` which would still result in `crate::cache::Cache`. If you are ok with that then I would:
- Move `trait Cache` to `mod.rs`
- Rename `cache.rs` to `redis.rs`

src/cache/cache_operations.rs
```diff
@@ -0,0 +1,156 @@
...
+
+pub enum InvalidationPattern {
+    FlushAll,
+    RequestsResponses(String),
+}
+
+impl Invalidate {
+    pub fn new() -> Self {
+        Invalidate {
+            pattern: InvalidationPattern::FlushAll,
```
rmeissner, line 26: This is not a good default pattern :D

src/cache/cache_op_executors.rs
```diff
@@ -0,0 +1,97 @@
...
+    cache: &dyn Cache,
+    client: &reqwest::blocking::Client,
+    operation: &RequestCached,
+) -> ApiResult<String> {
+    let cache_key = format!("{}_{}", CACHE_REQS_PREFIX, &operation.url);
+    match cache.fetch(&cache_key) {
+        Some(cached) => CachedWithCode::split(&cached).to_result(),
+        None => {
+            let mut request = client.get(&operation.url);
+            if operation.request_timeout > 0 {
```
rmeissner, line 53: Didn't you want to remove this?

src/cache/cache_operations.rs
```diff
@@ -0,0 +1,149 @@
...
+
+impl Invalidate {
+    pub fn new(pattern: InvalidationPattern) -> Self {
+        Invalidate {
+            pattern,
+            database: Database::Default,
+        }
+    }
+
+    pub fn pattern(&mut self, pattern: InvalidationPattern) -> &mut Self {
```
rmeissner, line 33: Not sure if we really need it (but also doesn't harm)

rmeissner (approved):

src/cache/cache_op_executors.rs
jpalvarezl: oops forgot, will do now

src/cache/cache_operations.rs
jpalvarezl: will kick it out as well, no point on having it :)

PR #342, opened 2021-04-09
Added config entries for request timeouts agains the core services

### LINKED ISSUE #340: Add request timeouts to core service requests

jpalvarezl:
Closes #340 
This PR will be addressed in #341 

- Introduced request time outs for requests against the core services
- Added specific values for certain endpoints (see #340 )

### DIFF
--- a/src/config/mod.rs
+++ b/src/config/mod.rs
@@ -87,7 +87,7 @@ pub fn safe_app_manifest_cache_duration() -> usize {
     usize_with_default("SAFE_APP_MANIFEST_CACHE_DURATION", indefinite_timeout())
 }
 
-// REQUEST TIMEOUTS
+// REQUEST TIMEOUTS in milliseconds
 pub fn internal_client_connect_timeout() -> u64 {
     u64_with_default("INTERNAL_CLIENT_CONNECT_TIMEOUT", 1000)
 }
--- a/src/config/mod.rs
+++ b/src/config/mod.rs
@@ -105,7 +105,7 @@ pub fn safe_info_request_timeout() -> u64 {
 }
 
 pub fn token_info_request_timeout() -> u64 {
-    u64_with_default("TOKEN_INFO_REQUEST_TIMEOUT", 5000)
+    u64_with_default("TOKEN_INFO_REQUEST_TIMEOUT", 30000)
 }
 
 pub fn default_request_timeout() -> u64 {
--- a/src/config/mod.rs
+++ b/src/config/mod.rs
@@ -96,6 +96,22 @@ pub fn safe_app_info_request_timeout() -> u64 {
     u64_with_default("SAFE_APP_INFO_REQUEST_TIMEOUT", 3000)
 }
 
+pub fn transaction_request_timeout() -> u64 {
+    u64_with_default("TRANSACTION_REQUEST_TIMEOUT", 30000)
+}
+
+pub fn safe_info_request_timeout() -> u64 {
+    u64_with_default("SAFE_INFO_REQUEST_TIMEOUT", 10000)
+}
+
+pub fn token_info_request_timeout() -> u64 {
+    u64_with_default("TOKEN_INFO_REQUEST_TIMEOUT", 5000)
+}
+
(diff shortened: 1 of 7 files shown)


src/utils/cache.rs
```diff
@@ -111,9 +111,8 @@ pub trait CacheExt: Cache {
             Some(cached) => CachedWithCode::split(&cached).to_result(),
             None => {
                 let mut request = client.get(url);
-                if request_timeout > 0 {
-                    request = request.timeout(Duration::from_millis(request_timeout));
-                }
+                request = request.timeout(Duration::from_millis(request_timeout));
```
jpalvarezl, line 114: removed the check against `0` as there is a default now

rmeissner: What happens if we pass in 0 now?

src/utils/cache.rs
```diff
@@ -164,7 +163,14 @@ pub trait CacheExt: Cache {
         cache_duration: usize,
         error_cache_duration: usize,
     ) -> ApiResult<String> {
-        self.request_cached_advanced(client, url, cache_duration, error_cache_duration, false, 0)
```
rmeissner, line 167: What are the flows where this method is still used?

Should we completely remove this method to enforce a timeout everywhere?

src/config/mod.rs
```diff
@@ -96,6 +96,22 @@ pub fn safe_app_info_request_timeout() -> u64 {
...
+pub fn transaction_request_timeout() -> u64 {
+    u64_with_default("TRANSACTION_REQUEST_TIMEOUT", 30000)
+}
+
+pub fn safe_info_request_timeout() -> u64 {
+    u64_with_default("SAFE_INFO_REQUEST_TIMEOUT", 10000)
+}
+
+pub fn token_info_request_timeout() -> u64 {
+    u64_with_default("TOKEN_INFO_REQUEST_TIMEOUT", 5000)
```
rmeissner, line 108: Actually might makes sense to up this to something like 10, 15 or 30

rmeissner (approved):

src/config/mod.rs
jpalvarezl: I will leave it as 30 to match the clients

src/utils/cache.rs
jpalvarezl: you can pass 0 and it will immediately timeout. I think that you used 0 in the fix for the safe app info to simply not set a request timeout. In this case, we are always setting it to `request_timeout` which should be the `DEFAULT_REQUEST_TIMEOUT = 5000`

src/utils/cache.rs
jpalvarezl: I would rather do this cleanup in the refactor task, to avoid double work.

rmeissner: But where do we use the "default" timeout now?

src/utils/cache.rs
```diff
@@ -164,7 +163,14 @@ pub trait CacheExt: Cache {
...
+        self.request_cached_advanced(
+            client,
+            url,
+            cache_duration,
+            error_cache_duration,
+            false,
+            default_request_timeout(),
```
jpalvarezl, line 172: https://github.com/gnosis/safe-client-gateway/pull/342/files#r610626821

Here

src/utils/cache.rs
jpalvarezl: line 172

jpalvarezl:
These changes will be addressed by #341

PR #351, opened 2021-04-23
`InfoProvider` no longer mutable. Restored closures

### LINKED ISSUE #350: Remove mutability of `InfoProvider`where possible and inline closures
Before having partial mutability, `InfoProvider` had to be passed around as a mutable reference for methods. Where we had code with closures like so:

```rust
// ...
some_option
   .as_ref()
   .map(|token_address| async move {
     info_provider.token_info(token_address).await.ok()
}
// ...
```
`info_provider` would be moved as a side effect of using an `async move` closure. This is generally ok if only done once, as mutable reference have to be unique as language restriction. If `info_provider` would be used in another closure within the same scope, the code would no longer compile. 

In this situation we worked around this by extracting the mapping function into an `async fn` to work around this moving semantics restriction. Like so:

```rust
async fn (info_provider: &mut impl InfoProvider, token_address: &String) -> Option<TokenInfo> {
// ....
}
```

Withouh the mutability requirement, we can pass InfoProvider as immutable and restore the closure usage.

jpalvarezl:
Closes #350 

- Removed mutability of `InfoProvider`
- Inlined functions that were extracted from closures previously

### DIFF
--- a/src/models/converters/safes.rs
+++ b/src/models/converters/safes.rs
@@ -24,7 +24,15 @@ impl SafeInfo {
                 })
                 .collect(),
             // "move" forces us to have a function that borrows info_provider
-            modules: map_modules_address_ex(&self.modules, info_provider).await,
+            modules: OptionFuture::from(self.modules.as_ref().map(|modules| async move {
+                let mut results = Vec::with_capacity(modules.len());
+                for module in modules {
+                    results.push(to_address_ex(module, info_provider).await)
+                }
+                results
+            }))
+            .await,
+            // map_modules_address_ex(&self.modules, info_provider).await,
             //This one can use async move as it is the last use of the info_provider and can be moved
             fallback_handler: OptionFuture::from(
                 self.fallback_handler
@@ -35,19 +43,6 @@ impl SafeInfo {
             version: self.version.to_owned(),
         }
     }
-}
-
-async fn map_modules_address_ex(
-    modules: &Option<Vec<String>>,
-    info_provider: &impl InfoProvider,
-) -> Option<Vec<AddressEx>> {
-    // early return if modules are None
-    let modules = modules.as_ref()?;
-    let mut results = Vec::with_capacity(modules.len());
-    for module in modules {
-        results.push(to_address_ex(module, info_provider).await)
-    }
-    Some(results)
 }
 
 async fn to_address_ex(address: &str, info_provider: &impl InfoProvider) -> AddressEx {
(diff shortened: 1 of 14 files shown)


src/models/converters/safes.rs
```diff
@@ -37,20 +45,7 @@ impl SafeInfo {
     }
 }
 
-async fn map_modules_address_ex(
```
rmeissner, line 40: Why are closures better than a method? Personally a method looks cleaner to me

rmeissner (approved):
In general it looks nice. Personally I prefer in most cases the method over the closure, is there any real tradeoff beside subjective code style?

src/models/converters/safes.rs
jpalvarezl: No particular reason. Just restored it to look as it used to before async/await.

jpalvarezl:
No particular reason, other than restoring the code to how it looked before the async/await refactor. I don't mind either way to be honest. Closures tend to produce duplicate code, like we have here. But in the other hand it's nice to be able to see what the closure does directly where it does it. 

I will add something like a module with methods that map `Option<String> -> Option<TokenInfo>` and `Option<Vec<String> -> ... ` etc. for `TokenInfo` and `AddresInfo`  and such, but only the ones we need for now.

src/providers/ext.rs
```diff
@@ -0,0 +1,56 @@
+use crate::models::service::safes::AddressEx;
+use crate::providers::address_info::AddressInfo;
+use crate::providers::info::{InfoProvider, TokenInfo};
+use rocket::futures::future::OptionFuture;
+
+// Using the pattern here:
+// use rocket::futures::stream::StreamExt;
```
rmeissner, line 7: Is this a todo or why is this commented?

src/utils/context.rs
```diff
@@ -17,7 +17,6 @@ impl<'r> Context<'r> {
         self.client.inner()
     }
 
-    // TODO: we would want to return the Cache trait here
```
rmeissner, line 20: Why do we remove this TODO? it is not solved, right?

src/providers/ext.rs
jpalvarezl: I left it as reference of the `TraitExt` pattern, just to show that is common. I can remove it, if you prefer.

src/utils/context.rs
jpalvarezl: I removed it because of the conversation with Nick in the second pair programming session. We can add the TODO back in if you prefer. If you want I can look into the recording and point you to the timestamp so you can have a listen. 

I remember it was something about `ServiceCache` already wrapping the implementation of cache. If you look at: https://github.com/gnosis/safe-client-gateway/blob/48a957fdcf80d9986a20dcbb1a0da9aa60008854/src/cache/redis.rs#L11

We could probably introduce that cache implementation decision logic there, more easily when necessary.

jpalvarezl: We could just have a different `State<'r, SomeOtherCacheType>`

PR #361, opened 2021-04-28
Changed response code for bad ids in tx details

### LINKED ISSUE #355: Malformed `tx_detail` id responds with a 500
https://gnosisinc.slack.com/archives/CEXL1G0MQ/p1619441097237000

We should respond with a 404 (or 422 ?) 

Investigated that it is not the GW itself generating the wrong id.

jpalvarezl:
Closes #355 

Changes:
- We return 422s now for bad ids

### DIFF
--- a/src/services/transactions_details.rs
+++ b/src/services/transactions_details.rs
@@ -142,45 +142,45 @@ pub(super) fn parse_id(details_id: &str) -> ApiResult<TransactionIdParts> {
         ID_PREFIX_MULTISIG_TX => TransactionIdParts::Multisig {
             safe_address: id_parts
                 .get(1)
-                .ok_or(api_error!("No safe address provided"))?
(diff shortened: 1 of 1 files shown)


rmeissner (approved):
Do we have tests for this?

jpalvarezl:
Yes, https://github.com/gnosis/safe-client-gateway/blob/main/src/services/tests/parse_id.rs

Not for all the branches though, be we do assert that we error on malformed ids: https://github.com/gnosis/safe-client-gateway/blob/7f3e17972ece74bc3ad11e2a4e8d0d54ef4660fe/src/services/tests/parse_id.rs#L72

rmeissner:
No was more checking, because no test had to be adjusted ;) (but I assume that is because we just check that it panics, not how exactly)

PR #363, opened 2021-04-28
Service Poll: Safe endpoint to include timestamps

### LINKED ISSUE #352: Service poll endpoint for web
**edit:** renamed query parameter from `currentTimestamp` to `timestamp`

## Feature

Web app currently polls `/transactions/queued` , `transactions/history`, `/balances/` and `/collectibles` every 15 seconds. 

We would like an endpoint for funneling the polling to these 4 endpoints into a single choke point where the web app would be able to discern what data needs refreshing. 

## Proposal

Add fields to the existing endpoint:

`GET /v1/safes/{safe_address}/`

New fields:
- `collectibles: string` : UTC timestamp (or `0`) of the last transfer (hopefully filtered by `TokenType == ERC721`.
- `tx_history: string` : UTC timestamp (or `0`) of the top most transaction in the `all-transactions` endpoint.
- `tx_queued: string`: UTC timestamp (or `0`) of the top most nonce sorted transaction in the `multisig-transactions` endpoint. 

```json
{
  "collectibles": "0", 
  "tx_history": "0",
  "tx_queued": "0",
  //...  all the fields of SafeInfoEx
}
```

The fields of `SafeInfoEx` can be found [here](https://github.com/gnosis/safe-client-gateway/blob/7f3e17972ece74bc3ad11e2a4e8d0d54ef4660fe/src/models/service/safes.rs#L5)

## Data sources

- History `/safes/{address}/all-transactions/ with limit 1`
- Queued: `​/safes​/{address}​/multisig-transactions​/` sorted by nonce (to get queued) and modified, and limit 1
- Collectibles: `/safes/{address}/transfers/ limit 1` (currently there is no way to filter for ERC721, we could check if this is possible in the core services)

## More

https://github.com/gnosis/safe-transaction-service/issues/247 there will be a filter for ERC721 that we can leverage.

jpalvarezl:
Closes #352

### DIFF
--- a/src/services/safes.rs
+++ b/src/services/safes.rs
@@ -35,14 +35,14 @@ pub fn get_safe_info_ex(context: &Context, safe_address: &String) -> ApiResult<S
 }
 
 fn get_last_collectible(context: &Context, safe_address: &String) -> ApiResult<i64> {
+    //TODO: Apply ERC721 filter when available in core services
     let url = format!(
         "{}/v1/safes/{}/transfers/?\
         &limit=1",
         base_transaction_service_url(),
         safe_address,
     );
-    //1620053387
-    //1620053387
+
     let body = RequestCached::new(url)
         .request_timeout(transaction_request_timeout())
         .execute(context.client(), context.cache())?;
--- a/src/services/safes.rs
+++ b/src/services/safes.rs
@@ -16,8 +16,6 @@ pub fn get_safe_info_ex(context: &Context, safe_address: &String) -> ApiResult<S
 
     let safe_info_ex = safe_info.to_safe_info_ex(&mut info_provider);
 
-    // let last_transfer = ;
-
     let safe_state = SafeState {
         safe_config: safe_info_ex,
         safe_state: SafeLastChanges {
--- a/src/services/safes.rs
+++ b/src/services/safes.rs
@@ -1,6 +1,7 @@
 use crate::cache::cache_operations::RequestCached;
 use crate::config::{base_transaction_service_url, transaction_request_timeout};
 use crate::models::backend::transactions::{MultisigTransaction, Transaction};
+use crate::models::backend::transfers::Transfer;
 use crate::models::commons::Page;
 use crate::models::service::safes::{SafeInfoEx, SafeLastChanges, SafeState};
 use crate::models::service::transactions::summary::TransactionSummary;
@@ -20,13 +21,41 @@ pub fn get_safe_info_ex(context: &Context, safe_address: &String) -> ApiResult<S
     let safe_state = SafeState {
         safe_config: safe_info_ex,
         safe_state: SafeLastChanges {
-            collectibles: 0,
+            collectibles: get_last_collectible(context, safe_address).unwrap_or(0),
             tx_queued: get_last_queued_tx(context, safe_address).unwrap_or(0),
             tx_history: get_last_history_tx(context, safe_address).unwrap_or(0),
         },
     };
 
     Ok(safe_state)
+}
+
+fn get_last_collectible(context: &Context, safe_address: &String) -> ApiResult<i64> {
+    let url = format!(
+        "{}/v1/safes/{}/transfers/?\
+        &ordering=executionDate\
+        &limit=1",
+        base_transaction_service_url(),
+        safe_address,
+    );
+
+    let body = RequestCached::new(url)
+        .request_timeout(transaction_request_timeout())
+        .execute(context.client(), context.cache())?;
+    let transaction: Page<Transfer> = serde_json::from_str(&body)?;
+
+    log::debug!("{:#?}", &transaction);
+    transaction
+        .results
+        .get(0)
+        .as_ref()
+        .map(|transfer| match transfer {
+            Transfer::Erc721(transfer) => transfer.execution_date.timestamp(),
+            Transfer::Erc20(transfer) => transfer.execution_date.timestamp(),
+            Transfer::Ether(transfer) => transfer.execution_date.timestamp(),
+            Transfer::Unknown => 0,
+        })
+        .ok_or(api_error!("Couldn't get tx timestamps"))
 }
 
 fn get_last_queued_tx(context: &Context, safe_address: &String) -> ApiResult<i64> {
@@ -59,6 +88,7 @@ fn get_last_history_tx(context: &Context, safe_address: &String) -> ApiResult<i6
 
     let url = format!(
         "{}/v1/safes/{}/all-transactions/?\
+        &ordering=executionDate
         &queued=false\
         &executed=true",
         base_transaction_service_url(),
--- a/src/services/safes.rs
+++ b/src/services/safes.rs
@@ -63,7 +63,7 @@ fn get_last_collectible(context: &Context, safe_address: &String) -> ApiResult<i
 fn get_last_queued_tx(context: &Context, safe_address: &String) -> ApiResult<i64> {
     let url = format!(
         "{}/v1/safes/{}/multisig-transactions/?\
-        &ordering=nonce,submissionDate\
+        &ordering=submissionDate\
         &executed=false\
         &trusted=true\
         &limit=1",
--- a/src/services/safes.rs
+++ b/src/services/safes.rs
@@ -7,7 +7,10 @@ use crate::models::service::safes::{SafeLastChanges, SafeState};
 use crate::providers::info::{DefaultInfoProvider, InfoProvider};
 use crate::utils::context::Context;
 use crate::utils::errors::ApiResult;
+use chrono::Utc;
 
+// We use Utc::now().timestamp() as the fallback value so that we don't block clients from reloading
+// as returning always 0, and the clients invalidating on value changes, would prevent reloading
 pub fn get_safe_info_ex(context: &Context, safe_address: &String) -> ApiResult<SafeState> {
     let mut info_provider = DefaultInfoProvider::new(context);
     let safe_info = info_provider.safe_info(safe_address)?;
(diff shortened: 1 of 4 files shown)


src/services/safes.rs
```diff
@@ -1,11 +1,116 @@
...
+        .execute(context.client(), context.cache())?;
+    let transaction: Page<Transfer> = serde_json::from_str(&body)?;
+
+    log::debug!("{:#?}", &transaction);
+    transaction
+        .results
+        .get(0)
+        .as_ref()
+        .map(|transfer| match transfer {
+            Transfer::Erc721(transfer) => transfer.execution_date.timestamp(),
```
rmeissner, line 51: `execution_date` is a blockchain related date ... we should check if uxio could render our a `modified_at` or similar and that we can sort by that too 🤔 

But for a first version this should do

src/services/safes.rs
```diff
@@ -1,11 +1,116 @@
...
+            Transfer::Ether(transfer) => transfer.execution_date.timestamp(),
+            Transfer::Unknown => 0,
+        })
+        .ok_or(api_error!("Couldn't get tx timestamps"))
+}
+
+fn get_last_queued_tx(context: &Context, safe_address: &String) -> ApiResult<i64> {
+    let url = format!(
+        "{}/v1/safes/{}/multisig-transactions/?\
+        &ordering=nonce,submissionDate\
```
rmeissner, line 62: we should order by `modified`

src/services/safes.rs
```diff
@@ -1,11 +1,116 @@
...
+        .request_timeout(transaction_request_timeout())
+        .execute(context.client(), context.cache())?;
+    let transaction: Page<MultisigTransaction> = serde_json::from_str(&body)?;
+
+    log::debug!("{:#?}", &transaction);
+    transaction
+        .results
+        .get(0)
+        .as_ref()
+        .map(|tx| tx.submission_date.timestamp())
```
rmeissner, line 80: we should use `modified` insread of `submission_date`

src/services/safes.rs
```diff
@@ -1,11 +1,116 @@
...
+        .execute(context.client(), context.cache())?;
+    let transaction: Page<Transaction> = serde_json::from_str(&body)?;
+
+    log::debug!("{:#?}", &transaction);
+    transaction
+        .results
+        .get(0)
+        .as_ref()
+        .and_then(|tx| {
+            tx.to_transaction_summary(&mut info_provider, safe_address)
```
rmeissner, line 107: If possible I would like to avoid this as this might pull all kind of additional information (and therefore trigger requests) that we don't need here

src/services/safes.rs
jpalvarezl: Should I remove `nonce` too?

src/services/safes.rs
jpalvarezl: Here if we use the backend tx the date selection tree changes a little. You will see in the next commit.

src/services/safes.rs
jpalvarezl: Here if we use the backend tx the date selection tree changes a little. You will see in the next commit.

jpalvarezl: We can either default to `0` or `submission_date`

src/services/safes.rs
rmeissner: Resolved by changing to the `Tag` approach

src/services/safes.rs
rmeissner: Yeah we have to do a little bit more mapping, but I think it is better. I would go with the solution from the commit you mentioned

src/services/safes.rs
rmeissner: You need to adjust https://github.com/gnosis/safe-client-gateway/pull/363/commits/dd92d6c263596928991bcd7cce6bab4d56b38c54#diff-52174f28ea57ac3eec59fb061ef68b5f1e96c0091c7d9dfb1c897bfc866b149fL62

src/services/safes.rs
rmeissner: yes I would

src/services/safes.rs
rmeissner: Nvm this is handled by the other comment :)

src/services/safes.rs
jpalvarezl: do we set `ordering=modified` ?

rmeissner: Descending, right?

jpalvarezl: yes, it would be `descending`

rmeissner: We can filter by executed, right? (same as in the queued endpoint)

src/services/safes.rs
```diff
@@ -1,11 +1,119 @@
...
+            Transaction::Multisig(tx) => tx
+                .modified
+                .as_ref()
+                .map(|it| it.timestamp())
+                .unwrap_or(tx.submission_date.timestamp()),
+            Transaction::Ethereum(tx) => tx.execution_date.timestamp(),
+            Transaction::Module(tx) => tx.execution_date.timestamp(),
+            Transaction::Unknown => 0,
+        })
+        .ok_or(api_error!("Couldn't get tx timestamps"))
```
rmeissner, line 118: Currently we would error for an empty tx list, right? That is probably not optimal :D

src/services/safes.rs
```diff
@@ -1,11 +1,119 @@
...
+    let transaction: Page<MultisigTransaction> = serde_json::from_str(&body)?;
+
+    log::debug!("{:#?}", &transaction);
+    transaction
+        .results
+        .get(0)
+        .as_ref()
+        .map(|tx| tx.modified.as_ref().map(|it| it.timestamp()))
+        .flatten()
+        .ok_or(api_error!("Couldn't get tx timestamps"))
```
rmeissner, line 85: Same as below

src/services/safes.rs
```diff
@@ -1,11 +1,119 @@
...
+            Transfer::Ether(transfer) => transfer.execution_date.timestamp(),
+            Transfer::Unknown => 0,
+        })
+        .ok_or(api_error!("Couldn't get tx timestamps"))
+}
+
+fn get_last_queued_tx(context: &Context, safe_address: &String) -> ApiResult<i64> {
+    let url = format!(
+        "{}/v1/safes/{}/multisig-transactions/?\
+        &ordering=submissionDate\
```
rmeissner, line 65: Was actually playing around with modified and it seems this is not updated when a new submission is added. I am not sure if we have a good way to detect a change in the queued without that 🤔 

How about we just set it to a random value and let it get invalidated via the cache by a hook xD ... idk ... I have no good idea right now.

src/services/safes.rs
jpalvarezl: yeah, but this is for queued, so I wouldn't

src/services/safes.rs
jpalvarezl: the caller does a  `unwrap_or(0)`

src/services/safes.rs
jpalvarezl: same here, caller does the `unwrap_or(0)`

src/services/safes.rs
jpalvarezl: that would work, but we'd need a deterministic way of generating that, right? 

What if we replicate the same url for the `transactions/queued` endpoint? I got a bit confused by the `trusted` flag. Do the clients send it? Or is it for our debugging?

src/services/safes.rs
rmeissner: yes :D

src/services/safes.rs
rmeissner: > that would work, but we'd need a deterministic way of generating that, right?
Not really .. either that random value present in cache (as part of the cached response) and we return it or not then we create a new one and return it

> What if we replicate the same url for the transactions/queued endpoint? I got a bit confused by the trusted flag.
The problem is that we need to look at the whole response from the queued endpoint (which potentially are multiple pages)

src/services/safes.rs
jpalvarezl: ok so we do `ordering=modified&executed=false`?

jpalvarezl: Just for clarity:
- Queued endpoint: `{}/v1/safes/{}/multisig-transactions/?{}&nonce__gte={}&ordering=nonce,submissionDate&trusted={}`
- Current state of the PR for last queued fetching: 
```rust
 "{}/v1/safes/{}/multisig-transactions/?\
        &ordering=submissionDate\
        &executed=false\
        &trusted=true\
        &limit=1"
```
- Proposed changes:
```rust
 "{}/v1/safes/{}/multisig-transactions/?\
        &ordering=modified\
        &executed=false\
        &trusted=true\
        &limit=1"
```

right?

src/services/safes.rs
```diff
@@ -1,11 +1,119 @@
...
+}
+
+fn get_last_collectible(context: &Context, safe_address: &String) -> ApiResult<i64> {
+    let url = format!(
+        "{}/v1/safes/{}/transfers/?\
+        &limit=1",
+        base_transaction_service_url(),
+        safe_address,
+    );
+    //1620053387
```
rmeissner, line 44: Magic numbers

src/services/safes.rs
```diff
@@ -1,11 +1,119 @@
...
+                .to_string(),
+        },
+    };
+
+    Ok(safe_state)
+}
+
+fn get_last_collectible(context: &Context, safe_address: &String) -> ApiResult<i64> {
+    let url = format!(
+        "{}/v1/safes/{}/transfers/?\
```
rmeissner, line 39: Can we add a todo, to add the filter for erc721 transfers once it is available?

rmeissner (approved):

PR #372, opened 2021-05-04
Known Addresses v3: `dataDecoded`

### LINKED ISSUE #365: Known Addresses V3 : `dataDecodedAddressInfo` map
## Feature

We need to return `AddressInfo` for any addressed contained within `TxData::dataDecoded` **but** only up to the 1st level of nesting of a multiSend.  

In order to do so, we add the `addressInfoIndex` map of the type `Map<String, Option<AddressInfo>>` , where the `String` type will represent check-summed ids.

To identify `address` fields, check for fields with string starting with `0x` and having a length of `42` (good enough for a first iteration, worst case scenario, we don't get a hit).

**Once values are added to the map, we filter any key that has `None` as a value.**

## Sample JSON

```json
{
"txData": {
    "hexData": "0x8d80ff...",
    "dataDecoded": {
        "method": "multiSend",
        "parameters": []
    },
    "addressInfoIndex" : {
        "0x8D29bE29923b68abfDD21e541b9374737B49cdAD": {
            "name": "Gnosis Safe: Multi Send 1.1.1",
            "logoUri": "https://safe-transaction-assets.staging.gnosisdev.com/contracts/logos/0x8D29bE29923b68abfDD21e541b9374737B49cdAD.png"
        }
    },
    "to": "0x8D29bE29923b68abfDD21e541b9374737B49cdAD",
    "value": "0",
    "operation": 1
},

```

jpalvarezl:
Closes #365


src/models/converters/data_decoded.rs
```diff
@@ -73,3 +75,48 @@ impl DataDecoded {
...
+        let mut index = HashMap::new();
+        if self.method == MULTI_SEND {
+        } else {
+            if let Some(parameters) = &self.parameters {
+                for parameter in parameters {
+                    match &parameter.value {
+                        ParamValue::SingleValue(value) => {
+                            if value.len() == 42 && value.starts_with("0x") {
+                                if let Some(address_info) =
+                                    info_provider.full_address_info_search(&value).ok()
```
rmeissner, line 93: If the address is already known you should not have to search again ... but probably doesn't matter as the `info_provider` handles that

src/models/converters/data_decoded.rs
```diff
@@ -73,3 +75,48 @@ impl DataDecoded {
...
+                                {
+                                    index.insert(value.to_owned(), address_info);
+                                };
+                            }
+                        }
+                        ParamValue::ArrayValue(values) => {
+                            for value in values {
+                                if let ParamValue::SingleValue(value) = value {
+                                    if value.len() == 42 && value.starts_with("0x") {
```
rmeissner, line 102: Could we extract these couple lines into a method to avoid code duplication?

src/models/converters/transactions/tests/details.rs
```diff
@@ -31,7 +31,7 @@ fn multisig_custom_transaction_to_transaction_details() {
         .returning(move |_| bail!("Token Address 0x0"));
     mock_info_provider
         .expect_full_address_info_search()
-        .times(1)
+        .times(2) //to_info and data_decoded "spender" address parameter
```
rmeissner, line 34: Could you add a space after `//`

src/models/converters/data_decoded.rs
jpalvarezl: I think it does so only for `token_info` actually. In the case of `contract_info` there isn't even an early return if we search for `0x0`, I will look into this.

src/models/converters/data_decoded.rs
jpalvarezl: Planning on doing that as soon as I've also implemented the multiSend branch and covered with unit tests :)

src/models/converters/data_decoded.rs
```diff
@@ -73,3 +75,80 @@ impl DataDecoded {
...
+        let mut index = HashMap::new();
+        if self.method == MULTI_SEND {
+            if let Some(value_decoded_type) =
+                &self.get_parameter_value_decoded(MULTI_SEND_TRANSACTIONS_PARAM)
+            {
+                match value_decoded_type {
+                    ValueDecodedType::InternalTransaction(transactions) => {
+                        transactions.iter().for_each(|transaction| {
+                            insert_value_into_index(&transaction.to, &mut index, info_provider);
+                            put_parameter_into_index(
```
rmeissner, line 93: This is theoretically already one level further down, lets see if we see an performance impact.

rmeissner (approved):
More general question: Do we want to have a "feature flag" to be able to disable this and fallback to `null`?  (Just in case this cause big performance issues)

rmeissner (approved):

PR #375, opened 2021-05-06
Added support to spectrum in the cache warmer script

jpalvarezl:
Made it easier to run the cache warmer script in `spectrum` env


jpalvarezl:
I just find it handy, but I wouldn't merge it as it's a bit nasty :) will convert to draft

PR #388, opened 2021-05-17
Partial response on `unwrap_or_default`

### LINKED ISSUE #368: `unwrap_or(vec![])` results in partial responses cached for too long
https://github.com/gnosis/safe-client-gateway/blob/main/src/services/transactions_history.rs#L144

Failure of a single mapping shouldn't invalidate the entire collection mapping. We should `try` for each individual transaction. Using something like `filter_map` if available.

jpalvarezl:
Closes #368 

Changes:
- We filter out errors in mapping but we return all the correct values in one page


jpalvarezl:
Timebox exceeded

PR #391, opened 2021-05-19
Domain separator for new safe version

### LINKED ISSUE #348: Adjust `domainSeparator` for calculating the `safeTxHash` for cancellation TXs
Version 1.3.0 of the safe contracts introduce a new way of calculating the `domainSeparator`.

Currently, this is simply hardcoded in the GW which will no longer be valid for transactions issued from +1.3.0

## Proposal

1. Fetch `SafeInfo` and verify `version`
2. Fallback to old `domainSeparator` if version < 1.3.0
3. Calculate the `domainSeparator` according to: https://github.com/gnosis/safe-contracts/blob/53122d14af0a3f6aa30c5fcd3861573d44c9d7b9/contracts/GnosisSafe.sol#L360

## Dependencies

How do we determine the `chainId` ?
Env variable.

jpalvarezl:
Closes #348 

Changes:
- Added checking of safe version for calculating the `domain_separator`
- I still need to find real world data for `domain_hash_for_safe_address` test

### DIFF
--- a/src/config/mod.rs
+++ b/src/config/mod.rs
@@ -169,5 +169,5 @@ pub fn version() -> String {
 }
 
 pub fn chain_id() -> u64 {
-    u64_with_default("CHAIN_ID", 4) // Is there a more sensible default than Rinkeby?
+    u64_with_default("CHAIN_ID", 1)
 }
--- a/src/config/mod.rs
+++ b/src/config/mod.rs
@@ -167,3 +167,7 @@ pub fn version() -> String {
         .unwrap_or(env!("CARGO_PKG_VERSION"))
         .to_string()
 }
+
+pub fn chain_id() -> u64 {
+    u64_with_default("CHAIN_ID", 4) // Is there a more sensible default than Rinkeby?
+}
--- a/src/utils/transactions.rs
+++ b/src/utils/transactions.rs
@@ -30,8 +30,7 @@ pub async fn fetch_rejections(
     let safe_address: Address =
         serde_json::from_value(serde_json::value::Value::String(safe_address.to_string())).unwrap();
 
-    let safe_tx_hash =
-        to_hex_string!(hash(safe_address, nonce, domain_hash(&safe_address, is_legacy)).to_vec());
+    let safe_tx_hash = to_hex_string!(hash(safe_address, nonce, is_legacy).to_vec());
 
     let multisig_tx = fetch_cancellation_tx(context, safe_tx_hash).await;
     multisig_tx
@@ -47,9 +46,10 @@ pub async fn fetch_rejections(
         .flatten()
 }
 
-pub(super) fn hash(safe_address: Address, nonce: u64, domain_hash: [u8; 32]) -> [u8; 32] {
+pub(super) fn hash(safe_address: Address, nonce: u64, is_legacy: bool) -> [u8; 32] {
     let erc_191_byte = u8::from_str_radix(ERC191_BYTE, 16).unwrap();
     let erc_191_version = u8::from_str_radix(ERC191_VERSION, 16).unwrap();
+    let domain_hash = domain_hash(&safe_address, is_legacy);
 
     let mut encoded = ethabi::encode(&[
         ethabi::Token::Uint(Uint::from(domain_hash)),
--- a/src/utils/transactions.rs
+++ b/src/utils/transactions.rs
@@ -1,12 +1,12 @@
 use crate::cache::cache_operations::RequestCached;
 use crate::config::{base_transaction_service_url, chain_id, transaction_request_timeout};
 use crate::models::backend::transactions::MultisigTransaction;
-use crate::providers::info::{DefaultInfoProvider, InfoProvider};
+use crate::providers::info::{DefaultInfoProvider, InfoProvider, SAFE_V_1_3_0};
 use crate::utils::context::Context;
 use ethabi::Uint;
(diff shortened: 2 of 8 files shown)


src/utils/transactions.rs
```diff
@@ -52,15 +61,29 @@ pub(super) fn hash(safe_address: Address, nonce: u64) -> [u8; 32] {
     keccak256(encoded)
 }
 
-pub(super) fn domain_hash(safe_address: &Address) -> [u8; 32] {
+pub(super) fn domain_hash(safe_address: &Address, is_legacy: bool) -> [u8; 32] {
+    let domain_separator_typehash = if is_legacy {
+        DOMAIN_SEPARATOR_TYPEHASH_LEGACY
+    } else {
+        DOMAIN_SEPARATOR_TYPEHASH
+    };
```
fmrsabino, line 69: Wondering if the type hash would change in the future we would end up with more legacy references. What if we match the version here (not sure if it's possible). Was thinking something along the lines:

```rust
let safe_version = Version::parse(safe_info.version.unwrap().as_str());
let domain_separator_typehash = match safe_version {
  Ok(Version::parse(SAFE_V_1_3_0)) => { DOMAIN_SEPARATOR_TYPEHASH }
  Ok(_) => { DOMAIN_SEPARATOR_TYPEHASH_LEGACY }
  Err(_) => {} // panic/error result in case we cannot retrieve the version?
};
```

src/config/mod.rs
```diff
@@ -167,3 +167,7 @@ pub fn version() -> String {
         .unwrap_or(env!("CARGO_PKG_VERSION"))
         .to_string()
 }
+
+pub fn chain_id() -> u64 {
+    u64_with_default("CHAIN_ID", 4) // Is there a more sensible default than Rinkeby?
```
rmeissner, line 172: We should use mainnet (1)

src/utils/transactions.rs
```diff
@@ -19,10 +23,14 @@ pub async fn fetch_rejections(
     safe_address: &str,
     nonce: u64,
 ) -> Option<Vec<String>> {
+    let info_provider = DefaultInfoProvider::new(&context);
+    let safe_info = info_provider.safe_info(safe_address).await.ok();
+    let is_legacy = is_legacy_domain_separator(safe_info);
```
rmeissner, line 28: I would call this `use_legacy_domain_separator` as we do not test the donain separator and therefore `is_...` is misleading

src/utils/transactions.rs
```diff
@@ -86,6 +109,19 @@ pub(super) fn cancellation_parts_hash(safe_address: &Address, nonce: u64) -> [u8
...
 }
 
+pub(super) fn is_legacy_domain_separator(safe_info: Option<SafeInfo>) -> bool {
+    let version = safe_info
+        .as_ref()
+        .and_then(|safe_info| safe_info.version.as_ref().map(|it| Version::parse(it).ok()))
+        .flatten();
+
+    if let Some(version) = version {
+        version < Version::parse(SAFE_V_1_3_0).unwrap_or(Version::new(1, 3, 0))
```
rmeissner, line 119: Can we store the parsed versuon in a global constant?

src/utils/transactions.rs
rmeissner: I would not do this as in your code any newer version would also use the legacy typehash. But for version 1.4.0 we most likely will have the same typehash as for 1.3.0

rmeissner (approved):
Functional wise it looks good, but I would structure the code differently.

E.g. have a method `async domainHash(safeAddress)` in this method we fetch the version and then either calls `domainHashV130(safeAddress)` or `domainHashV100(safeAddress)`

This way you don't have to pass along a boolean flag and have an if in multiple places

src/utils/transactions.rs
fmrsabino: I'd actually remove the default that I mentioned and be more explicit on the versions in the matching or throw an error otherwise

fmrsabino: Also doesn't need to be for each specific version but if the branches can be interval of versions I think that would also work :)

src/utils/transactions.rs
jpalvarezl: will rename, but we do test it!

src/utils/transactions.rs
jpalvarezl: I will have to introduce `lazy_static` for this, just FYI

jpalvarezl:
sounds good, also, I need to somehow pass down the `chain_id` since I don't have direct control of the env var for the test. Would be nice to inject the value easily.

src/utils/transactions.rs
jpalvarezl: I think that makes sense, but I would match the interval and not the `Result` , I think that if we get an `Err` we would fallback to the legacy hash, right?

src/utils/transactions.rs
jpalvarezl: Lol it was already there and being used xD

src/utils/transactions.rs
fmrsabino: Not sure what the best behaviour here would be for the fallback. I'm more leaning toward an error an being more conservative regarding the version matching. But this is mostly a suggestion! It looks good :)

fmrsabino (approved):

PR #393, opened 2021-05-19
Decode Module Transactions

### LINKED ISSUE #387: Decode module transactions
Currently module transactions are always decoded as `Custom` (see https://github.com/gnosis/safe-client-gateway/blob/main/src/models/converters/transactions/mod.rs#L253)

We should apply similar decoding as for a multisig transaction (see https://github.com/gnosis/safe-client-gateway/blob/main/src/models/converters/transactions/mod.rs#L68)

#### Notes
- [ ] Check with the clients that this is no issue (especially check how this affects web)
  - Maybe add a feature flag for it or a new endpoint
- [x] ~Check if we should make [`ExecutionInfo`](https://github.com/gnosis/safe-client-gateway/blob/main/src/models/service/transactions/summary.rs#L13) polymorphic to also return module information (would be a new endpoint if this breaks the clients) -> might be a separate issue.~ (Not Now)

fmrsabino:
Closes #387

- Extract common transaction components from `MultisigTransaction` and `ModuleTransaction` to `SafeTransaction`
- JSON serialisation does not change as `SafeTransaction` is flattened in `MultisigTransaction` and `ModuleTransaction`
- Add common transaction decoding functionality to `SafeTransaction`
  * The difference between the decoding of `MultisigTransaction` and `ModuleTransaction` is on the custom transaction creation, more specifically the `is_cancellation` check
  * In `MultisigTransaction` – `is_cancellation` is computed using data from the `MultisigTransaction` which is not available in `SafeTransaction` (`base_gas`, `gas_price`, `gas_token`, `refund_receiver` and `safe_tx_gas`)
  * In `ModuleTransaction` – `is_cancellation` is `false`

### Open Questions

- One important change in terms of the model representation is that `operation` was optional in `MultisigTransaction` but not in `ModuleTransaction`. It is now optional in both. This needs to be double checked with the core services regarding the nullability of these fields.


src/models/converters/transactions/details.rs
```diff
@@ -18,20 +18,23 @@ impl MultisigTransaction {
         rejections: Option<Vec<String>>,
         info_provider: &(impl InfoProvider + Sync),
     ) -> ApiResult<TransactionDetails> {
-        let safe_info = info_provider.safe_info(&self.safe.to_string()).await?;
+        let safe_transaction = &self.safe_transaction;
```
jpalvarezl, line 21: Why not inline this? Not that I would be against it if it's just for readability, just curious :)

src/models/converters/transactions/details.rs
```diff
@@ -105,17 +108,19 @@ impl ModuleTransaction {
...
-                hex_data: self.data.to_owned(),
-                data_decoded: self.data_decoded.clone(),
-                operation: self.operation,
-                address_info_index: OptionFuture::from(self.data_decoded.as_ref().map(
+                to: safe_transaction.to.to_owned(),
+                value: safe_transaction.value.to_owned(),
+                hex_data: safe_transaction.data.to_owned(),
+                data_decoded: safe_transaction.data_decoded.clone(),
+                // TODO: check default value and if it is optional
+                operation: safe_transaction.operation.unwrap_or(Operation::CALL),
```
jpalvarezl, line 122: I think we should resolve this `TODO` before merging. We could just talk about it in the internal services channel.

src/models/converters/transactions/mod.rs
```diff
@@ -99,12 +60,35 @@ impl MultisigTransaction {
...
+                    self.custom_decode(info_provider, is_cancellation).await,
+                ),
             }
         } else {
-            TransactionInfo::Custom(self.to_custom(info_provider).await)
+            TransactionInfo::Custom(self.custom_decode(info_provider, is_cancellation).await)
+        }
+    }
+
+    async fn custom_decode(
```
jpalvarezl, line 76: Just being a bit nit picky, but could we rename this to `to_custom` to maintain the style of the other methods?

src/models/converters/transactions/mod.rs
```diff
@@ -19,58 +21,17 @@ use crate::providers::info::{InfoProvider, SafeInfo, TokenInfo, TokenType};
...
-        }
-    }
-
-    async fn transaction_info(&self, info_provider: &impl InfoProvider) -> TransactionInfo {
+impl SafeTransaction {
+    async fn transaction_info(
+        &self,
+        info_provider: &impl InfoProvider,
+        is_cancellation: bool,
+    ) -> TransactionInfo {
```
jpalvarezl, line 29: I realise that this is not part of your changes; but I am a it surprise we don't wrap the result value of this function in `ApiResult` 😁

jpalvarezl (commented):
It looks really good! Thanks for the effort of gathering so much JSON for testing this thoroughly!
I only see the `TODO` blocking the merging of this PR, the rest is totally optional :)

src/models/converters/transactions/details.rs
fmrsabino: Completely agree. Let's sync on this before merging!

jpalvarezl (approved):

src/models/converters/transactions/details.rs
fmrsabino: Fixed and rebased. Since optional was changed the object hash also changed. Fixed the test here: 02daef0

jpalvarezl (approved):

src/models/converters/transactions/details.rs
jpalvarezl: Makes sense! We definitely need to do something about what we hash for this!

PR #400, opened 2021-05-21
Closes #395: Update to latest rocket changes

### LINKED ISSUE #395: Update to latest dev state of Rocket

rmeissner:
Closes #395 
- Update to https://github.com/SergioBenitez/Rocket/commit/dd9629697a3cbe45f50645579f41f1a957a92968
- Update Rust nightly to version from `2021-05-21`


src/utils/tests/errors.rs
```diff
@@ -21,7 +21,7 @@ async fn api_error_responder_json() {
     let mut response = api_error.respond_to(&request).unwrap();
 
     let status_code: u16 = response.status().code;
-    let body_json = &response.body_string().await.unwrap();
+    let body_json = &response.body_mut().to_string().await.unwrap();
```
jpalvarezl, line 24: oh it kind of sucks that they got rid of that method I suppose.

src/services/transactions_list.rs
```diff
@@ -50,20 +50,20 @@ pub async fn get_all_transactions(
             .as_ref()
             .and_then(|link| extract_query_string(link))
             .map(|link| {
-                context.build_absolute_url(uri!(
-                    crate::routes::transactions::all: safe_address,
+                context.build_absolute_url(uri!(crate::routes::transactions::all(
```
jpalvarezl, line 53: this now looks more like the method actually being called than the lambda. There is no change in behaviour between this and the previous code?

src/utils/tests/errors.rs
rmeissner: I actually like the new one more, as it is more explicit what is happening. Before it was really clear that you are consuming the response.

src/services/transactions_list.rs
rmeissner: For us it only changed the syntax and I like the new syntax more :D ... but there are also more changes, would have to search for the commit where they changed it to find the details :/

Cargo.toml
```diff
@@ -13,8 +13,12 @@ dotenv_codegen = "0.15.0"
...
 lazy_static = "1.4.0"
 
-rocket = { git = "https://github.com/SergioBenitez/Rocket", rev = "4f3511786cb8b6540124b867591c7c5a6933409e" }
-rocket_codegen = { git = "https://github.com/SergioBenitez/Rocket", rev = "4f3511786cb8b6540124b867591c7c5a6933409e" }
+rocket = { git = "https://github.com/SergioBenitez/Rocket", rev = "dd9629697a3cbe45f50645579f41f1a957a92968", features = ["tls", "json"] }
+rocket_codegen = { git = "https://github.com/SergioBenitez/Rocket", rev = "dd9629697a3cbe45f50645579f41f1a957a92968" }
+
+# Force rocket peer-dependencies versions
+proc-macro2 = "1.0.27"
+tokio = "1.6.1"
```
jpalvarezl, line 21: Is this actually okay? I think `rocket` exposes the version of async that it uses, if we force a different version onto it we could have issues.

src/models/converters/tests/safes.rs
```diff
@@ -1,6 +1,7 @@
 use crate::models::service::safes::{AddressEx, SafeInfoEx};
 use crate::providers::address_info::AddressInfo;
 use crate::providers::info::*;
+use rocket::serde::json::json;
```
jpalvarezl, line 4: This import seems unused

src/utils/context.rs
```diff
@@ -9,12 +8,12 @@ pub struct Context<'r> {
     uri: String,
     host: Option<String>,
     cache: ServiceCache<'r>,
-    client: State<'r, reqwest::Client>,
+    client: &'r reqwest::Client,
```
jpalvarezl, line 11: Why don't we wrap this in `State` anymore?

src/utils/errors.rs
```diff
@@ -132,11 +129,11 @@ impl From<serde_json::error::Error> for ApiError {
     }
 }
 
-impl From<rocket_contrib::json::JsonError<'_>> for ApiError {
-    fn from(err: JsonError<'_>) -> Self {
+impl From<rocket::serde::json::Error<'_>> for ApiError {
```
jpalvarezl, line 132: We can probably get rid of the full module path import here, as you have it in a `use` statement at the top

jpalvarezl (approved):
Remember to add "Closes #395" to the description

Cargo.toml
jpalvarezl: Also curious about why we need `proc-macro2` as well. Same issue?

src/utils/context.rs
rmeissner: State doesn'T exist anymore

src/utils/errors.rs
rmeissner: Yes, but I only did a find replace refactor, and no additional changes

src/models/converters/tests/safes.rs
rmeissner: https://github.com/gnosis/safe-client-gateway/blob/2b4173766df67da8304bc18cef4d129af7bb279a/src/models/converters/tests/safes.rs#L170

Cargo.toml
rmeissner: This corresponds to the version that `rocket` requires. If the versions are not compatible the compiler will complain

PR #404, opened 2021-05-26
Pinned `grcov` to v0.8.0

### LINKED ISSUE #369: GitHub actions implementation improvements
The current implementation of our buildscripts, do cache builds across commits in the same PR but the deployment step does not re-use the image generated in previous steps (verify that this is generated), making deployments very slow.

Figure out how to reduce the times we build the project.

- [x] improve safety of how coverge report is uploaded
- [x] improve the deployment time by reusing builds from previous steps

jpalvarezl:
Closes #369 

Changes:
- Pinned version of grcov to `v0.8.0`
- Removed any `spectrum` associated branch in the deployment script

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -68,11 +68,11 @@ jobs:
         uses: actions/cache@v2
         with:
           path: /home/runner/.cargo/bin/
-          key: ${{ runner.os }}-grcov-latest
+          key: ${{ runner.os }}-grcov-v080
 
       - name: Fetch grcov
         if: steps.grcov-cache.outputs.cache-hit != 'true'
-        run: curl --location https://github.com/mozilla/grcov/releases/latest/download/grcov-linux-x86_64.tar.bz2 | tar jxf -
+        run: curl --location https://github.com/mozilla/grcov/releases/download/v0.8.0/grcov-tcmalloc-linux-x86_64.tar.bz2 | tar jxf -
 
       - name: Run grcov
         id: coverage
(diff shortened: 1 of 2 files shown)


.github/workflows/rust.yml
```diff
@@ -116,7 +116,3 @@ jobs:
       - name: Deploy develop # unused
         if: github.ref == 'refs/heads/develop'
         run: bash scripts/deploy_docker.sh develop
```
fmrsabino, line 118: This can also be removed right?

fmrsabino (approved):

.github/workflows/rust.yml
```diff
@@ -68,11 +68,11 @@ jobs:
         uses: actions/cache@v2
```
fmrsabino, line 68: Not sure if the scope of this PR is just for pinning `grcov` but we can also update the cache action (can be done in a different PR)

```suggestion
        uses: actions/cache@v2.1.5
```

jpalvarezl: Thanks for that! Yeah, I don't see why we couldn't include those changes :)

fmrsabino (approved):

.github/workflows/rust.yml
jpalvarezl: just for reference: https://github.com/actions/cache/releases if we want to check release notes.

.github/workflows/rust.yml
jpalvarezl: I think I would keep it, as it is not entirely unlikely we introduce a dev branch at some point

PR #406, opened 2021-05-27
Refactor `safeTxHash` calculation

### LINKED ISSUE #399: Refactor `safeTxHash` calculation
- [x] use https://doc.rust-lang.org/std/env/fn.set_var.html to set the chain id and test it. Maybe move `chain_id()` call further down
- [x] Pass only the version to `use_legacy_domain_separator` instead of the full `safe_info` object

jpalvarezl:
Closes #399 

Changes:
- Moved down `get_chain_id` call, setting the `env` var in the test, thus validating the name
- Passed only the version for `use_legacy_domain_separator` method

### DIFF
--- a/src/utils/transactions.rs
+++ b/src/utils/transactions.rs
@@ -25,8 +25,7 @@ pub async fn fetch_rejections(
     nonce: u64,
 ) -> Option<Vec<String>> {
     let info_provider = DefaultInfoProvider::new(&context);
-    let safe_info = info_provider.safe_info(safe_address).await.ok();
-    let version = safe_info
+    let version = info_provider.safe_info(safe_address).await.ok()
         .as_ref()
         .and_then(|safe_info| safe_info.version.as_ref().map(|it| Version::parse(it).ok()))
         .flatten();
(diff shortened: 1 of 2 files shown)


src/utils/transactions.rs
```diff
@@ -112,12 +118,7 @@ pub(super) fn cancellation_parts_hash(safe_address: &Address, nonce: u64) -> [u8
     keccak256(encoded_parts)
 }
 
-pub(super) fn use_legacy_domain_separator(safe_info: Option<SafeInfo>) -> bool {
-    let version = safe_info
-        .as_ref()
-        .and_then(|safe_info| safe_info.version.as_ref().map(|it| Version::parse(it).ok()))
-        .flatten();
-
+pub(super) fn use_legacy_domain_separator(version: Option<Version>) -> bool {
```
fmrsabino, line 121: nitpick: shall we name it `uses_legacy_domain_separator` instead? `use_legacy_domain_separator` seems like it's a "decorator" pattern. Wdyt?

rmeissner (approved):

src/utils/transactions.rs
rmeissner: Hmm I always read it like asking the method "should I use the legacy domain separator". And your would be more "does this version use the legacy domain separator". No preference from my side.

fmrsabino: Ah I see your point. No preference either 👍

fmrsabino (approved):

PR #414, opened 2021-06-01
Move documentation to `cargo doc`

jpalvarezl:


src/routes/about.rs
```diff
@@ -7,6 +7,22 @@ use crate::utils::context::Context;
 use crate::utils::errors::ApiResult;
 use rocket::response::content;
 
+/**
+ * `/About` <br />
```
rmeissner, line 11: wrong casing

PR #419, opened 2021-06-03
Invalidation handles

### LINKED ISSUE #408: Endpoints for staged cache invalidation
Currently we have very coarse control over what to invalidate from the cache using our endpoints.

We should introduce endpoints that enable the invalidation of:
- specific group of endpoints (transaction, balances, collectibles, tokens, safes)
- invalidate only responses or internal requests
- allow to provide partial address or fields with regexes to have staged flushing (0xa*, 0xb*, 0xc* ...)
- remove `flush_all` endpoint

jpalvarezl:
Closes #408 

Changes:
- Introduced more variants for `InvalitationPattern` to accomodate for finer control
- Introduced `InvalidationScope` to control whether to invalidate `Responses`, `Requests` or `Both`
- Added `POST /v1/flush/<token>` endpoint
- Removed `GET /v1/flush_all/<token>` endpoint

Example for invalidating transactions for safe `0xd6f5Bef6bb4acD235CF85c0ce196316d10785d67`

```json
{
    "invalidate" :"Transactions",
    "pattern_details":  ["Responses", "0xd6f5Bef6bb4acD235CF85c0ce196316d10785d67"]
}
```

This will result in the invalidation of any cached responses for the following endpoints:
- `/v1/safes/<address>/transactions/history`
- `/v1/safes/<address>/transactions/queued`

`TransactionDetails` don't get invalidated with this pattern, as those may only contain the `safe_tx_hash` for that, we continue to use the same pattern as before, which would also continue to invalidate the transaction list endpoints (as the id contains the `safe_tx_hash`)

Webhook triggered invalidation continues to:
- 1st: invalidate requests and responses containing the `safe_address` from the payload: string pattern `c_re*<safe_address>*`
- 2nd: invalidate requests and responses containing the `safe_tx_hash` from the payload: string pattern `c_re*<safe_tx_hash>*`


src/cache/cache_operations.rs
```diff
@@ -20,12 +25,68 @@ pub struct Invalidate {
     database: Database,
 }
 
+#[derive(Deserialize, Debug)]
+pub enum InvalidationScope {
+    Requests,
+    Responses,
+    Both,
```
fmrsabino, line 32: Wondering if we can get away without `Both` but have something like:

```javascript
{
    "invalidate" :"Transactions",
    "pattern_details":  ["0xd6f5Bef6bb4acD235CF85c0ce196316d10785d67","Responses,Requests"]
}
```

to invalidate `Both`. Maybe this increases the complexity of the implementation that we have too 🤔

jpalvarezl: Right, in that case we would have to somehow tokenise the string and check every entry in the array against the enum, which as it is, we get for free if we use individual values.

src/cache/cache_op_executors.rs
```diff
@@ -1,26 +1,17 @@
...
 use rocket::response::content;
 use serde::Serialize;
 use std::time::Duration;
 
-const CACHE_REQS_PREFIX: &'static str = "c_reqs";
-const CACHE_RESP_PREFIX: &'static str = "c_resp";
-const CACHE_REQS_RESP_PREFIX: &'static str = "c_re";
+pub(super) const CACHE_REQS_PREFIX: &'static str = "c_reqs";
+pub(super) const CACHE_RESP_PREFIX: &'static str = "c_resp";
+pub(super) const CACHE_REQS_RESP_PREFIX: &'static str = "c_re";
```
rmeissner, line 11: Should we move them to the other file instead of making them `pub(super)`?

src/cache/cache_operations.rs
```diff
@@ -20,12 +25,64 @@ pub struct Invalidate {
...
+    Any(String, InvalidationScope),
+    Transactions(String, InvalidationScope),
+    Balances(String, InvalidationScope),
+    Collectibles(String, InvalidationScope),
+    Contracts,
     Tokens,
 }
 
+impl InvalidationPattern {
+    pub(super) fn to_pattern_string(&self) -> String {
```
rmeissner, line 47: Why not return a `&str`?

src/cache/cache_operations.rs
```diff
@@ -20,12 +25,64 @@ pub struct Invalidate {
...
+    Responses,
+    Both,
+}
+
+#[derive(Deserialize, Debug)]
+#[serde(tag = "invalidate", content = "pattern_details")]
 pub enum InvalidationPattern {
-    FlushAll,
-    RequestsResponses(String),
+    Any(String, InvalidationScope),
```
rmeissner, line 38: Minor: I find the order of pattern, scope confusing. I would have the scope first and then the pattern

src/cache/cache_operations.rs
```diff
@@ -20,12 +25,64 @@ pub struct Invalidate {
...
+    Transactions(String, InvalidationScope),
+    Balances(String, InvalidationScope),
```
rmeissner, line 40: Theoretically we also have transfers

rmeissner (approved):
Some minor stuff, but looks good to merge

src/cache/cache_operations.rs
jpalvarezl: Yeah, I tried to find a regex that hits both the core and gw urls keys. That's the only thing I see as source of issues.

src/cache/cache_operations.rs
jpalvarezl: That would change the JSON I placed in the description of this issue. I don't mind either though :) just FYI. I will update accordingly

PR #430, opened 2021-06-09
Core request builder

### LINKED ISSUE #413: Build request backend request URL using `chain_id`
## Context

Currently we use the `format!` macro to build requests to the core services. 

## Feature

We would want to build this URL having a macros of our own, which takes in the `chain_id` along with a string with the placeholder tokens for interpolation and the values to be placed in them. 

example:

```rust
let url = core_service_url!(
    "{}/v1/safes/{}/collectibles/?trusted={}&exclude_spam={}", 
    chain_id, 
    safe_address, 
    trusted, 
    exclude_spam);
```

## Open questions

We would need to pass the `info_provider` in this macros as there would have to be network request to fetch the network details 

## Dependencies

We need the Config Service endpoint that returns the list of supported network with their respective devOps urls

jpalvarezl:
Closes #413


src/providers/info.rs
```diff
@@ -211,6 +221,17 @@ impl<C: Cache> DefaultInfoProvider<'_, C> {
         }
     }
 
+    async fn load_chain_info(&self, chain_id: &str) -> ApiResult<Option<ChainInfo>> {
+        let url = format!("{}/v1/chains/{}/", base_config_service_url(), chain_id);
+        let data = RequestCached::new(url)
+            // .cache_duration(safe_info_cache_duration())
+            // .error_cache_duration(short_error_duration())
+            // .request_timeout(safe_info_request_timeout())
+            .execute(self.client, self.cache)
```
jpalvarezl, line 230: Do we want custom cache duration (and error duration) and request timeouts for this? I would also add a new variant to the `InvalidationPattern::Chains` so we can more easily debug this.

rmeissner: Yes I would say so

src/services/mod.rs
```diff
@@ -20,3 +22,12 @@ pub fn offset_page_meta(meta: &PageMetadata, offset: i64) -> String {
     }
     .to_url_string()
 }
+
+pub async fn backend_url(
```
rmeissner, line 26: I would maybe name this something like `core_sevices_url`

src/services/mod.rs
```diff
@@ -20,3 +22,12 @@ pub fn offset_page_meta(meta: &PageMetadata, offset: i64) -> String {
...
+    chain_id: &str,
+    info_provider: impl InfoProvider,
+    path_query: impl Fn() -> String,
```
rmeissner, line 29: For me it feels weird to pass a function for this. This would be a very nice use case to put this into a macro

jpalvarezl: I agree with you but after reading a bit about macros, there are certain things that make me think that this might not be a good use case for them. As I understand it, the nice thing to have here would be `core_backend_url!(chain_id, infor_provider, varargs)` , right? Basically having a variadic function.

`macros` don't have type checks and in general wouldn't prevent us from passing the arguments in whichever order, unlike regular functions do. Which makes me think, how do I even know which variable in the macro is the `info_provider` ? I found this example: https://www.reddit.com/r/learnrust/comments/ipbtge/how_to_generate_async_methods_with_macro/

Where I could technically try passing the `InfoProvider` trait as a first argument and even have an async context. I am not sure how this would play out, but yeah, I've read a bit as well that macros would slow down compilation time, even though right now it is not an issue.

Maybe the callback isn't the nices solution. Maybe I could pass on the path parameters as a `&[&str]` , if you prefer that in terms of style. I personally prefer the callback.

src/services/tests/backend_url.rs
```diff
@@ -0,0 +1,80 @@
...
+        mock_info_provider,
+        chain_id,
+        "/v1/safes/{}/balances/usd/?trusted={}&exclude_spam={}",
+        safe_address,
+        trusted,
+        exclude_spam
+    );
+
+    assert_eq!(url.unwrap(),"https://safe-transaction.mainnet.gnosis.io/v1/safes/0x1230B3d59858296A31053C1b8562Ecf89A2f888b/balances/usd/?trusted=false&exclude_spam=true".to_string());
+    Ok(())
```
jpalvarezl, line 38: Because the `macros` contains `?` we are forced to return `Result` in the enclosing context`

src/services/tests/backend_url.rs
```diff
@@ -0,0 +1,80 @@
...
+#[rocket::async_test]
+async fn core_uri_error() -> ApiError {
+    let mut mock_info_provider = MockInfoProvider::new();
+    mock_info_provider
+        .expect_chain_info()
+        .times(1)
+        .returning(move |_| bail!("Unsupported net"));
+
+    let url = core_uri!(mock_info_provider, "1", "/nice/path");
+    url.unwrap_err()
```
jpalvarezl, line 79: I have tried a combination of methods here. I can't correctly assert this error for some reason.

src/services/mod.rs
jpalvarezl: How about `core_uri` as you did in the example you sent me? It's a bit shorter which is nice.

src/macros.rs
```diff
@@ -65,3 +65,25 @@ macro_rules! to_hex_string {
...
+        let result: ApiResult<String> =
+        match $info_provider.chain_info($chain_id).await {
+            Ok(chain_info) => Ok(format!("{}{}",chain_info.tx_service_url, $path)),
+            Err(error) => Err(error,)
+        };
+        result
+    }};
+    ($info_provider:tt, $chain_id:expr, $path:literal, $($arg:tt)*) => {{
+        let full_path = format!($path, $($arg)*);
+        let result: ApiResult<String> =
```
rmeissner, line 82: You can solve this by calling the other macro section

src/macros.rs
```diff
@@ -65,3 +65,25 @@ macro_rules! to_hex_string {
...
+        match $info_provider.chain_info($chain_id).await {
```
rmeissner, line 83: Why can we not use `?`?

rmeissner (approved):

src/macros.rs
jpalvarezl: Becauses `?` can only be used in methods that return `Result<V, E>` , which in our case it is most of the time, except for unit tests. By `match`ing the `result` we are writing the macro in a way that makes no assumption on the calling context.

PR #434, opened 2021-06-11
Safe `guard` field

### LINKED ISSUE #428: Return guard info in safe info endpoint
Guards info is not returned as part of safe info endpoint, it's already implemented in transaction service. 

We need to get the guard info as ExtendedAddressInfo structure, more info in the iOS ticket below

**Additional context**
[Transaction Service implementation](https://github.com/gnosis/safe-transaction-service/issues/268)
[iOS ticket](https://app.zenhub.com/workspaces/safe-multisig---mobile-5c1b98250e13551e8aae81eb/issues/gnosis/safe-ios/1222)

jpalvarezl:
Closes #428 

Changes:
- `guard` and `fallback_handler` are always expected from core services (they are not nullable, even for older safe versions)
- When `0x0` is return for either address, the gw serves `None` 
- `modules` are no longer served as an empty list, if not available. (not a breaking change as they were already nullable)


src/providers/ext.rs
```diff
@@ -19,6 +19,9 @@ pub trait InfoProviderExt: InfoProvider {
         addresses: &Option<Vec<String>>,
     ) -> Option<Vec<AddressEx>> {
         let addresses = addresses.as_ref()?;
+        if addresses.is_empty() {
+            return None;
```
jpalvarezl, line 23: This is safe because `modules` was exposed as nullable already by the gw. Instead of `Some([])`, `None` seems more correct.

rmeissner: Hmm do you really think it makes more sense? ... I mean I have no strong opinion, but I would like that the client teams agree to this.

rmeissner (approved):

src/providers/ext.rs
jpalvarezl: it's by the same token as the reasoning behind `0x0...0` is mapped into `null` I would say. There is no meaningful information or alternative meaning to having an empty list instead of `null`, AFAIK. I will double check this just to be totally sure.

fmrsabino (approved):

Uxio0 (approved):

src/providers/ext.rs
jpalvarezl: To give an update: the teams were notified and the behaviour was agreed on.

PR #435, opened 2021-06-14
Use `core_uri!` macro

### LINKED ISSUE #433: Wire `core_service_url` with the `chain_id` passed in the endpoints
generate the url using the correct `ChainInfo`

- [x] `GET /v1/safe/<address>` is used for polling and safe state 
- [x] `GET /v1/transactions/<tx_id>` is used to fetch expanded data on a single transaction
- [x] `GET /v1/safes/<address>/transactions/queued` is used to fetch “pending” and “next” transactions (depending on polling state returned by `/v1/safe/<address>`)
- [x] `GET /v1/safes/<address>/transactions/history` is used to fetch the history of executed transactions (depending on polling state returned by `/v1/safe/<address>`)
- [x] `POST /v1/transactions/<safe_address>/propose` proposes a multisig transaction
- [x] `POST /v1/transactions/<safe_address>/confirmation` submits a confirmation from an owner
- [x] `GET /v1/safes/<address>/balances/<fiat_code>` list of ERC20 balances (depending on polling state returned by `/v1/safe/<address>`)
- [x] `GET /v1/safes/<address>/collectibles` list of NFTs (no pagination)
- [x] usages of any methods from `InfoProvider` and its extension functions 

Forked further cleanup to #436

jpalvarezl:
Closes #433 

- Using new macros
- Use the safe config url as provided in `.env.example`
- The only network available for now is `"4"`

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -72,7 +72,7 @@ macro_rules! core_uri {
     ($info_provider:tt, $chain_id:expr, $path:expr) => {{
         let result: ApiResult<String> =
         match $info_provider.chain_info($chain_id).await {
-            Ok(chain_info) => Ok(format!("{}{}",chain_info.tx_service_url, $path)),
+            Ok(chain_info) => Ok(format!("{}/api{}",chain_info.transaction_service_url, $path)),
             Err(error) => Err(error,)
         };
         result
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -72,7 +72,7 @@ macro_rules! core_uri {
     ($info_provider:tt, $chain_id:expr, $path:expr) => {{
         let result: ApiResult<String> =
         match $info_provider.chain_info($chain_id).await {
-            Ok(chain_info) => Ok(format!("{}/api{}",chain_info.transaction_service_url, $path)),
+            Ok(chain_info) => Ok(format!("{}/api{}",chain_info.transaction_service, $path)),
             Err(error) => Err(error,)
         };
         result
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -69,16 +69,16 @@ macro_rules! to_hex_string {
 #[doc(hidden)]
 #[macro_export]
 macro_rules! core_uri {
-    ($info_provider:tt, $chain_id:expr, $path:expr) => {{
+    ($info_provider:tt, $path:expr) => {{
         let result: ApiResult<String> =
-        match $info_provider.chain_info($chain_id).await {
+        match $info_provider.chain_info().await {
             Ok(chain_info) => Ok(format!("{}/api{}",chain_info.transaction_service, $path)),
             Err(error) => Err(error,)
         };
         result
     }};
-    ($info_provider:tt, $chain_id:expr, $path:literal, $($arg:tt)*) => {{
+    ($info_provider:tt, $path:literal, $($arg:tt)*) => {{
         let full_path: String = format!($path, $($arg)*);
-        core_uri!($info_provider, $chain_id, full_path)
+        core_uri!($info_provider, full_path)
     }};
 }
--- a/src/models/converters/data_decoded.rs
+++ b/src/models/converters/data_decoded.rs
@@ -19,7 +19,7 @@ impl DataDecoded {
             SET_FALLBACK_HANDLER => {
                 let handler = self.get_parameter_single_value_at(0)?;
                 Some(SettingsInfo::SetFallbackHandler {
-                    handler_info: info_provider.contract_info(chain_id, &handler).await.ok(),
+                    handler_info: info_provider.contract_info(&handler).await.ok(),
                     handler,
                 })
             }
@@ -52,24 +52,21 @@ impl DataDecoded {
             CHANGE_MASTER_COPY => {
                 let implementation = self.get_parameter_single_value_at(0)?;
                 Some(SettingsInfo::ChangeImplementation {
-                    implementation_info: info_provider
-                        .contract_info(chain_id, &implementation)
-                        .await
-                        .ok(),
+                    implementation_info: info_provider.contract_info(&implementation).await.ok(),
                     implementation,
                 })
             }
             ENABLE_MODULE => {
                 let module = self.get_parameter_single_value_at(0)?;
                 Some(SettingsInfo::EnableModule {
-                    module_info: info_provider.contract_info(chain_id, &module).await.ok(),
+                    module_info: info_provider.contract_info(&module).await.ok(),
                     module,
                 })
             }
             DISABLE_MODULE => {
                 let module = self.get_parameter_single_value_at(1)?;
                 Some(SettingsInfo::DisableModule {
-                    module_info: info_provider.contract_info(chain_id, &module).await.ok(),
+                    module_info: info_provider.contract_info(&module).await.ok(),
                     module,
                 })
             }
@@ -167,11 +164,7 @@ async fn insert_value_into_index(
         && value != "0x0000000000000000000000000000000000000000"
         && !index.contains_key(value)
     {
-        if let Some(address_info) = info_provider
-            .full_address_info_search(chain_id, &value)
-            .await
-            .ok()
-        {
+        if let Some(address_info) = info_provider.full_address_info_search(&value).await.ok() {
             index.insert(value.to_owned(), address_info);
         };
     }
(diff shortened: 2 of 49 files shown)


src/routes/collectibles.rs
```diff
@@ -158,18 +157,12 @@ pub async fn get_collectibles(
...
-        exclude_spam.unwrap_or(true)
-    );
-
-    Ok(content::Json(
-        RequestCached::new(url)
-            .request_timeout(collectibles_request_timeout())
-            .execute(context.client(), context.cache())
-            .await?,
-    ))
+    collectibles(
```
jpalvarezl, line 160: I've extracted this into a method to match balances. We are not using `CacheResponse` here, not even prior to this change. We only cache the backend request. Is this ok? I would say so, as we don't map anything.

src/providers/info.rs
```diff
@@ -223,18 +232,20 @@ impl<C: Cache> DefaultInfoProvider<'_, C> {
     }
 
     async fn load_chain_info(&self, chain_id: &str) -> ApiResult<Option<ChainInfo>> {
-        let url = format!("{}/v1/chains/{}/", base_config_service_url(), chain_id);
+        let url = format!("{}/v1/chains/{}", base_config_service_url(), chain_id);
```
jpalvarezl, line 235: Please note the lack of trailing slash. Currently, this would produce a `404` so we need to decide whether we need it or not.

src/models/converters/data_decoded.rs
```diff
@@ -13,12 +13,13 @@ impl DataDecoded {
...
         &self,
         info_provider: &impl InfoProvider,
+        chain_id: &str,
     ) -> Option<SettingsInfo> {
         match self.method.as_str() {
             SET_FALLBACK_HANDLER => {
                 let handler = self.get_parameter_single_value_at(0)?;
                 Some(SettingsInfo::SetFallbackHandler {
-                    handler_info: info_provider.contract_info(&handler).await.ok(),
+                    handler_info: info_provider.contract_info(chain_id, &handler).await.ok(),
```
rmeissner, line 22: Should we initialize the provider with the chain id? this way we don'T have to pass it around

src/providers/info.rs
rmeissner: Maybe we should build the url using some URL utils from Rocket

src/models/converters/data_decoded.rs
jpalvarezl: ok I will try that.

jpalvarezl: I will move this PR back to draft as this change changes substantially how I did everything. I think it would even simplify the macros

jpalvarezl: Having a struct like so:

```rust
pub struct DefaultInfoProvider<'p, C: Cache> {
    pub chain_id: &'p str,
    client: &'p reqwest::Client,
    cache: &'p C,
    // Mutex is an async Mutex, meaning that the lock is non-blocking
    safe_cache: Mutex<HashMap<String, Option<SafeInfo>>>,
    token_cache: Mutex<HashMap<String, Option<TokenInfo>>>,
    chain_cache: Mutex<HashMap<String, Option<ChainInfo>>>,
}
```

Will allow to simplify the `core_uri!` macros further, we don't need to pass the `chain_id` as it is part of the `info_provider`

src/providers/info.rs
```diff
@@ -102,19 +102,27 @@ pub trait InfoProvider {
 }
 
 pub struct DefaultInfoProvider<'p, C: Cache> {
+    pub chain_id: &'p str,
```
jpalvarezl, line 105: @rmeissner I removed the mutex cache, as `chain_id` is now a "constructor" parameter

src/providers/info.rs
```diff
@@ -188,15 +192,19 @@ impl<C: Cache> InfoProvider for DefaultInfoProvider<'_, C> {
...
             cache: context.cache(),
-            chain_cache: Default::default(),
             safe_cache: Default::default(),
             token_cache: Default::default(),
         }
     }
+
+    pub fn new_no_net(context: &'a Context) -> Self {
+        Self::new("No network", context)
+    }
```
jpalvarezl, line 207: There is a need for this method when we have information that is not network independant provided by this dependency. Namely, the fiat exchange rates. Are you ok with this @rmeissner ? It would just be a bit time consuming to adjust the tests again.

jpalvarezl:
If you want to check out this branch, it already "runs" but the tests need to be adjusted

src/providers/info.rs
jpalvarezl: I think that if we want to introduce a tool from an external crate here, something from `reqwest` would be more appropriate? https://docs.rs/reqwest/0.4.0/reqwest/struct.Url.html

src/providers/info.rs
fmrsabino: Wondering if we can make `chain_id` as an optional field in this case. 

Or have a `ChainInfoProvider` and an `InfoProvider` (where `chain_id` is not available). I'd prefer this over `FiatInfoProvider` because having `Fiat` in the name might hint that only fiat is chain independent (where there might be more chain Independent fields that can be represented.

So we'd have `InfoProvider` (with chain independent data) and `ChainInfoProvider → InfoProvider` which includes chain data. Wdyt?

src/macros.rs
```diff
@@ -69,16 +69,16 @@ macro_rules! to_hex_string {
 #[doc(hidden)]
 #[macro_export]
 macro_rules! core_uri {
```
fmrsabino, line 71: nitpick: when I first saw the `core_uri` I thought it'd be something more "core" to Rust URI representations 😬 But this is related to our `transaction_service`. I'd say that renaming this macro to `transaction_service_uri` would make it more explicit (out of scope for this PR).

src/providers/info.rs
jpalvarezl: So that is very similar to what I proposed. In the current state `DefaultInfoProvider` is the `ChainInfoProvider` from your proposal and `InfoProvider` is the `trait` that gets implemented by it. In my proposal `FiatInfoProvider` would be `InfoProvider` of your proposal.

I am just thinking about where we should make the cut, in terms of responsibilities. Do we distinguish chain and non chain data? For me that creates kind of only 2 groups of possible data sources, but if we wanted to have more, like for instance `TokenInfoProvider`, `ContractsInfoProvider`, etc. the `ChainInfoProvider` would get more bloated (kind of what we already have with `DefaultInfoProvider`).

 I would prefer the approach where we split this up by "topic" rather than being stuff that is chain related or not. I am just wondering if there is more fiat related stuff that we could use this for eventually. There is also an argument to be made in terms of "single responsibility" of a class. Admittedly, `DefaultInfoProvider` violates this currently.

src/macros.rs
jpalvarezl: That's a good point! I took that from a naming suggestion from @rmeissner , but I am open to discus this if it brings more clarity to the code base.

src/providers/info.rs
jpalvarezl: I've pushed this approach, seems clean and nice, but if for some reason we prefer the one commented on, we can revert the commit.

src/providers/info.rs
fmrsabino: This was fixed as part of https://github.com/gnosis/safe-config-service/pull/110

Now both versions work with a slight difference: ending with slash hits the endpoint directly while without slash redirects to the endpoint with slash

src/routes/about.rs
```diff
@@ -17,25 +17,26 @@ use rocket::response::content;
...
         .execute(context.cache())
         .await
 }
 
 #[doc(hidden)]
-#[get("/about/backbone")]
-pub async fn backbone(context: Context<'_>) -> ApiResult<content::Json<String>> {
-    let url = format!("{}/v1/about/", base_transaction_service_url());
+#[get("/<chain_id>/about/backbone")]
+pub async fn backbone(context: Context<'_>, chain_id: String) -> ApiResult<content::Json<String>> {
```
fmrsabino, line 37: nitpick: here we can have a string slice right? `chain_id: &str` and then we don't need to do `chain_id.as_str()`

fmrsabino (approved):

src/routes/about.rs
jpalvarezl: Just to be more regular with the types but I think your are right, here we could perfectly just use `&str` . I think it would actually make sense to refactor this as it is more idiomatic. We should only use `String` if we make it mutable. Otherwise we don't really add any value.

PR #438, opened 2021-06-16
Chain Endpoints

### LINKED ISSUE #422: Add `chains_info` endpoints
## Feature

- [x] Fetch and cache `indefinitely` the results returned by https://safe-config.staging.gnosisdev.com/api/v1/chains/
- [x] Expose this endpoint result (and cache it) under an endpoint of our own `https://safe-client.staging.gnosisdev.com/api/v1/chains/`

## Open questions

- ~https://safe-config.staging.gnosisdev.com/api/v1/chains/ endpoint is paginated. Should we exhaust all the pages of data when we request this information?~
  - See comments
- The chain info (either coming from the cache or config service) should be requested on every incoming request to the client GW that is network specific?
  - Yes, this should be cached in redis and if possible in memory

jpalvarezl:
Closes #422 

- Added a forward endpoint for `/v1/chains`
- Added an endpoint returning the `ChainInfo` of a single chain requiring the the `chain_id` as input

### DIFF
--- a/src/routes/chains.rs
+++ b/src/routes/chains.rs
@@ -25,8 +25,8 @@ pub async fn get_chain(context: Context<'_>, chain_id: String) -> ApiResult<cont
     CacheResponse::new(context.uri())
         .duration(chain_info_cache_duration())
         .resp_generator(async || {
-            let info_provider = DefaultInfoProvider::new(&context);
-            info_provider.chain_info(&chain_id).await
+            let info_provider = DefaultInfoProvider::new(&chain_id, &context);
+            info_provider.chain_info().await
         })
         .execute(context.cache())
         .await
@@ -47,10 +47,14 @@ pub async fn get_chain(context: Context<'_>, chain_id: String) -> ApiResult<cont
  */
 #[get("/v1/chains")]
 pub async fn get_chains(context: Context<'_>) -> ApiResult<content::Json<String>> {
-    let url = format!("{}/v1/chains", base_config_service_url(),);
+    let mut url = reqwest::Url::parse(base_config_service_url().as_str())
+        .expect("Bad base config service url");
+    url.path_segments_mut()
+        .expect("Cannot add chain_id to path")
+        .extend(["v1", "chains"]);
 
     Ok(content::Json(
-        RequestCached::new(url)
+        RequestCached::new(url.to_string())
             .request_timeout(chain_info_request_timeout())
             .cache_duration(chain_info_cache_duration())
             .execute(context.client(), context.cache())
--- a/src/routes/chains.rs
+++ b/src/routes/chains.rs
@@ -11,7 +11,7 @@ use rocket::response::content;
  * `/v1/chains/<chain_id>/` <br/>
  * Returns [ChainInfo](crate::models::chains::ChainInfo)
  *
- * # Balances
+ * # Chains
  *
  * This endpoint returns the [ChainInfo](crate::models::chains::ChainInfo) for a given `chainId`
  *
@@ -36,7 +36,7 @@ pub async fn get_chain(context: Context<'_>, chain_id: String) -> ApiResult<cont
  * `/v1/chains/` <br/>
  * Returns a [Page](crate::models::commons::Page) of [ChainInfo](crate::models::chains::ChainInfo)
  *
- * # Balances
+ * # Chains
  *
  * Returns a paginated list of all the supported [ChainInfo](crate::models::chains::ChainInfo)
  *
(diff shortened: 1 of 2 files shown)


src/routes/chains.rs
```diff
@@ -0,0 +1,59 @@
...
+ * Returns a paginated list of all the supported [ChainInfo](crate::models::chains::ChainInfo)
+ *
+ * ## Path
+ *
+ * - `/v1/chains/` Returns the `ChainInfo` for our services supported networks
+ *
+ */
+#[get("/v1/chains")]
+pub async fn get_chains(context: Context<'_>) -> ApiResult<content::Json<String>> {
+    let url = format!("{}/v1/chains", base_config_service_url(),);
```
jpalvarezl, line 50: I will adjust this accordingly to match what we decide here: https://github.com/gnosis/safe-client-gateway/pull/435#discussion_r651818592

src/routes/chains.rs
```diff
@@ -0,0 +1,59 @@
...
+ *
+ * This endpoint returns the [ChainInfo](crate::models::chains::ChainInfo) for a given `chainId`
+ *
+ * ## Path
+ *
+ * - `/v1/chains/<chain_id>/`returns the `ChainInfo` for `<chain_id>`
+ *
+ */
+#[get("/v1/chains/<chain_id>")]
+pub async fn get_chain(context: Context<'_>, chain_id: String) -> ApiResult<content::Json<String>> {
```
fmrsabino, line 24: Regarding the type of `chain_id` if it makes things easier for validation on the CGW we can assume this to be a Big Int. This is how it's represented internally in the safe-config even though it's exposed as a string (this is how the framework maps a BigInt to a Json field). 

But if this is mostly used for the path I'd say it should be fine – I don't think we need to validate it. Eg.: giving as a `chain_id = foo` should return a `404` from the config service and the CGW would propagate that.

src/routes/chains.rs
```diff
@@ -0,0 +1,59 @@
...
+use crate::providers::info::{DefaultInfoProvider, InfoProvider};
+use crate::utils::context::Context;
+use crate::utils::errors::ApiResult;
+use rocket::response::content;
+
+/**
+ * `/v1/chains/<chain_id>/` <br/>
+ * Returns [ChainInfo](crate::models::chains::ChainInfo)
+ *
+ * # Balances
```
fmrsabino, line 14: Here it should `Chain Info` right? 😅

src/routes/chains.rs
```diff
@@ -0,0 +1,59 @@
...
+        })
+        .execute(context.cache())
+        .await
+}
+
+/**
+ * `/v1/chains/` <br/>
+ * Returns a [Page](crate::models::commons::Page) of [ChainInfo](crate::models::chains::ChainInfo)
+ *
+ * # Balances
```
fmrsabino, line 39: Same as above ☝️

fmrsabino (changes requested):

src/routes/chains.rs
jpalvarezl: Ah yeah, hasty copy pasting! Thanks!

src/routes/chains.rs
jpalvarezl: Actually, there is one place where I wish it was just a plain number, which is when we calculate the `safeTxHash` but we just parse it:

https://github.com/gnosis/safe-client-gateway/blob/34e751b096b108174be8b4f8c4d8d0c2475f1433/src/utils/transactions.rs#L84

it's either way fine :) I don't see this as a solid argument against having it as a string :thinking: In fact, I am not sure if `chain_id`s are guaranteed to always be numbers, maybe at some point people will start putting all sorts of chars there :thinking:

fmrsabino: I was looking at EIP155 https://github.com/ethereum/EIPs/blob/master/EIPS/eip-155.md

And there it's used with `{0,1} + CHAIN_ID * 2 + 35` so I assume it should only be represented as a number. But this was more to check if on your side it'd make things easier as I'd say we can assume it as a number (else we can just use string) 👍

jpalvarezl: Ah that's good to know! If there is also a chance that the `chain_id` could potentially be a big number, which could trip up our (de)serializers too :thinking: Let's keep it as a `String` as it seems like the option with the fewest assumptions on the values :)

fmrsabino: Internally in the DB it's also `uint64` which I think it's ok given the current values that we see in the chains 👍 It's serialised as `String` so even if we increase the size of the parameter in the DB there are no changes in the API contract

fmrsabino (approved):

PR #439, opened 2021-06-16
Forward endpoints

### LINKED ISSUE #423: Forward endpoints
## Context

We need provide a forwarding endpoint for the following core services endpoints:
- `GET /about/master-copies` - cached indefinitely 
- `POST /data-decoder/` - not cached at all
- `POST /safes/{address}/multisig-transactions/estimations/` not cached

## Feature

The client gateway should simply create forward endpoints. Namely, use the same http method and models as the core services.

Each endpoint should be exposed as following:
- `GET /{chain_id}/about/master-copies`
- `POST /{chain_id}/data-decoder/` (probably `chain_id` is ignored at first)
- `POST /{chain_id}/safes/{address}/multisig-transactions/estimations/` not cached

## Open questions

- Duration of caches should be evaluated.

jpalvarezl:
Closes #423 

Changes:
- Added `data_decoder` endpoint
- Added `gas_estimation` endpoint
- Added `master_copies` endpoint

### DIFF
--- a/src/routes/utils.rs
+++ b/src/routes/utils.rs
@@ -1,6 +1,5 @@
-use crate::models::service::utils::{
-    DataDecoderRequest, SafeTransactionEstimation, SafeTransactionEstimationRequest,
-};
+use crate::models::service::utils::{DataDecoderRequest, SafeTransactionEstimationRequest};
+use crate::services::utils;
 use crate::services::utils::request_data_decoded;
 use crate::utils::context::Context;
 use crate::utils::errors::ApiResult;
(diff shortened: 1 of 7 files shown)


src/routes/utils.rs
```diff
@@ -0,0 +1,121 @@
+use crate::models::service::utils::{
```
rmeissner, line 1: I find this file name very confusing, as I would expect utility methods here

jpalvarezl: that's my understanding of the gas estimation and a data decoder endpoint. How would you classify these endpoints?

rmeissner: Ibmean these are endpoint methods which is something else for me than utils :D

The estimation endpoint is a safe related endpoint.

The master copies could be part of the about.

And data decoding could be a separate module ... Not sure about this one.

rmeissner (approved):

PR #447, opened 2021-06-24
Add Gas Price to /v1/chains/<chain_id>

### LINKED ISSUE #444: Add Gas Price Oracle to ChainInfo
The Safe Config Service now returns the chain gas price oracle for each chain:

```javascript
{
  "gasPriceOracle": {
    "url": <string>,
    "gasParameter": <string>
  }
}
```

The gas price oracle should be exposed to the clients via GET /v1/chains/<chain_id>

fmrsabino:
Closes #444 

- Add Gas Price to `ChainInfo`

### DIFF
--- a/src/models/backend/chains.rs
+++ b/src/models/backend/chains.rs
@@ -12,6 +12,7 @@ pub struct ChainInfo {
     pub native_currency: NativeCurrency,
     pub theme: Theme,
     pub ens_registry_address: Option<String>,
+    pub gas_price: GasPrice,
 }
 
 #[derive(Deserialize, Debug, PartialEq, Clone)]
@@ -28,4 +29,18 @@ pub struct NativeCurrency {
 pub struct Theme {
     pub text_color: String,
     pub background_color: String,
+}
+
+#[derive(Deserialize, Debug, PartialEq, Clone)]
+#[serde(tag = "type")]
+#[serde(rename_all = "lowercase")]
+pub enum GasPrice {
+    #[serde(rename_all = "camelCase")]
+    Oracle {
+        url: String,
+        gas_parameter: Option<String>,
+        gwei_factor: String,
+    },
+    #[serde(rename_all = "camelCase")]
+    Fixed { wei_value: String },
 }
--- a/src/models/backend/chains.rs
+++ b/src/models/backend/chains.rs
@@ -43,4 +43,6 @@ pub enum GasPrice {
     },
     #[serde(rename_all = "camelCase")]
     Fixed { wei_value: String },
+    #[serde(other)]
+    Unknown,
 }
--- a/src/models/backend/chains.rs
+++ b/src/models/backend/chains.rs
@@ -38,7 +38,7 @@ pub enum GasPrice {
     #[serde(rename_all = "camelCase")]
     Oracle {
         url: String,
-        gas_parameter: Option<String>,
+        gas_parameter: String,
         gwei_factor: String,
     },
     #[serde(rename_all = "camelCase")]
--- a/src/models/service/chains.rs
+++ b/src/models/service/chains.rs
@@ -3,7 +3,8 @@ use serde::Serialize;
 #[derive(Serialize, Debug, PartialEq, Clone)]
 #[serde(rename_all = "camelCase")]
 pub struct ChainInfo {
-    pub transaction_service: String, // do we need to expose this?
+    pub transaction_service: String,
+    // do we need to expose this?
     pub chain_id: String,
     pub chain_name: String,
     pub rpc_url: String,
@@ -11,6 +12,7 @@ pub struct ChainInfo {
     pub native_currency: NativeCurrency,
     pub theme: Theme,
     pub ens_registry_address: Option<String>,
+    pub gas_price: GasPrice,
 }
 
 #[derive(Serialize, Debug, PartialEq, Clone)]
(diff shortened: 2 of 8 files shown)


src/models/chains.rs
```diff
@@ -18,3 +20,17 @@ pub struct NativeCurrency {
...
+pub struct Theme {
+    pub text_color: String,
+    pub background_color: String,
+}
+
+#[derive(Serialize, Deserialize, Debug, PartialEq, Clone)]
+#[serde(rename_all = "camelCase")]
+pub struct GasPriceOracle {
+    pub url: String,
+    pub gas_parameter: Option<String>,
```
rmeissner, line 35: Is this the only format we will have? 

Afaik the web version also supports having a constant value set. And there was the discussion about a multiplier for the oracle solution.

rmeissner (changes requested):

fmrsabino:
Blocked by: https://github.com/gnosis/safe-config-service/issues/87#issuecomment-872203630

src/models/chains.rs
jpalvarezl: If that is the situation we should probably change this from a `struct` into a `enum` and make it polymorphic.

fmrsabino: The `gas_parameter` can have any `String` value (it'd depend on the oracle itself)

fmrsabino:
Config Service PR: https://github.com/gnosis/safe-config-service/pull/146

src/models/chains.rs
```diff
@@ -27,3 +28,11 @@ pub struct Theme {
     pub text_color: String,
     pub background_color: String,
 }
+
+#[derive(Serialize, Deserialize, Debug, PartialEq, Clone)]
```
jpalvarezl, line 32: We probably don't need `Clone`, right?

jpalvarezl (approved):
LGTM , I might get rid of `Clone` derive if it's not needed, but not a mandatory change :)

src/models/chains.rs
```diff
@@ -28,3 +29,11 @@ pub struct Theme {
     pub text_color: String,
     pub background_color: String,
 }
+
+#[derive(Serialize, Deserialize, Debug, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub struct GasPriceOracle {
+    pub url: String,
+    pub gas_parameter: Option<String>,
+    pub gwei_factor: String,
```
jpalvarezl, line 38: The `gwei_factor` is always present or does it have any default value ?

fmrsabino: Always present. Default value from the config service is `"1.000000000"`

jpalvarezl (approved):
LGTM, although I am not entirely aware of the web team expectation. If that's all covered then I would say we are good to merge.

fmrsabino:
@rmeissner I need your review 😄 🙏

src/models/service/chains.rs
```diff
@@ -28,3 +30,17 @@ pub struct Theme {
     pub text_color: String,
     pub background_color: String,
 }
+
+#[derive(Serialize, Debug, PartialEq, Clone)]
+#[serde(tag = "type")]
+#[serde(rename_all = "lowercase")]
```
jpalvarezl, line 36: Clients generally expect enum variants as `SCREAMING_SNAKE_CASE`

src/models/backend/chains.rs
```diff
@@ -28,3 +29,17 @@ pub struct Theme {
...
+#[serde(rename_all = "lowercase")]
+pub enum GasPrice {
+    #[serde(rename_all = "camelCase")]
+    Oracle {
+        url: String,
+        gas_parameter: Option<String>,
+        gwei_factor: String,
+    },
+    #[serde(rename_all = "camelCase")]
+    Fixed { wei_value: String },
```
jpalvarezl, line 44:
```suggestion
    Fixed { wei_value: String },
    #[serde(other)]
    Unknown,
```

Just to protect ourselves from ourselves in the future :D

jpalvarezl (approved):
I would approve as to not block you, but I think it would be good to implement my suggestions. `serde tag` will try to match the underlying structure of the variant, so testing for `unknown` would be nice too. Just by adding some wacky structure.

src/models/service/chains.rs
fmrsabino: Good catch. The config service exposes this as a lower case but this can be different when serialising

fmrsabino:
> I would approve as to not block you, but I think it would be good to implement my suggestions. `serde tag` will try to match the underlying structure of the variant, so testing for `unknown` would be nice too. Just by adding some wacky structure.

Good point. Will add a test for this 👍

jpalvarezl (approved):

src/models/backend/chains.rs
```diff
@@ -28,3 +29,19 @@ pub struct Theme {
...
 }
+
+#[derive(Deserialize, Debug, PartialEq, Clone)]
+#[serde(tag = "type")]
+#[serde(rename_all = "lowercase")]
+pub enum GasPrice {
+    #[serde(rename_all = "camelCase")]
+    Oracle {
+        url: String,
+        gas_parameter: Option<String>,
```
rmeissner, line 40: Why is this optional? I was thinking about if there is a default used by the clients :thinking: or does `null` imply something else on config service side?

In the typescript definition it is not nullable: https://github.com/gnosis/safe-react/blob/dev/src/config/networks/network.d.ts#L74

fmrsabino: It's optional (safe-config side) https://github.com/gnosis/safe-config-service/issues/87#issuecomment-865994250. I can change that.

fmrsabino: Since this matches what the safe-config service returns I'd say to tackle this in a different PR

fmrsabino: Gas parameter is now required: 86fb7a5

fmrsabino:
Conflicts fixed cc/ @jpalvarezl @rmeissner

rmeissner (approved):

PR #448, opened 2021-06-25
Use ChainInfo for pagination

### LINKED ISSUE #445: Use ChainInfo for /v1/chains endpoint
Currently `ChainInfo` is only being used for `/v1/chains/<chain_id>`. `/v1/chains` returns a copy of payload returned by the Safe Config Service however, some data should be formatted (eg.: data that is only used by the client gateway and should not be exposed to the clients).

Therefore `ChainInfo` should also be used to represent the chains present in `/v1/chains`. The pagination should remain the same.

fmrsabino:
Closes #445

- `/v1/chains` now uses `ChainInfo` in order to parse and validate the returned payload by the Safe Config Service

### DIFF
--- a/src/routes/chains.rs
+++ b/src/routes/chains.rs
@@ -1,8 +1,7 @@
-use crate::cache::cache_operations::{CacheResponse, RequestCached};
-use crate::config::{
-    base_config_service_url, chain_info_cache_duration, chain_info_request_timeout,
-};
+use crate::cache::cache_operations::CacheResponse;
+use crate::config::chain_info_cache_duration;
 use crate::providers::info::{DefaultInfoProvider, InfoProvider};
+use crate::services::chains::get_chains_paginated;
 use crate::utils::context::Context;
 use crate::utils::errors::ApiResult;
 use rocket::response::content;
@@ -45,19 +44,14 @@ pub async fn get_chain(context: Context<'_>, chain_id: String) -> ApiResult<cont
  * - `/v1/chains/` Returns the `ChainInfo` for our services supported networks
  *
  */
-#[get("/v1/chains")]
-pub async fn get_chains(context: Context<'_>) -> ApiResult<content::Json<String>> {
-    let mut url = reqwest::Url::parse(base_config_service_url().as_str())
-        .expect("Bad base config service url");
-    url.path_segments_mut()
-        .expect("Cannot add chain_id to path")
-        .extend(["v1", "chains"]);
-
-    Ok(content::Json(
-        RequestCached::new(url.to_string())
-            .request_timeout(chain_info_request_timeout())
-            .cache_duration(chain_info_cache_duration())
-            .execute(context.client(), context.cache())
-            .await?,
-    ))
+#[get("/v1/chains?<limit>&<offset>")]
+pub async fn get_chains(
+    context: Context<'_>,
+    limit: Option<String>,
+    offset: Option<String>,
+) -> ApiResult<content::Json<String>> {
+    CacheResponse::new(context.uri())
+        .resp_generator(|| get_chains_paginated(&context, limit.as_ref(), offset.as_ref()))
+        .execute(context.cache())
+        .await
 }
--- a/src/routes/chains.rs
+++ b/src/routes/chains.rs
@@ -44,14 +44,10 @@ pub async fn get_chain(context: Context<'_>, chain_id: String) -> ApiResult<cont
  * - `/v1/chains/` Returns the `ChainInfo` for our services supported networks
  *
  */
-#[get("/v1/chains?<limit>&<offset>")]
-pub async fn get_chains(
-    context: Context<'_>,
-    limit: Option<String>,
-    offset: Option<String>,
-) -> ApiResult<content::Json<String>> {
+#[get("/v1/chains")]
+pub async fn get_chains(context: Context<'_>) -> ApiResult<content::Json<String>> {
     CacheResponse::new(context.uri())
-        .resp_generator(|| get_chains_paginated(&context, limit.as_ref(), offset.as_ref()))
+        .resp_generator(|| get_chains_paginated(&context))
         .execute(context.cache())
         .await
 }
--- a/src/routes/chains.rs
+++ b/src/routes/chains.rs
@@ -44,10 +44,13 @@ pub async fn get_chain(context: Context<'_>, chain_id: String) -> ApiResult<cont
  * - `/v1/chains/` Returns the `ChainInfo` for our services supported networks
  *
  */
-#[get("/v1/chains")]
-pub async fn get_chains(context: Context<'_>) -> ApiResult<content::Json<String>> {
+#[get("/v1/chains?<limit>")]
+pub async fn get_chains(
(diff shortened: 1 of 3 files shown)


src/services/chains.rs
```diff
@@ -0,0 +1,36 @@
...
+    base_config_service_url, chain_info_cache_duration, chain_info_request_timeout,
+};
+use crate::models::chains::ChainInfo;
+use crate::models::commons::Page;
+use crate::utils::context::Context;
+use crate::utils::errors::ApiResult;
+
+pub async fn get_chains_paginated(
+    context: &Context<'_>,
+    limit: Option<&String>,
```
rmeissner, line 12: In all our current views we do not expose `limit`and `offset`. 

While in this case it would not really pose an issue I like that the clients don't make any assumptions on how the pagination works. They should just assume that there is a "cursor" in the url ... and even that they should probably never know as they should just use the returned `next` url. See https://github.com/gnosis/safe-client-gateway/blob/main/src/routes/transactions.rs#L112

src/services/chains.rs
```diff
@@ -0,0 +1,36 @@
...
+    offset: Option<&String>,
+) -> ApiResult<Page<ChainInfo>> {
+    let mut queries: Vec<(String, String)> = vec![];
+    if let Some(limit) = limit {
+        queries.push(("limit".to_string(), limit.to_owned()))
```
rmeissner, line 17: I think we always use `to_string` instead of `to_owned`. We should stick to this (at least in this method :P )

src/services/chains.rs
```diff
@@ -0,0 +1,36 @@
...
+    }
+    if let Some(offset) = offset {
+        queries.push(("offset".to_string(), offset.to_owned()))
+    }
+    let mut url = reqwest::Url::parse_with_params(base_config_service_url().as_str(), queries)
```
rmeissner, line 22: we could import `Url` to avoid that we have to specify the `reqwest` module. (There are other places too where we should do that)

src/services/chains.rs
```diff
@@ -0,0 +1,36 @@
...
+        .expect("Cannot add chain_id to path")
+        .extend(["v1", "chains"]);
+
+    let body = RequestCached::new(url.to_string())
+        .request_timeout(chain_info_request_timeout())
+        .cache_duration(chain_info_cache_duration())
+        .execute(context.client(), context.cache())
+        .await?;
+
+    let page: Page<ChainInfo> = serde_json::from_str::<Page<ChainInfo>>(&body)?;
```
rmeissner, line 34: In a future PR we should decouple the internal model (the one fetched from the config service)  and the external model (the one returned to the clients). Maybe we should open an issue for that

src/services/chains.rs
fmrsabino: I agree. I will remove this. This can be easily adjusted if required 👍

src/services/chains.rs
fmrsabino: Created here #451

src/services/chains.rs
jpalvarezl: I would also use `HashMap` here and maybe `<&str, String>` instead, so you don't have have to call `to_string` on the key too.

src/services/chains.rs
fmrsabino: This was added back because `limit` is being used by the clients

src/services/chains.rs
fmrsabino: Changed to `HashMap` 👍 bde0d85

src/routes/chains.rs
```diff
@@ -45,19 +44,13 @@ pub async fn get_chain(context: Context<'_>, chain_id: String) -> ApiResult<cont
...
-        RequestCached::new(url.to_string())
-            .request_timeout(chain_info_request_timeout())
-            .cache_duration(chain_info_cache_duration())
-            .execute(context.client(), context.cache())
-            .await?,
-    ))
+#[get("/v1/chains?<limit>")]
+pub async fn get_chains(
+    context: Context<'_>,
+    limit: Option<String>,
```
jpalvarezl, line 50: We should probably also take a note in #451 to get rid of `limit` . Generally, like we did for the transaction endpoints, we provide a `page_url` where the clients inject whatever we return in `next` and `previous` in the this struct: 

```pub struct Page<T> {
    pub next: Option<String>,
    pub previous: Option<String>,
    pub results: Vec<T>,
}
```
This would allow us to migrate to cursor based pagination eventually, if we decide on it.

src/services/chains.rs
```diff
@@ -0,0 +1,34 @@
...
+use std::collections::HashMap;
+
+pub async fn get_chains_paginated(
+    context: &Context<'_>,
+    limit: &Option<String>,
+) -> ApiResult<Page<ChainInfo>> {
+    let mut queries: HashMap<&str, String> = HashMap::new();
+    if let Some(limit) = limit {
+        queries.insert("limit", limit.to_string());
+    }
```
jpalvarezl, line 19: Like I mentioned above, maybe it is worth building some `structs` and methods that adjust the pagination from the config and core services to the gateway's. Totally out of the scope of this PR though :)

jpalvarezl (approved):
I think this looks good! We should probably adjust having `limit` to `page_url` and do the same we do for transactions. If it's simple and you want to do it in this PR go for it, otherwise we can include this change in #451 or even open a separate issue for it.

src/routes/chains.rs
jpalvarezl: I am thinking that it would probably be enough to simply renaming this param to `page_url` and then simply storing `limit=10`  or `limit=10&offset=10` in `next` and `previous` in the response we return

jpalvarezl:
Note: I noticed that one of the proposed approaches in the mobile refinement meeting was to fetch the data with `limit=100` which doesn't seem great, although that's based on a context-less opinion.

If that became a hard requirement, I don't think we could use the approach with `page_url` we have for the `transaction` endpoints.

jpalvarezl (changes requested):
As mentioned on a previous comment, we should align with the result of the mobile refinement meeting.

fmrsabino:
> Note: I noticed that one of the proposed approaches in the mobile refinement meeting was to fetch the data with `limit=100` which doesn't seem great, although that's based on a context-less opinion.
> 
> If that became a hard requirement, I don't think we could use the approach with `page_url` we have for the `transaction` endpoints. In which case, I would think that we should either:
> A) Remove pagination entirely at the config service level
> B) Hardcode the `limit=100` in the gateway, removing the `limit` parameter, so that the clients don't touch it.
> 
> I will change my review to "requires changes" to prevent accidental merges, but I believe this issues should be resolved before we know which direction to take.

I wouldn't say that the `limit=100` is really applicable in our scenario. Actually the max limit of the configuration service is `10`. If a bigger value is given it still returns 10 chains.

For now I would say that this can work but a new endpoint should be provided to the clients – they sync all the chain data for all the safes that they have. Since no endpoint like that exists currently they dismiss the pagination part and fetch "all the chains".

jpalvarezl (approved):

PR #452, opened 2021-07-01
Remove Native Coin config helpers

### LINKED ISSUE #436: [Forked] Remove NATIVE_COIN related env variable usages in favour of `ChainInfo`
```
## Miscellaneous config
#NATIVE_COIN_DECIMALS=18
#NATIVE_COIN_SYMBOL=ETH
#NATIVE_COIN_NAME=Ether
```

Are removed from the env but the methods in the `config` module, through which they are exposed, are still in use.

These methods should be removed in favour of using `ChainInfo` from the config service.

fmrsabino:
Closes #436 

- Remove `native_coin_decimals`, `native_coin_symbol` and `native_coin_name`
- These functions were reading from the environment variables which no longer apply to the interstellar release

### DIFF
--- a/src/services/balances.rs
+++ b/src/services/balances.rs
@@ -38,12 +38,14 @@ pub async fn balances(
         .await
         .unwrap_or(0.0);
 
+    let native_currency = info_provider.chain_info().await?.native_currency;
+
     let mut total_fiat = 0.0;
 
     let mut service_balances: Vec<Balance> = backend_balances
         .into_iter()
         .map(|it| {
-            let balance = it.to_balance(usd_to_fiat);
+            let balance = it.to_balance(usd_to_fiat, &native_currency);
             total_fiat += balance.fiat_balance.parse::<f64>().unwrap_or(0.0);
             balance
         })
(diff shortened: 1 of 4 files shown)


src/services/balances.rs
```diff
@@ -38,12 +38,14 @@ pub async fn balances(
         .await
         .unwrap_or(0.0);
 
+    let native_currency = info_provider.chain_info().await?.native_currency;
```
fmrsabino, line 41: Maybe a default should be provided here wdyt? 🤔

jpalvarezl: I would say yes. As currently and error in the response for this request would result in the gw itself propagating the error invalidating the rest of the data. Not sure what a sensible default would be. 

We could either use
```
#NATIVE_COIN_DECIMALS=18
#NATIVE_COIN_SYMBOL=ETH
#NATIVE_COIN_NAME=Ether
```

Or replace `ETH` and `Ether` with `NC` and `Native Coind` respectively (not so happy about this as something called `NC` may already exist :thinking: )

fmrsabino: Added Ether as default value 5630708

jpalvarezl (approved):
Looks good! I double checked the `.env.example` and the `README` there doesn't seem to be any need for further cleanup.

Issue #455, opened 2021-07-05
Unregister device for push notifications


Uxio0:
**Endpoint URL**
`DELETE /v1/chains/<chain_id>/notifications/devices/<uuid>/safes/<address>/`

**Proposed payload**
No needed

**How it works**
It should relay the request to the corresponding transaction service

fmrsabino:
Wondering about the path itself:

1) `DELETE /notifications/devices/{uuid}/safes/{chainId}/{address}/` (current proposal)
2)  `DELETE /notifications/devices/{uuid}/{chainId}/safes/{address}/`

in `1` we have the tuple `chainId-safeAddress` to identify a safe under `/safes` which makes a lot of sense specially in scenarios where the safe address is most of the times the same across different chains (given the same configuration)

In `2` we'd follow a more hierarchical approach where under the `/notifications/devices/{uuid}` resource: in order to identify the safe we'd have to know the `chainId` first and then under the `/safes` resource we'd identify the safe that we want with its address.

Nothing critical but wondering about the consistency of our endpoint structure so it matches how we address specific safes in different endpoints.

jpalvarezl:
I just opened a ticket in order to address this: #457 

I think that the solution I've seen that most of us agree on being the nicest one would look something like:

`DELETE /v1/chains/<chain_id>/notifications/devices/<uuid>/safes/<address>`

I suppose we can wait until today's sync to somewhat "set this on stone" though :)

Uxio0:
> Wondering about the path itself:
> 
> 1. `DELETE /notifications/devices/{uuid}/safes/{chainId}/{address}/` (current proposal)
> 2. `DELETE /notifications/devices/{uuid}/{chainId}/safes/{address}/`

Sorry, it was not the proposal, I just copied the transaction service endpoint, that's why I added `TBD` 😉

Issue #454, opened 2021-07-05
Register device for push notifications


Uxio0:
# **Endpoint URL**
`/v1/register/notifications`

# **Proposed payload**

## Request object:
```json
{
   "uuid":"<str | null>",
   "cloudMessagingToken":"<str>",
   "buildNumber":"<sematic-version-str>",
   "bundle":"<str>",
   "deviceType":"<str>",
   "version":"ANDROID|IOS|WEB",
   "timestamp":"<iso-str | null>",
   "safeRegistrations":[
      {
         "chainId":"<stringyfied-int>",
         "safes":[
            "<checksummed-address>"
         ],
         "signatures": ["<str-hex>"]
      }
   ]
}
```

## Response 

No returned object

# **Open for discussion**
- Signing `chainId`? 
- No as the `chainId` is part of the payload, this is not necessary

# **Related**
- Signature calculation: https://github.com/gnosis/safe-transaction-service/issues/190

**Edits**
by @jpalvarezl : changed the type of `chainId` from `<int>` to `<stringyfied-int>`
by @jpalvarezl : added endpoint url
by @jpalvarezl : add result payload
by @jpalvarezl : adjusted nullability of fields
by @jpalvarezl : response section adjusted (no object returned) and array of signature: decisions from refinement

biafra23:
Sample JSON:
```
{
  "uuid": "c50750df-700c-4b17-98ca-b95a5c27ca18",
  "safes": [
    {
      "address": "0x00E17aA063fbDB3BFdEfc2c3b2c13173d2711a35",
      "chainId": 1,
      "signature": "0x4b574e7c729db54b427dd17a6b2ae3481221642a9d61c52a53f77500d98ddc1d739c39dfb117619fb09a20e3f5070d018e62c37f89fb622ae10b56a6be9af5c11b"
    },
    {
      "address": "0x00E17aA063fbDB3BFdEfc2c3b2c13173d2711a36",
      "chainId": 4,
      "signature": "0x4b574e7c729db54b427dd17a6b2ae3481221642a9d61c52a53f77500d98ddc1d739c39dfb117619fb09a20e3f5070d018e62c37f89fb622ae10b56a6be9af5c11b"
    }
  ],
[...]

Issue #440, opened 2021-06-17
Include `chain_id` in payloads for webhooks


jpalvarezl:
## Feature

Add the `chain_id` to the fields of payloads for a more fined grained invalidation.

## Proposal

Fetch the `ChainInfo` to have the invalidation prefix for the cache.

jpalvarezl:
This will be addressed correctly later as we need to discuss how we will re-construct cache keys and this doesn't affect the production release too much (as we would only invalidate entries which contain either `safeAddress`es or `safeTxHash`es which would be duplicate across networks)

PR #464, opened 2021-07-07
Normalise endpoints

### LINKED ISSUE #457: Normalise chain dependent paths
## Proposal

For ease of parsing for kibana, readability and restful complaince all chain dependent endpoints should be of the form:

`/v1/chains/<chain_id>/safes/...` 
`/v1/chains/<chain_id>/transactions/...` 
 `etc...`

More generally speaking we want to compose endpoints like so:

`/<<version>>/chains/<chain_id>/<<rest_of_the_old_path>>`

jpalvarezl:
Closes #457 

Changes: Endpoints follow the pattern `/v1/chains/<chaind_id>/...rest_of_the_old_endpoint`

### DIFF
--- a/src/routes/about.rs
+++ b/src/routes/about.rs
@@ -8,7 +8,7 @@ use crate::utils::errors::ApiResult;
 use rocket::response::content;
 
 /**
- * `/<chain_id>/about` <br />
+ * `/v1/chains/<chain_id>/about` <br />
  * Returns [About](crate::models::service::about::About)
  *
  * # About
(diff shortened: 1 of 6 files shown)


rmeissner (approved):
The changes look good, but I did not check if these were all the changes required

jpalvarezl:
I think we are still missing a couple of new endpoints. Such as the notification related (#455 #454 ) and also webhooks #440 

I think that would account for all that's left.

rmeissner:
For webhooks we should not change the url, right? The chain id isnpart of the post payload in that case

jpalvarezl:
yeah, we just need to adjust the payload and that should be it.

jpalvarezl:
something like `*/chains/<chain_id>*<safe_address>*`

PR #466, opened 2021-07-07
Add currency logo field to ChainInfo

### LINKED ISSUE #461: Mobile clients need currency logo for native currencies in /v1/chains endpoint
ATM the `/v1/chains` endpoint returns the following for native currencies:
```
"nativeCurrency": {
       "name": <string>, // optional
       "symbol": <string>,
       "decimals": <uint>,
   }
```
As documented here: https://github.com/gnosis/safe-config-service/issues/85

We would like to have an additional field in that structure called logoUri or logoUrl:
```
"nativeCurrency": {
       "name": <string>, // optional
       "symbol": <string>,
       "decimals": <uint>,
       "logoUri": <string>
   }
```
Alternatively the mobile clients could use one image for all currencies.

fmrsabino:
Closes #461 

- Consume and expose the `logo_url` from the config service to the clients

### DIFF
--- a/src/services/balances.rs
+++ b/src/services/balances.rs
@@ -45,6 +45,7 @@ pub async fn balances(
             name: "Ether".to_string(),
             symbol: "ETH".to_string(),
             decimals: 18,
+            logo_url: "https://cryptologos.cc/logos/ethereum-eth-logo.png".to_string(),
         },
     };
 
--- a/src/services/balances.rs
+++ b/src/services/balances.rs
@@ -39,15 +39,7 @@ pub async fn balances(
         .await
         .unwrap_or(0.0);
 
-    let native_currency: NativeCurrency = match info_provider.chain_info().await {
-        Ok(chain_info) => chain_info.native_currency,
-        Err(_) => NativeCurrency {
-            name: "Ether".to_string(),
-            symbol: "ETH".to_string(),
-            decimals: 18,
-            logo_url: "https://cryptologos.cc/logos/ethereum-eth-logo.png".to_string(),
-        },
-    };
+    let native_currency: NativeCurrency = info_provider.chain_info().await?.native_currency;
 
     let mut total_fiat = 0.0;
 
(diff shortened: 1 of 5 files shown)


fmrsabino:
This can be merged after the field is added to the config service: https://github.com/gnosis/safe-config-service/pull/142

src/services/balances.rs
```diff
@@ -45,6 +45,7 @@ pub async fn balances(
             name: "Ether".to_string(),
             symbol: "ETH".to_string(),
             decimals: 18,
+            logo_url: "https://cryptologos.cc/logos/ethereum-eth-logo.png".to_string(),
```
fmrsabino, line 48: This was the default provided by Tobias. I'd prefer however if this would be a resource hosted by Gnosis in this case as this would be the default if the Chain Native Currency cannot be provided 🤔 

Another option would be to actually propagate the error if no native currency is provided since the spec specifies it as required. Defaulting to ETH might be the wrong assumption.

jpalvarezl: I don't think this is my call, but my rational behind having a default value in the gateway is that in case we do have a balance object to return to the client but for some reason the native currency is not available, we would simply just show the wrong native coin info rather than discarding the entire response. 

It's similar to how we handle fiat conversions. If we have no conversion value, we simple set it to `0` instead of `panic`ing in the gw. Applying this rational to the native coin situation, seeing `ETH` in a network that doesn't correspond signals an error that we can (hopefully) easily address.

jpalvarezl (approved):
I would be ok merging this, but the default value for native coin in the gateway seems a bit off my jurisdiction in terms of expected behaviour :sweat_smile:

src/services/balances.rs
rmeissner: I am very much with @jpalvarezl here. Either the `logo_url` is optional on both services or it is required on both services. We should not have any defaults (neither for the `name`, `symbol` or `decimals`) as everything should be provided by the config service.

rmeissner: Also after looking at the code ... this `Err` case can never happen. We should unwrap or bail here.

fmrsabino: Default removed 63085c8

rmeissner (approved):

PR #467, opened 2021-07-07
Fix docker deploys

jpalvarezl:

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -153,7 +153,7 @@ jobs:
         if: github.ref == 'refs/heads/interstellar'
         run: |
           bash scripts/deploy_docker.sh interstellar |
-          curl -H "Content-Type: application/json" -X POST -d '{"push_data": {"tag": "staging" }}' ${{ AUTODEPLOY_URL }}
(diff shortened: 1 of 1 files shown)


.github/workflows/rust.yml
```diff
@@ -151,7 +151,9 @@ jobs:
 
       - name: Deploy interstellar
         if: github.ref == 'refs/heads/interstellar'
-        run: bash scripts/deploy_docker.sh interstellar
+        run: |
+          bash scripts/deploy_docker.sh interstellar
```
rmeissner, line 155: new line indicator missing at the end

.github/workflows/rust.yml
```diff
@@ -151,7 +151,9 @@ jobs:
...
+          curl -H "Content-Type: application/json" -X POST -d '{"push_data": {"tag": "staging" }}' ${{ AUTODEPLOY_URL }}
```
rmeissner, line 156: is `staging` the correct tag here?

rmeissner: shouldn't this be `interstellar`?

jpalvarezl: wasn't this supposed to be `staging` for the deployment prefix?

rmeissner (approved):

PR #468, opened 2021-07-07
Register and Unregister push notification endpoints

### LINKED ISSUE #454: Register device for push notifications

### LINKED ISSUE #455: Unregister device for push notifications

jpalvarezl:
Closes #454 
Closes #455 

Changes:
- Added endpoints for registering and unregistering push notification 
- Introduced `structs` for backend request
- Introduced `struct` for gw request

### DIFF
--- a/src/routes/notifications.rs
+++ b/src/routes/notifications.rs
@@ -1,7 +1,7 @@
 use crate::utils::context::Context;
 
 use crate::models::service::notifications::NotificationRegistrationRequest;
-use crate::services::notifications::post_registration;
+use crate::services::notifications::{delete_registration, post_registration};
 use crate::utils::errors::ApiResult;
 use rocket::response::content;
 use rocket::serde::json::Error;
@@ -37,4 +37,14 @@ pub async fn post_notification_registration<'e>(
     registration_request: Result<Json<NotificationRegistrationRequest>, Error<'e>>,
 ) -> ApiResult<content::Json<String>> {
     post_registration(context, registration_request?.0).await
+}
+
+#[delete("/v1/chains/<chain_id>/notificaitons/devices/<uuid>/safes/<safe_address>")]
+pub async fn delete_notification_registration(
+    context: Context<'_>,
+    chain_id: String,
+    uuid: String,
+    safe_address: String,
+) -> ApiResult<()> {
+    delete_registration(context, chain_id, uuid, safe_address).await
 }
--- a/src/routes/notifications.rs
+++ b/src/routes/notifications.rs
@@ -8,20 +8,18 @@ use rocket::serde::json::Error;
 use rocket::serde::json::Json;
 
 /**
- * `/v1/chains/<chain_id>/transactions/<safe_tx_hash>/confirmations` <br />
- * Returns [TransactionDetails](crate::models::service::transactions::details::TransactionDetails)
+ * `/v1/register/notifications` <br />
+ * Returns `()`
  *
- * # Transaction Confirmation
+ * # Register notifications
  *
- * This endpoint provides a way for submitting confirmations for clients making use of the `safe_tx_hash` as part of the path, and the very same `safe_tx_hash` signed by an owner corresponding to the safe from which the transaction is being sent.
+ * This endpoint provides a way for registering devices for push notifications.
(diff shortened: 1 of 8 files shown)


src/services/notifications.rs
```diff
@@ -33,28 +33,36 @@ pub async fn post_registration(
...
         let backend_request =
             build_backend_request(&registration_request.device_data, safe_registration);
-        client
+        let response = client
             .post(url.to_string())
             .json(&backend_request)
             .send()
             .await?;
+
+        if !response.status().is_success() {
```
rmeissner, line 47: This will stop the registration in case on an error, right? I thought we ignore.errors and try to register on all services that are relevant.

Also is it possible to do these registrations in parallel with async await?

jpalvarezl: we can do that, that was the old behaviour. 

Regarding your second question, I could add a stream implementation here, I guess. I will have a look

src/routes/notifications.rs
```diff
@@ -0,0 +1,65 @@
...
+ * No query parameters available for this endpoint.
+ */
+#[post(
+    "/v1/register/notifications",
+    format = "application/json",
+    data = "<registration_request>"
+)]
+pub async fn post_notification_registration<'e>(
+    context: Context<'_>,
+    registration_request: Result<Json<NotificationRegistrationRequest>, Error<'e>>,
```
rmeissner, line 34: Why do we require a `Result` here? (e.g. in the hook we directly have the `Json`, while in the confirmation screen we also use a result) Is this for better error generation?

src/services/notifications.rs
```diff
@@ -0,0 +1,60 @@
...
+
+    let info_provider = DefaultInfoProvider::new(&chain_id, &context);
+    let url = core_uri!(
+        info_provider,
+        "/v1/notifications/devices/{}/safes/{}/",
+        uuid,
+        safe_address
+    )?;
+
+    client.delete(url).send().await?;
```
rmeissner, line 25: Should we check the status here, to indicate if an unsubscribe was successful, or do the clients ignore the status code of this request? (e.g. would the client block the removal of a Safe if it cannot unsubscribe)

rmeissner (approved):
Small comments, none should be blocking.

We don't have an easy way to generate any meaningful tests here, right?

src/routes/notifications.rs
jpalvarezl: Yeah, better errors from rocket :+1:

src/services/notifications.rs
jpalvarezl: I would merge the issue and consult with the clients if the errors are meaningful to them.

jpalvarezl:
Unfortunately, there is not much to test here, the mapping function just re-arranges values and the rest are raw http requests :(

PR #474, opened 2021-07-08
Expose limit query parameter for /chains

### LINKED ISSUE #473: Expose limit query parameter for the /chains endpoint
Clients should be able to adjust the number of chains returned from the paginated `/chains` endpoint with the `limit` query parameter

fmrsabino:
Closes #473

- The `limit` query parameter is now exposed to the clients so the default number of returned chains on a given page can be adjusted


jpalvarezl (approved):
LGTM!

PR #475, opened 2021-07-08
Add Dependabot

### LINKED ISSUE #456: Introduce `dependabot` to the repo
Tool for automation of dependency update. The `config` and `transaction` service use it. 

https://dependabot.com/

fmrsabino:
Closes #456

- Add Dependabot 
- Dependabot will run on a weekly basis, on Mondays
- Assign @fmrsabino, @jpalvarezl and @rmeissner as reviewers


jpalvarezl:
I would prefer holding off this PR until we are finished with the production deployment, as I am guessing a bunch of PRs would be automatically opened and then we would have to see if the dependency updates affect us somehow. Would you agree?

fmrsabino:
> I would prefer holding off this PR until we are finished with the production deployment of `interstellar`, as I am guessing a bunch of PRs would be automatically opened and then we would have to see if the dependency updates affect us somehow. Would you agree?

Sounds good to me! 🙂

jpalvarezl (approved):

PR #477, opened 2021-07-09
Error forwading for notification endpoints

### LINKED ISSUE #476: Implement error forwarding for notification endpoints
Un-registering:
 - No errors, nothing but a 200
 - forwards the error fully

Registering:
 - No errors, nothing but a 200
 - Partial errors, provide a (service side?) return list of failed ids. 
 - Error on the request to the gw, errors as usual

jpalvarezl:
Closes #476 

Current state:

The CGW responds with a status code `500` and the following body:

```json
{
    "code": 1337,
    "message": "Push notification registration failed for chain ids: 1, 4"
}
```
Where the `chain_id`s that failed are listed at the end of the `message` field

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -27,10 +27,10 @@ macro_rules! create_id {
 
 macro_rules! bail {
     ($msg:literal $(,)?) => {
-        return Err($crate::api_error!($msg))
+        return Err($crate::api_error!($msg));
     };
     ($fmt:expr, $($arg:tt)*) => {
-        return Err($crate::api_error!($fmt, $($arg)*))
+        return Err($crate::api_error!($fmt, $($arg)*));
     };
 }
 
@@ -44,7 +44,6 @@ macro_rules! api_error {
     };
     ($fmt:expr, $($arg:tt)*) => {
         $crate::utils::errors::ApiError::new_from_message(format!($fmt, $($arg)*))
-        $crate::private::new_adhoc()
     };
 }
 
--- a/src/services/notifications.rs
+++ b/src/services/notifications.rs
@@ -95,13 +95,13 @@ fn build_backend_request(
 }
 
 async fn forward_error(response: reqwest::Response) -> ApiResult<()> {
-    if !response.status().is_success() {
+    if response.status().is_success() {
+        Ok(())
+    } else {
         Err(ApiError::from_http_response(
             response,
             String::from("Unexpected notification registration error"),
         )
         .await)
-    } else {
-        Ok(())
     }
 }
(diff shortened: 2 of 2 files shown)


src/macros.rs
```diff
@@ -27,10 +27,10 @@ macro_rules! create_id {
 
 macro_rules! bail {
     ($msg:literal $(,)?) => {
-        return Err($crate::api_error!($msg))
```
fmrsabino, line 30: For the Rust newbie: what does this accomplish? 😄

fmrsabino (approved):

src/macros.rs
jpalvarezl: just because we have `return` somehow in the past it was working though not sure why :/

src/services/notifications.rs
```diff
@@ -41,35 +40,68 @@ pub async fn post_registration(
...
+            };
+        }
+        output
+    };
 
-    Ok(())
+    if chain_id_errors.is_empty() {
+        Ok(())
+    } else {
+        bail!(
```
rmeissner, line 79: Could we document the expected response in the PR description and let the client teams give a :+1: ?

src/services/notifications.rs
```diff
@@ -41,35 +40,68 @@ pub async fn post_registration(
...
+            } else {
+                // reqwest failures (unreachable urls, do we need to handle?)
+                output.push(chain_id);
+            };
+        }
+        output
+    };
 
-    Ok(())
+    if chain_id_errors.is_empty() {
```
rmeissner, line 76: For this logic a `boolean` would be enough, no need to create a vector

src/services/notifications.rs
```diff
@@ -41,35 +40,68 @@ pub async fn post_registration(
...
         notification_device_data: device_data.clone(),
-        safes: safe_registration.safes,
-        signatures: safe_registration.signatures,
+        safes: safe_registration.safes.to_owned(),
+        signatures: safe_registration.signatures.to_owned(),
+    }
+}
+
+async fn forward_error(response: reqwest::Response) -> ApiResult<()> {
+    if !response.status().is_success() {
```
rmeissner, line 98: Would it be nicer to switch this if around to avoid the "not"?

src/services/notifications.rs
jpalvarezl: We basically use the vector anyway afterwards, as it contains the `chain_id`s of the failed requests that the clients wanted to keep track of.

rmeissner (approved):

PR #478, opened 2021-07-09
Renamed anything `Ether` to `NativeCoin`

### LINKED ISSUE #470: Replace "ETHER" TOKEN_TYPE with new "NATIVE_TOKEN" type
- Now mobile clients get native token information (logo, name, symbol, decimals) from the `/v1/chains` endpoint.
- In `balances` endpoint this info is used when the balance item has `ETHER` type.
- This client-based logic is done on both Android and iOS
- `ETHER` is not relevant on other chains

This is not blocking mobile release

- Replace the `ETHER` token type with a new `NATIVE_TOKEN` type in the [`TokenInfo`](https://github.com/gnosis/safe-client-gateway/blob/e1dc52675e755dd4389174f68a1d7dee04d8469a/src/providers/info.rs#L80)
- Required fields for that type are `name`, `symbol`, `decimals`, and `logoUri`

- Making no changes on the gateway and handling native token/ether type on the client - this is a working solution, but it would be more consistent to stop using 'ether' when the meaning is 'native token'

jpalvarezl:
Closes #470 

Changes:
 - renamed `TransferInfo::Ether`-> `TransferInfo::NativeCoin`
 - renamed `EtherTransfer` -> `NativeCoinTransfer`
 - renamed `TokenType::Ether` -> `TokenType::NativeCoin` 
 - Fixed issue with `logo_uri` for native coins

### DIFF
--- a/src/models/converters/balances.rs
+++ b/src/models/converters/balances.rs
@@ -12,6 +12,12 @@ impl BalanceDto {
             .as_ref()
             .map(|_| TokenType::Erc20)
             .unwrap_or(TokenType::NativeToken);
+
+        let logo_uri = if token_type == TokenType::NativeToken {
+            Some(native_coin.logo_url.to_string())
+        } else {
+            self.token.as_ref().map(|it| it.logo_uri.to_string())
+        };
         Balance {
             token_info: TokenInfo {
                 token_type,
@@ -34,7 +40,7 @@ impl BalanceDto {
                     .as_ref()
                     .map(|it| it.name.to_string())
                     .unwrap_or(native_coin.name.to_string()),
-                logo_uri: self.token.as_ref().map(|it| it.logo_uri.to_string()),
+                logo_uri,
             },
             balance: self.balance.to_owned(),
             fiat_balance: fiat_balance.to_string(),
--- a/src/models/converters/balances.rs
+++ b/src/models/converters/balances.rs
@@ -11,7 +11,7 @@ impl BalanceDto {
             .token_address
             .as_ref()
             .map(|_| TokenType::Erc20)
-            .unwrap_or(TokenType::Ether);
+            .unwrap_or(TokenType::NativeToken);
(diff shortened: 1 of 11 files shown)


src/models/converters/balances.rs
```diff
@@ -11,7 +11,13 @@ impl BalanceDto {
...
             .as_ref()
             .map(|_| TokenType::Erc20)
-            .unwrap_or(TokenType::Ether);
+            .unwrap_or(TokenType::NativeToken);
+
+        let logo_uri = if token_type == TokenType::NativeToken {
+            Some(native_coin.logo_url.to_string())
+        } else {
+            self.token.as_ref().map(|it| it.logo_uri.to_string())
+        };
```
fmrsabino, line 20: What's your opinion on removing the `native_coin` entirely from this function?

This is related to the balance of a specific token right? Wouldn't providing the `native_coin` as a fallback be incorrect? 🤔

fmrsabino (commented):
Couple suggestions:

- `safe_client_gateway::models::backend::webhooks::IncomingEther` to `IncomingNativeCoin`
- `safe_client_gateway::models::backend::transfers::Transfer` we have `Ether` and `EtherTransfer` but we are already deserialising as a different name (from the tx service) so maybe we can rename this to native coin here too?

src/models/converters/balances.rs
rmeissner: When the core service don't return an address we assume that it is the native token. And as the clients would like to have the `logo_uri` required for the native coind this is the correct way to provide it

rmeissner (approved):

jpalvarezl:
I only modified the structs and enums exposed to the clients. 

We can adjust the two structs you mentioned, but the change has to be done first in the core services. As this would break our deserialisation. We'd also have to invalidate caches if we do this after the production deployment :thinking:

PR #479, opened 2021-07-12
Adjusting logs for multiple chains

### LINKED ISSUE #450: Add chainId/rawPath information to logging regex
With the unified interface it will be difficult to parse the logs.

Add `chainId` and path without url resolving the variables `raw_path` (if possible) to the logs: https://github.com/gnosis/safe-client-gateway/blob/301dc3265af56f91546c90f795ee16547a8de4da/src/monitoring/performance.rs#L31

Proposed format:

`log::info!("MT::{}::{}::{}::{}::{}::{}::{}", method, raw_path, delta, status_code, path_data, chain_id)`

jpalvarezl:
Closes #450 

Note: 

I can also include the query parameters for `request_path` if you like

We'd need to just add:

```rust
let request_params = request.uri().query();
log::error!("query params: {:#?}", request_params.as_ref().map(|it| it.to_string()));
```

### DIFF
--- a/src/monitoring/performance.rs
+++ b/src/monitoring/performance.rs
@@ -1,6 +1,11 @@
 use chrono::Utc;
+use lazy_static::lazy_static;
 use rocket::fairing::{Fairing, Info, Kind};
 use rocket::{Data, Request, Response};
+
+lazy_static! {
+    static ref CHAIN_ID_PATH_PATTERN: Regex = Regex::new(r"/v1/chains/d{+}/+").unwrap();
+}
 
 pub struct PerformanceMonitor();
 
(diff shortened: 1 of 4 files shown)


src/monitoring/performance.rs
```diff
@@ -28,6 +33,23 @@ impl Fairing for PerformanceMonitor {
...
+            status_code,
+            request.uri().to_string(), // full path with query params
+            chain_id
+        );
+    }
+}
+
+pub(super) fn extract_chain_id(path: &Path) -> String {
+    let chain_id = path.segments().get(2);
+    if path.to_string().starts_with("/v1/chains/") && chain_id.is_some() {
```
rmeissner, line 50: Why don't we check `segments().get(1) == "chains"`?

src/monitoring/tests/path_patterns.rs
```diff
@@ -0,0 +1,25 @@
+use crate::monitoring::performance::extract_chain_id;
+
+#[test]
+fn chain_dependent_endpoint() {
```
rmeissner, line 4: Can we add a test with a `uri` that has no `v1` and another that has not `chains`?

rmeissner (approved):

PR #480, opened 2021-07-12
Added missing operator

jpalvarezl:
Changes:

- ~~Added `&&` to multi-command multi-line `run` statement~~
- Did it like in the docs


fmrsabino (approved):
Was about to comment on it 😄 

But here are the docs: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstepsrun

PR #481, opened 2021-07-13
Added `config_uri!` macro

### LINKED ISSUE #441: Homogenise usage of either `format!` or `reqwest::Uri`
## Context

Currently `core_uri!` uses internally `fomart!` whereas request to the config service use `reqwest::Uri` to build the URLs

## Feature

Evaluate which is the quickest option, to either stick to `format!` or use `reqwest::Uri` in the `core_uri!`. Eliminate all usages of `format!` or `reqwest::Uri` for URLs depending on the used approach.

jpalvarezl:
Closes #441 

Change:
- For sake of simplicity and control, went with the `format!` approach
- To hide implementation, added `config_uri!` macro, to match `core_uri!`

### DIFF
--- a/src/services/chains.rs
+++ b/src/services/chains.rs
@@ -1,27 +1,18 @@
 use crate::cache::cache_operations::RequestCached;
-use crate::config::{
-    base_config_service_url, chain_info_cache_duration, chain_info_request_timeout,
-};
+use crate::config::{chain_info_cache_duration, chain_info_request_timeout};
 use crate::models::chains::ChainInfo;
 use crate::models::commons::Page;
 use crate::utils::context::Context;
 use crate::utils::errors::ApiResult;
-use reqwest::Url;
-use std::collections::HashMap;
 
 pub async fn get_chains_paginated(
     context: &Context<'_>,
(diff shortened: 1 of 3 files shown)


src/services/chains.rs
```diff
@@ -1,27 +1,18 @@
...
-        .expect("Bad base config service url");
-    url.path_segments_mut()
-        .expect("Cannot add chain_id to path")
-        .extend(["v1", "chains"]);
+    let url = config_uri!(
+        "/v1/chains/?limit={}",
+        limit.as_ref().unwrap_or(&"".to_string())
+    );
 
     let body = RequestCached::new(url.to_string())
```
fmrsabino, line 17: This `to_string()` is no longer required right?

```suggestion
    let body = RequestCached::new(url)
```

fmrsabino (approved):

src/services/chains.rs
jpalvarezl: Totally right! Thanks :raised_hands: !

fmrsabino (approved):

PR #483, opened 2021-07-14
Split backend/service `ChainInfo` struct

### LINKED ISSUE #451: Use different model for the Chain exposed to the clients
Currently `ChainInfo` is being used for both parsing the payload from the safe-config service and to serialize the Json payload that will get exposed to the clients.

This might make it more difficult to manage when the payloads are expected to be different (between what the clients get and what the safe-config exposes)

jpalvarezl:
Closes #451 

Changes:
- Split up structures with a backend and service model
- `InfoProvider` returns the backend type, if we need it exposed outwards we can use `From`


src/models/converters/chains.rs
```diff
@@ -0,0 +1,27 @@
...
+                logo_url: chain_info.native_currency.logo_url,
+            },
+            theme: ServiceTheme {
+                text_color: chain_info.theme.text_color,
+                background_color: chain_info.theme.background_color,
+            },
+            ens_registry_address: chain_info.ens_registry_address,
+        }
+    }
+}
```
fmrsabino, line 27: ❓ If one one of the fields changes (eg.: is removed, is added) then a new type would have to be created no? Currently is mapping the type `ChainInfo` into `ChainInfo`.

Would that be done only when there's a mismatch in the payloads?

src/models/converters/balances.rs
```diff
@@ -1,5 +1,5 @@
 use crate::models::backend::balances::Balance as BalanceDto;
-use crate::models::chains::NativeCurrency;
+use crate::models::backend::chains::NativeCurrency;
```
fmrsabino, line 2: nitpick: I would actually consider anything under `crate::models` to be part of this application so I would not put them under `::backend`. As for the client models then it'd make more sense to put them under `crate::models::clients::` wdyt?

This is out of scope for this PR anyways.

src/models/converters/chains.rs
jpalvarezl: Right, so right now they are identical. We allow for divergence now, we can remove fields in the `ServiceChainInfo` that we don't want the clients to receive. Or we can add fields to `BackendChainInfo` without having to adjust the clients each time.

We would need to adjust this mapping function depending on what we want to expose. If we add something new to `BackendChainInfo` we don't really need to change anything if we don't want it returned to the clients.

fmrsabino: If we remove fields from `ServiceChainInfo` in its current form then we would just need to adjust this `From`?

src/models/converters/balances.rs
jpalvarezl: We for sure change that if you like. In a different PR :) . For now the distinction is:

- `crate::model::backend` for classes in the config and tx services
- `crate::model::service` for classes exposed by this service
- `crate::model::commons` for classes which aren't mapped and are forwarded as is to clients

These names are nice because the have the same length in a monospaced font :P

jpalvarezl: Admittedly, `commons` should be refactored into a folder and not just as a flat `.rs` file

src/models/converters/chains.rs
jpalvarezl: Correct, and the clients would have to be notified (when this is deployed of course) since it would represent an API breaking change.

fmrsabino (approved):

PR #486, opened 2021-07-14
Implementation version state exposed in `SafeInfo` endpoint

### LINKED ISSUE #462: Add master copy state to Safe information
## Feature

The client want to know if the mastercopy of a Safe is up to date. This information should be added to Safe info endpoint

## Proposal

The information could be provided in the form of an `enum`.  

**The current states expected are**
- Up to date 
- Not up to date -> if there is a newer version
- Unknown -> if it is an unknown version

**Additional states could be**
 - Not supported (we need to specify the difference with "Unknown" state)
 - Split “not up to date” into 
   * Required
   * Available

## Open questions

- Where do we get this information on a chain basis? (this can diverge from chain to chain)
- What versions do we allow apps to support? (changes in the signing logic could cause the gw to return invalid data e.g. cancellations)

jpalvarezl:
Closes #462


src/models/converters/safes.rs
```diff
@@ -1,10 +1,24 @@
...
+            .await
+            .ok()
+            .as_ref()
+            .map(|chain_info| chain_info.recommended_master_copy_version.to_string());
+        let implementation_version_state = self
+            .version
+            .as_ref()
+            .map_or(ImplementationVersionState::Unknown, |it| {
+                calculate_version_state(it, &min_chain_version)
+            });
```
jpalvarezl, line 21: We return `Unknown` for failure on fetching details. Shouldn't happen, but also, we shouldn't fail the rest of the response in the eventual case.

src/models/converters/safes.rs
```diff
@@ -29,6 +43,29 @@ impl SafeInfo {
...
+        if sem_ver_min.is_err() || sem_ver_safe.is_err() {
+            return ImplementationVersionState::Unknown;
+        }
+
+        return match sem_ver_safe.unwrap().cmp(&sem_ver_min.unwrap()) {
+            Ordering::Less => ImplementationVersionState::Outdated,
+            Ordering::Equal => ImplementationVersionState::UpToDate,
+            Ordering::Greater => ImplementationVersionState::UpToDate,
+        };
+    }
```
fmrsabino, line 68: Rust question: is there a way to have the `None` matching first and then dealing with the case of having a value (this would reduce branching in the function).

```rust
pub(super) fn calculate_version_state(
    safe_version: &str,
    min_chain_version: &Option<String>,
) -> ImplementationVersionState {
    if min_chain_version.is_none() {
        return ImplementationVersionState::Unknown
    }
    // from here on `min_chain_version` is `Some(T)`
}

Comment (jpalvarezl) [reply src/models/converters/safes.rs:68]: Yeah, you can absolutely do that. But in my implementation I am taking advantage of 2 features of rust, de-structuring and name shadowing (well the later, not sure if a "feature" is a bit of a strong word).

By doing 
```rust
let value = Some(1); // optional value

if let Some(value) = value {
   assert_eq!(value, 1);
}
```

I am reusing the variable name, while at the same time obtaining access to the value within the `Option<T>` if it is a some. the `if let` block is skipped if `value` turns out to be a `None`. You can also have `else` for `if let`s . If you want I can enclose the outer path in an `else` if it brings clarity to this.

fmrsabino: Ok what I was thinking can be accomplished with:

```rust
pub(super) fn calculate_version_state(
    safe_version: &str,
    min_chain_version: &Option<String>,
) -> ImplementationVersionState {
    // Deal with the exceptional case first
    let min_chain_version = match min_chain_version {
        Some(min_chain_version) => min_chain_version,
        None => return ImplementationVersionState::Unknown,
    };

    let sem_ver_safe = Version::parse(safe_version);
    let sem_ver_min = Version::parse(min_chain_version);

    if sem_ver_min.is_err() || sem_ver_safe.is_err() {
        return ImplementationVersionState::Unknown;
    }

    return match sem_ver_safe.unwrap().cmp(&sem_ver_min.unwrap()) {
        Ordering::Less => ImplementationVersionState::Outdated,
        Ordering::Equal => ImplementationVersionState::UpToDate,
        Ordering::Greater => ImplementationVersionState::UpToDate,
    };
}
```

Honestly in terms of readability I have no preference!

fmrsabino (approved):

src/models/converters/safes.rs
```diff
@@ -1,10 +1,24 @@
...
+use semver::Version;
+use std::cmp::Ordering;
 
 // We need to add Sync as trait bound as info_provider moves across threads
 impl SafeInfo {
     pub async fn to_safe_info_ex(&self, info_provider: &(impl InfoProvider + Sync)) -> SafeInfoEx {
+        let min_chain_version = info_provider
+            .chain_info()
+            .await
+            .ok()
```
rmeissner, line 13: Why do we not force unwrap this? The general assumption should be: if we cannot fetch the chain info we should not proceed, as we do not know if it is a valid chain.

jpalvarezl: I will use `expect` in the following PR so we get some logged message in the unlikely case this would fail.

PR #488, opened 2021-07-15
Closes #484: Use AddressEx everywhere 

### LINKED ISSUE #484: Use `AddressEx` everywhere possible
## Feature

Currently, due to potentially causing API breaking changes, we introduced `AddressInfo` wherever we wanted a `KnownAddress` additional contract (or token) info on a Ethereum address.

We would like to use 
```rust
pub struct AddressEx {
    pub value: String,
    pub name: Option<String>,
    pub logo_url: Option<String>,
}
```

Wherever we are using a combination of `address: String` and `address_info: AddressInfo`

## Affected locations with existing AddressInfo

- `SettingsChange` transactions:
  - Every variant with a field is affected 
- `Custom` transactions :
  -  `to` and `to_info` 
- `Creation` transaction: 
  - `creator` and `creator_info`
  -  `implementation` and `implementation_info`
  -  `factor` and `factory_info`
- `Transfer` transactions: 
  - `sender` and `sender_info`
  - `recipient` and `recipient_info`
- In `TransactionDetails::TransactionData`
  - `to` has no `AddressInfo` field, could be expanded
  - `address_info_index` is a `HashMap<String, AddressInfo>`, can be simplified into a `Vec<AddressEx>`  

## New locations with AddressInfo

There are fields for which we don't provide `AddressInfo` yet, and we may need to expand them in the future
 
- In `MultisigExecutionDetails`
  - `executor` field 
  - `signers` and `rejectors`
- In `ModuleExecutionDetails`
  - `address` of the module

rmeissner:
Closes #484


src/models/converters/data_decoded.rs
```diff
@@ -81,12 +74,12 @@ impl DataDecoded {
...
         &self,
         info_provider: &impl InfoProvider,
-    ) -> Option<HashMap<String, AddressInfo>> {
+    ) -> Option<HashMap<String, AddressEx>> {
         if !feature_flag_nested_decoding() {
             return None;
         }
 
-        let mut index = HashMap::new();
+        let mut index: HashMap<String, AddressEx> = HashMap::new();
```
jpalvarezl, line 82: I think we can change this to a `Vec<AddressEx>`

jpalvarezl: I mean the field `address_info_index`

src/providers/info.rs
```diff
@@ -76,6 +77,7 @@ struct Manifest {
 pub struct TokenInfo {
     #[serde(rename = "type")]
     pub token_type: TokenType,
+    // No need to map to AddressEx as the information are present in this struct
```
jpalvarezl, line 80: This would also break deserialisation as we rely on the same structure for the backend returned types.

src/providers/info.rs
```diff
@@ -90,8 +92,8 @@ pub trait InfoProvider {
     async fn safe_info(&self, safe: &str) -> ApiResult<SafeInfo>;
     async fn token_info(&self, token: &str) -> ApiResult<TokenInfo>;
     async fn safe_app_info(&self, url: &str) -> ApiResult<SafeAppInfo>;
-    async fn contract_info(&self, address: &str) -> ApiResult<AddressInfo>;
-    async fn full_address_info_search(&self, address: &str) -> ApiResult<AddressInfo>;
+    async fn add_address_info_from_any_source(&self, address: &str) -> ApiResult<AddressEx>;
+    async fn add_address_info_from_contract_info(&self, address: &str) -> ApiResult<AddressEx>;
```
jpalvarezl, line 96: the prefix `add_` throws me off a bit. Maybe:
- `address_ex_from_any_source`
- `address_ex_from_contracts`

src/providers/ext.rs
```diff
@@ -14,7 +13,11 @@ pub trait InfoProviderExt: InfoProvider {
         self.token_info(address).await.ok()
     }
 
-    async fn addresses_to_address_ex(
+    async fn add_address_info_from_contract_info_or_empty(&self, address: &String) -> AddressEx {
```
jpalvarezl, line 16: could we rename this to `to_address_ex_contracts` ? I suppose we don't have `to_address_ex_any_source` in the `ext`

src/macros.rs
```diff
@@ -86,7 +86,7 @@ macro_rules! core_uri {
 #[macro_export]
 macro_rules! config_uri {
     ($path:expr) => {{
-        format!("{}/{}", $crate::config::base_config_service_url(), $path)
```
fmrsabino, line 89: Why this change?

src/models/converters/data_decoded.rs
rmeissner: This would make the lookup more complicated, also the guarantee that you have no duplicates would not be given anymore

src/providers/ext.rs
rmeissner: The point hwre is that it doesn't baul, and instead returns an addressex where only the address is populated. I would like to make this clear in zhe method name, as this was throwing me off in the original naming

src/macros.rs
rmeissner: Because we define the slash when using the macro, right now we have two slashes in the url

src/providers/info.rs
rmeissner: I would add the `to_` in fron of it

src/macros.rs
fmrsabino: I'd create a different PR for this as it seems outside of the scope for the `AddressEx` changes

rmeissner: I created https://github.com/gnosis/safe-client-gateway/issues/489

src/providers/ext.rs
rmeissner: See above.

src/macros.rs
jpalvarezl: Hahah I would not have minded the fix here xD as this triggers unnecessary forward requests

src/models/converters/safes.rs
```diff
@@ -1,4 +1,5 @@
-use crate::models::service::safes::{AddressEx, ImplementationVersionState, SafeInfoEx};
+use crate::models::service::addresses::AddressEx;
```
jpalvarezl, line 1: This used to be next so `safes`, but it's cool that we create a module of its own

src/models/converters/tests/get_address_ex_from_any_source.rs
```diff
@@ -0,0 +1,60 @@
...
+    let safe = "0x1234";
+
+    let mut mock_info_provider = MockInfoProvider::new();
+    mock_info_provider
+        .expect_address_ex_from_contracts()
+        .times(0);
+
+    let actual = get_address_ex_from_any_source(safe, address, &mut mock_info_provider).await;
+    assert_eq!(AddressEx::address_only(address), actual);
+}
```
jpalvarezl, line 60: ah ok this is just a rename from `src/models/converters/tests/get_address_info.rs`. Not terribly important but I find this https://git-scm.com/docs/git-mv preferable, if commit history matters more (in our case we squash anyhow)

src/models/converters/tests/transfer_erc20.rs
```diff
@@ -60,23 +58,23 @@ async fn erc20_transfer_dto_to_incoming_transfer_transaction_with_address_info()
...
+        .expect_address_ex_from_any_source()
         .times(1)
-        .return_once(move |_| {
-            Ok(AddressInfo {
-                name: "".to_string(),
-                logo_uri: None,
+        .return_once(move |address| {
+            Ok(AddressEx {
+                value: address.to_string(),
+                name: Some("".to_string()),
```
jpalvarezl, line 66: Judging by this change it seems like `name` used to not be nullable. Is that really the case? It doesn't matter too much but maybe we are being unnecessarily defensive here.

src/models/converters/transactions/details.rs
```diff
@@ -99,23 +107,30 @@ impl MultisigTransaction {
...
 }
 
 impl ModuleTransaction {
     pub async fn to_transaction_details(
         &self,
-        info_provider: &impl InfoProvider,
+        info_provider: &(impl InfoProvider + Sync),
     ) -> ApiResult<TransactionDetails> {
         let safe_transaction = &self.safe_transaction;
+        let module_info = info_provider
```
jpalvarezl, line 125: Sneaky feature :) good though, glad it was an easy change.

src/models/converters/transactions/mod.rs
```diff
@@ -198,15 +202,20 @@ impl MultisigTransaction {
...
 
-    fn missing_signers(&self, owners: &Vec<String>) -> Vec<String> {
+    fn missing_signers(&self, owners: &Vec<String>) -> Vec<AddressEx> {
         self.confirmations.as_ref().map_or_else(
-            || owners.to_owned(),
+            || {
+                owners
+                    .iter()
+                    .map(|owner| AddressEx::address_only(&owner))
+                    .collect()
```
jpalvarezl, line 211: I wanted to suggest to have a `From` implementation for `Vec<&str>` into `Vec<AddressEx>` for these sort of cases, but doesn't seem easily possible. Just as a general comment, unrelated to the PR, maybe we should start leveraging more `From` as it seems to be more idiomatic for these sort of situations.

jpalvarezl: Although a disadvantage of `From` is  that we can't pass dependencies in addition to the `struct` instance to the function. And it's also not `async`. For simple use case should be fine though.

src/models/service/addresses.rs
```diff
@@ -0,0 +1,28 @@
+use serde::Serialize;
+
+#[derive(Serialize, Debug, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub struct AddressEx {
+    pub value: String,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub name: Option<String>,
```
jpalvarezl, line 8: Ah ok, so we were exposing this as nullable from the past. This answers the above question :point_up:

src/models/service/transactions/details.rs
```diff
@@ -155,41 +155,43 @@ pub struct MultisigExecutionDetails {
     pub safe_tx_gas: usize,
     pub base_gas: usize,
     pub gas_price: String,
+    // As this is a token we will keep the token information separated
```
jpalvarezl, line 158: Do you want to open an issue for this? We could map this into `TokenInfo`. Another surprise breaking change.

image

src/models/service/transactions/details.rs
```diff
@@ -155,41 +155,43 @@ pub struct MultisigExecutionDetails {
...
     pub hex_data: Option<String>,
     pub data_decoded: Option<DataDecoded>,
-    pub to: String,
+    pub to: AddressEx,
     pub value: Option<String>,
     pub operation: Operation,
     #[serde(skip_serializing_if = "Option::is_none")]
-    pub address_info_index: Option<HashMap<String, AddressInfo>>,
+    // Mapping with info for the addresses in data_decoded
+    pub address_info_index: Option<HashMap<String, AddressEx>>,
```
jpalvarezl, line 196: Like you mention before, as we have it, we avoid duplicates, but we could implement it in a way to avoid duplicates while still service `Vec<AddressEx>` in a separate ticket.

src/providers/info.rs
```diff
@@ -165,20 +169,22 @@ impl<C: Cache> InfoProvider for DefaultInfoProvider<'_, C> {
...
-    async fn full_address_info_search(&self, address: &str) -> ApiResult<AddressInfo> {
+    async fn address_ex_from_any_source(&self, address: &str) -> ApiResult<AddressEx> {
         self.token_info(&address)
-            .map_ok(|it| AddressInfo {
-                name: it.name,
-                logo_uri: it.logo_uri,
+            .map_ok(|it| AddressEx {
+                value: address.to_owned(),
+                name: Some(it.name),
+                logo_url: it.logo_uri,
```
jpalvarezl, line 185: the `uri/url` renaming will be addressed in a different issue, right?

src/models/converters/tests/transfer_erc20.rs
jpalvarezl: This question was answered below :point_down: it was how it was originally served to the clients. And it makes sense as the `name` and `logo_uri` are the fallible parts of this `struct` if not resolved by the `info_provider`

jpalvarezl (approved):
LGTM, just one thing that I would like you to address, by either replying to this or implementing the change:

I see one method in `src/providers/ext.rx` called `address_to_token_info` , it seems to only be used once. Maybe we could inline it? as it appears to simply do:

```rust
optional_address.as_ref()
  .map(async move |it| {info_provider.token_info(it).await.ok())
  .flatten()
```

There could definitely be more uses for it, I would just then rename the method to `optional_address_to_token_info` like we have for `AddressEx` methods.

src/json/results/tx_details_with_origin.json
```diff
@@ -92,25 +96,43 @@
...
+      },
+      {
+        "value": "0xA3DAa0d9Ae02dAA17a664c232aDa1B739eF5ae8D"
+      },
+      {
+        "value": "0xF2CeA96575d6b10f51d9aF3b10e3e4E5738aa6bd"
+      },
+      {
+        "value": "0x65F8236309e5A99Ff0d129d04E486EBCE20DC7B0"
+      }
```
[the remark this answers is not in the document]
rmeissner: Done

src/models/converters/data_decoded.rs
```diff
@@ -12,61 +13,61 @@ use std::collections::HashMap;
 impl DataDecoded {
     pub(super) async fn to_settings_info(
         &self,
-        info_provider: &impl InfoProvider,
+        info_provider: &(impl InfoProvider + Sync),
```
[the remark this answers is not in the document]
rmeissner: yes, because we now request address information multiple times ... we should probably refactors this so that `InfoProvider` by default is also `Sync`

src/models/converters/safes.rs
rmeissner: I moved it out of Safes as it was used in a lot of places unrelated to the Safe model

src/models/converters/tests/get_address_ex_from_any_source.rs
rmeissner: If it would be a pure move this would automatically be detected from github. As the content of the file also changed (all method got renamed) it was not a simple move

src/models/service/addresses.rs
rmeissner: We had the `AddressInfo` (what is returned from the backend) and there it is an empty string instead of null. But null makes more sense here

src/models/service/transactions/details.rs
rmeissner: No, this was more a comment to explain what is going on. Some of our code is not always self explanatory :P

src/models/service/transactions/details.rs
rmeissner: But returning a dictionary to the clients will give the clients the guarantee that each address has one AddressEx. And they can easily look it up without having to build the map themselves beforehand. I don'T think we win much by avoiding 1 duplicated address. We do not have to optimize the response on every char.

src/providers/info.rs
rmeissner: Yes

rmeissner:
I didn't look at `address_to_token_info` as this was unrelated to this PR/ issue

PR #491, opened 2021-07-15
Closes #487: Add execution info for module transactions

### LINKED ISSUE #487: `TransactionSummary::ExecutionInfo` add new variant for `ModuleTxs`
## Feature

`TransactionSummary::ExecutionInfo` only has data relevant to `MultisigTransactions`. We need to introduce polymorphism and add a variant for `ModuleTransactions` containing information on the module that triggered the transaction.  

## Proposal

Should have the same variants as https://github.com/gnosis/safe-client-gateway/blob/main/src/models/service/transactions/details.rs#L145

rmeissner:
Closes #487: Add execution info for module transactions

### DIFF
--- a/src/models/converters/transactions/tests/summary.rs
+++ b/src/models/converters/transactions/tests/summary.rs
@@ -180,7 +180,7 @@ async fn module_transaction_to_custom_summary_and_module_info() {
             Ok(AddressEx {
                 value: address.to_string(),
                 name: Some(format!("{}_name", address)),
-                logo_url: None,
+                logo_uri: None,
             })
         });
     mock_info_provider
@@ -209,7 +209,7 @@ async fn module_transaction_to_custom_summary_and_module_info() {
             address: AddressEx {
                 value: "0xfa559f0932b7B60d90B4af0b8813d4088465096b".to_string(),
                 name: Some("0xfa559f0932b7B60d90B4af0b8813d4088465096b_name".to_string()),
-                logo_url: None,
+                logo_uri: None,
             },
         })),
         safe_app_info: None,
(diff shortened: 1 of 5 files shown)


src/models/service/transactions/summary.rs
```diff
@@ -138,16 +138,29 @@ pub struct TransactionSummary {
     pub safe_app_info: Option<SafeAppInfo>,
 }
 
+#[derive(Serialize, Debug, PartialEq)]
+#[serde(tag = "type", rename_all = "SCREAMING_SNAKE_CASE")]
+pub enum ExecutionInfo {
+    Multisig(MultisigExecutionInfo),
+    Module(ModuleExecutionInfo),
+}
```
jpalvarezl, line 146: We would need to adjust the sample json in this file to contain ` "type" : "MULTISIG_EXECUTION_INFO"` If you are tight on time, we can open an issue to update the docs after we've finished all the api breaking changes.

jpalvarezl (approved):
Only comment would be that we need to update docs

src/models/converters/transactions/tests/summary.rs
```diff
@@ -152,6 +166,61 @@ async fn module_tx_to_summary_transaction_failed() {
...
+    mock_info_provider.expect_safe_info().times(0);
+    mock_info_provider.expect_token_info().times(0);
+    mock_info_provider
+        .expect_address_ex_from_contracts()
+        .times(1)
+        .return_once(move |address| {
+            Ok(AddressEx {
+                value: address.to_string(),
+                name: Some(format!("{}_name", address)),
+                logo_url: None,
```
jpalvarezl, line 183: This should be adjusted after you rebase. Some how the PR is still mergable :/

jpalvarezl (changes requested):
Just to prevent some dodgy stuff from happening after rebase

src/models/converters/transactions/tests/summary.rs
fmrsabino: Yeah the checks were done on a state where the code that was added didn't have any conflicts (currently it still doesn't have conflicts so we should rebase).

fmrsabino (changes requested):

jpalvarezl (approved):

PR #492, opened 2021-07-16
Rename url field to uri in client payloads

### LINKED ISSUE #490: Use `uri` instead of `url` everywhere
Currently we mix `uri` and `url` for field names. As the core services uses `uri` we should use this everywhere.

Related: https://github.com/gnosis/safe-config-service/pull/160

fmrsabino:
Closes #490 

Given the current changes in the safe-config service https://github.com/gnosis/safe-config-service/pull/160, the `url` fields should be renamed to `uri`.


jpalvarezl (approved):
:+1:

jpalvarezl:
There is still the methods in `src/config/mod.rs` named `base_config_service_url`, `base_exchange_api_url` and `redis_url` as well as their respective env variables: `CONFIG_SERVICE_URL`, `EXCHANGE_API_BASE_URL` and `REDIS_URL`. 

@rmeissner could we simply fork these into a separate issue, if we need to change them too?

jpalvarezl (approved):

jpalvarezl (changes requested):
We are missing `SafeAppInfo::logo_url`

src/providers/info.rs
```diff
@@ -172,7 +172,7 @@ impl<C: Cache> InfoProvider for DefaultInfoProvider<'_, C> {
             Ok(AddressEx {
```
jpalvarezl, line 172: :point_up: in line 65 `SafeAppInfo` has still a `logo_uri`

rmeissner:
Yes, I would not adjust env variable names in this PR

jpalvarezl:
For reference: https://github.com/gnosis/safe-client-gateway/issues/494

jpalvarezl (approved):

PR #496, opened 2021-07-16
Set block_explorer_uri as Optional

fmrsabino:


rmeissner (changes requested):
There are no tests where this is None

jpalvarezl:
We should probably open an associated ticket so the clients can keep track of this change a bit more easily.

fmrsabino:
After discussing this issue internally we decided that this field should be required (so should stay as it is)

PR #501, opened 2021-07-19
Enhancing logic for `Up-to-Date` implementation version

### LINKED ISSUE #462: Add master copy state to Safe information

jpalvarezl:
Closes #462

### DIFF
--- a/src/models/backend/safes.rs
+++ b/src/models/backend/safes.rs
@@ -1,0 +1,11 @@
+use serde::{Deserialize, Serialize};
+
+#[derive(Serialize, Deserialize, Debug)]
+#[serde(rename_all = "camelCase")]
+pub struct MasterCopy {
+    pub address: String,
+    pub version: String,
+    pub deployer: String,
+    pub deployed_block_number: u64,
+    pub last_indexed_block_number: u64,
+}
--- a/src/models/converters/safes.rs
+++ b/src/models/converters/safes.rs
@@ -24,6 +24,7 @@ impl SafeInfo {
                 .map_or(ImplementationVersionState::Unknown, |safe_version| {
                     calculate_version_state(
                         safe_version,
+                        &self.master_copy,
                         &supported_master_copies,
                         min_chain_version,
                     )
@@ -58,6 +59,7 @@ impl SafeInfo {
 
 pub(super) fn calculate_version_state(
     safe_version: &str,
+    safe_implementation_address: &str,
     supported_master_copies: &Vec<MasterCopy>,
     min_chain_version: String,
 ) -> ImplementationVersionState {
--- a/src/models/converters/safes.rs
+++ b/src/models/converters/safes.rs
@@ -71,16 +71,15 @@ pub(super) fn calculate_version_state(
         .map(|it| it.address.as_str())
         .collect::<Vec<&str>>();
 
-    if sem_ver_min.is_err() || sem_ver_safe.is_err() {
+    if sem_ver_min.is_err()
+        || sem_ver_safe.is_err()
+        || !supported_addresses.contains(&safe_implementation_address)
+    {
         return ImplementationVersionState::Unknown;
     }
 
     let sem_ver_safe = sem_ver_safe.unwrap();
     let sem_ver_min = sem_ver_min.unwrap();
-
-    if !supported_addresses.contains(&safe_implementation_address) {
-        return ImplementationVersionState::Unknown;
-    }
 
     match sem_ver_safe.cmp(&sem_ver_min) {
         Ordering::Less => ImplementationVersionState::Outdated,
--- a/src/models/converters/safes.rs
+++ b/src/models/converters/safes.rs
@@ -3,7 +3,6 @@ use crate::models::service::addresses::AddressEx;
 use crate::models::service::safes::{ImplementationVersionState, SafeInfoEx};
 use crate::providers::ext::InfoProviderExt;
 use crate::providers::info::{InfoProvider, SafeInfo};
-use crate::utils::errors::ApiResult;
 use semver::Version;
 use std::cmp::Ordering;
 
@@ -12,20 +11,24 @@ impl SafeInfo {
     pub async fn to_safe_info_ex(
         &self,
         info_provider: &(impl InfoProvider + Sync),
-        // supported_master_copies: Vec<MasterCopy>,
+        supported_master_copies: Vec<MasterCopy>,
     ) -> SafeInfoEx {
         let min_chain_version = info_provider
             .chain_info()
             .await
             .expect("ChainInfo must be available")
             .recommended_master_copy_version;
+        let implementation_version_state =
+            self.version
+                .as_ref()
+                .map_or(ImplementationVersionState::Unknown, |safe_version| {
+                    calculate_version_state(
+                        safe_version,
+                        supported_master_copies,
+                        min_chain_version,
+                    )
+                });
 
-        let implementation_version_state = self
-            .version
-            .as_ref()
-            .map_or(ImplementationVersionState::Unknown, |it| {
-                calculate_version_state(it, min_chain_version)
-            });
         SafeInfoEx {
             address: AddressEx::address_only(&self.address),
             nonce: self.nonce,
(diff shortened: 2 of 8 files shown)


src/services/about.rs
```diff
@@ -16,3 +18,18 @@ pub async fn about(context: &Context<'_>, chain_id: &str) -> ApiResult<About> {
         build_number: build_number(),
     })
 }
+
+pub async fn get_master_copies(
```
jpalvarezl, line 22: Extracted for the sake of re-usability, but this internally created a new instance of the `info_provider` which in the case of the `safe/<address>` endpoint can have unforeseen consenquences.

src/models/backend/safes.rs
```diff
@@ -0,0 +1,11 @@
+use serde::{Deserialize, Serialize};
+
+#[derive(Serialize, Deserialize, Debug)]
+#[serde(rename_all = "camelCase")]
+pub struct MasterCopy {
+    pub address: String,
+    pub version: String,
+    pub deployer: String,
```
rmeissner, line 8: Do we need all of the attributes? 

If we add more attributes, the chances are higher that something fails because of changes to these attributes :P

src/models/converters/safes.rs
```diff
@@ -49,22 +58,32 @@ impl SafeInfo {
...
+        Ordering::Less => {
+            if supported_versions.contains(&sem_ver_safe) {
+                ImplementationVersionState::Outdated
+            } else {
+                ImplementationVersionState::Unknown
+            }
         }
-    } else {
-        ImplementationVersionState::Unknown
+        Ordering::Equal | Ordering::Greater => ImplementationVersionState::UpToDate,
```
rmeissner, line 87: Why do we not check the supported versions in this case? This should be checked always.

I would do the following steps:
1. Load Safe info
2. Load supported master copies
3. Check that Safe master copy is part of supported master copies
4. Load recommended version
5. Map state

src/services/safes.rs
```diff
@@ -18,7 +19,12 @@ pub async fn get_safe_info_ex(
 ) -> ApiResult<SafeState> {
     let info_provider = DefaultInfoProvider::new(chain_id, context);
     let safe_info = info_provider.safe_info(safe_address).await?;
-    let safe_info_ex = safe_info.to_safe_info_ex(&info_provider).await;
+    let supported_master_copies = get_master_copies(&context, chain_id)
```
rmeissner, line 22: We should not do this. As you mentioned above `get_master_copies` has a separate info provider and therefore it will not be reusing the in-memory cache, which here would be helpful.

rmeissner (changes requested):

src/models/backend/safes.rs
jpalvarezl: We don't this is just to match what the backend sends. I would then remove everything but `version` and `address`

src/models/converters/safes.rs
jpalvarezl: I would add this to the earlier return condition in line ~72

src/models/backend/safes.rs
jpalvarezl: This is also used in the `about/master-copies` endpoint. Do you want us to hide fields from the forward to this endpoint?

src/services/safes.rs
jpalvarezl: I think the only option here would be to duplicate some code. As I need access to the `client` in `Context` to issue the request.

src/models/backend/safes.rs
rmeissner: We would have to ask the clients what they required, but they might not require this endpoint at all anymore once this is merged

src/services/safes.rs
rmeissner: the fetching of the master copies could be part of the info provider interface.

jpalvarezl: I thought of it too, but I am starting to lose sight as to what is the purpose of the `info_provider`

jpalvarezl: maybe the `master_copies` do constitute a piece of information that can be considered for "enriching" other responses (namely `SafeInfoEx`) and it wouldn't be the first time we add a method we would return directly from the `info_provider`

jpalvarezl: I would still think of a way to refactor the `info_provider` into more specific subcomponents, as it's starting to look like a monolith

jpalvarezl: I would still think of a way to refactor the `info_provider` into more specific subcomponents, as it's starting to look like a monolith

src/services/safes.rs
```diff
@@ -18,7 +19,13 @@ pub async fn get_safe_info_ex(
 ) -> ApiResult<SafeState> {
     let info_provider = DefaultInfoProvider::new(chain_id, context);
     let safe_info = info_provider.safe_info(safe_address).await?;
-    let safe_info_ex = safe_info.to_safe_info_ex(&info_provider).await;
+    let master_copy_url = core_uri!(info_provider, "/v1/about/master-copies/")?;
+    let supported_master_copies = request_master_copies(&context, master_copy_url)
+        .await
+        .unwrap_or(vec![]);
```
rmeissner, line 25: Why don't we just fail?

src/services/about.rs
```diff
@@ -16,3 +18,25 @@ pub async fn about(context: &Context<'_>, chain_id: &str) -> ApiResult<About> {
...
+    context: &Context<'_>,
+    chain_id: &str,
+) -> ApiResult<Vec<MasterCopy>> {
+    let info_provider = DefaultInfoProvider::new(chain_id, &context);
+    let url = core_uri!(info_provider, "/v1/about/master-copies/")?;
+    request_master_copies(&context, url).await
+}
+
+pub async fn request_master_copies(
```
rmeissner, line 31: Why not make this part of the info provider trait?

rmeissner (commented):
Some small open questions

src/services/about.rs
rmeissner: Just saw the thread above

src/services/safes.rs
jpalvarezl: So that in the case the case we can't fetch the master copies for whatever reason, we still return the rest of the `SafeInfo` object. In the case of an empty vector, we would just return `Unknown` up-to-date state.

src/services/about.rs
jpalvarezl: I thought the discussion was still going on here: https://github.com/gnosis/safe-client-gateway/pull/501#discussion_r672156548

src/services/safes.rs
rmeissner: I still think the current approach feels weird. Having a method that just perform the request. And that method is also in the module of another request handler ... idk ... I would prefer it to be part of the info provider or as you mentioned and extension. but we alreay pull safe info and token info from there, so with our current approach it would fit. And then in a next step these method could be extracted to a separate entity

src/services/safes.rs
rmeissner: hmmm ok, not sure if necessary, but I would comment this.

I think in general we should comment more on when we just fail and when do we use a fallback and why.

src/models/converters/safes.rs
```diff
@@ -49,22 +58,30 @@ impl SafeInfo {
...
-    if let Some(min_chain_version) = min_chain_version {
-        let sem_ver_safe = Version::parse(safe_version);
-        let sem_ver_min = Version::parse(min_chain_version);
+    let sem_ver_safe = Version::parse(safe_version);
+    let sem_ver_min = Version::parse(&min_chain_version);
 
-        if sem_ver_min.is_err() || sem_ver_safe.is_err() {
-            return ImplementationVersionState::Unknown;
-        }
+    let supported_versions = supported_master_copies
```
rmeissner, line 67: We check supported mastercopies by address, not by version

rmeissner (changes requested):

src/models/converters/safes.rs
jpalvarezl: Does this mean that the config service should return the address as well? We only get the stringyfied semver.

jpalvarezl: ah also for `1.3.0` and `1.3.0+L2` we would have different addresses, right? In that case the semver check passes but the addresses wouldn't match

jpalvarezl: I will add unit tests to this and move this PR back to draft

rmeissner (approved):

PR #502, opened 2021-07-19
Fix config_uri macro uri formatting

### LINKED ISSUE #489: Fix ` config_uri` macro
Currently the `config_uri` macro is building the uri using `<base>/<path>` as `path` is commonly provided with a leading `/` the path  has `//` which is unexpected.

Note: we should also add tests for this

fmrsabino:
Closes #489 

- Remove the extra `/` in the `config_uri` macro – the paths already include the initial slash.

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -93,28 +93,3 @@ macro_rules! config_uri {
         config_uri!(full_path)
     }};
 }
-
-#[cfg(test)]
-mod tests {
-    use std::env;
-
-    #[test]
-    fn config_uri_formats_correctly() {
-        env::set_var("CONFIG_SERVICE_URL", "https://config-url-example.com");
-        let expected = "https://config-url-example.com/api/example";
-
-        let actual = config_uri!("/example");
-
-        assert_eq!(expected, actual)
-    }
-
-    #[test]
-    fn config_uri_formats_correctly_with_substitution() {
-        env::set_var("CONFIG_SERVICE_URL", "https://config-url-example.com");
-        let expected = "https://config-url-example.com/api/example/safe";
-
-        let actual = config_uri!("/example/{}", "safe");
-
-        assert_eq!(expected, actual)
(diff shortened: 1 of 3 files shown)


src/macros.rs
```diff
@@ -86,10 +87,30 @@ macro_rules! core_uri {
...
-        format!("{}/{}", $crate::config::base_config_service_url(), $path)
+        format!("{}{}", $crate::config::base_config_service_url(), $path)
     }};
     ($path:literal, $($arg:tt)*) => {{
         let full_path: String = format!($path, $($arg)*);
         config_uri!(full_path)
     }};
 }
+
+#[test]
```
rmeissner, line 98: We normally put our tests in separate test files

src/macros.rs
```diff
@@ -1,3 +1,4 @@
+use std::env;
```
rmeissner, line 1: This is only required for the test, right?

src/macros.rs
fmrsabino: Ah yes. I was discussing with @jpalvarezl regarding this because the `macros` live in the root `src` dir. I thought about moving to a different folder 🤔 What's your recommendation here?

src/macros.rs
fmrsabino: Correct. I should scope this import.

src/macros.rs
rmeissner: No strong opinion, but it should be consistent in our project

jpalvarezl: We could move them to `/utils/tests/macros.rs`

fmrsabino: Yes I can do that. For me it was just that the macros are on `/src/macros.rs/` and not under `src/utils`

fmrsabino: Moved in 2b21092

jpalvarezl (approved):

PR #506, opened 2021-07-21
Made `TokenInfo` caching network dependent

### LINKED ISSUE #505: Make `TokenInfo` entries in the cache chain dependant
We should add different `hset` ids that contain the chain_id

e.g.
`dip_ti_4` for rinkeby

jpalvarezl:
Closes #505

### DIFF
--- a/src/providers/info.rs
+++ b/src/providers/info.rs
@@ -29,7 +29,7 @@ use std::collections::HashMap;
 use std::future::Future;
 use std::time::Duration;
 
-pub const TOKENS_KEY: &'static str = "dip_ti";
+pub const TOKENS_KEY_BASE: &'static str = "dip_ti";
 lazy_static! {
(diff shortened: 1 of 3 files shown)


src/providers/info.rs
```diff
@@ -285,6 +287,10 @@ impl<C: Cache> DefaultInfoProvider<'_, C> {
         Ok(result)
     }
 
+    fn generate_token_key(&self) -> String {
+        format!("{}_{}", TOKENS_KEY_BASE, self.chain_id)
+    }
```
fmrsabino, line 292: Should this function be extracted out of `self` and take the `chain_id` as parameter?

This can be used also by `InvalidationPattern.to_pattern_string` – if we change the pattern we need to make sure that it's changed in both places.

fmrsabino (approved):

PR #517, opened 2021-07-26
Add Block Explorer URI templates

### LINKED ISSUE #515: Add support for Block Explorer URI templates
Different block explorers might have different ways of addressing an `address` or `tx_hash` or any other _addressable_ entity.

Given the current scope of the block explorers that we currently support, we assume that most of this _addressable entities_ can be set on the path of the url.

Therefore we propose the following change:

```javascript
GET /v1/chains/<chain_id>/

{
  ...
  blockExplorerUriTemplate: {
    "address": <string>, // eg.: https://etherscan.io/address/{{address}}
    "txHash": <string> // eg.: https://etherscan.io/tx/{{txHash}}
  }
}
```

The clients would then replace the agreed pattern with its correct value.

### 1. What would be the pattern?

I thought about having `{{address}}`, or `{{txHash}}` but this needs to be aligned with the client teams.

### 2. What should be initial level of support for the different templates?

I'm under the assumption that the clients use a block explorer for both the `address` and `txHash`. Is there any other entity that can and should be added to the payload? (eg.: block number: `https://etherscan.io/block/{{blockNumber}}`)

@gnosis/safe-android @gnosis/safe-ios @gnosis/safe-web @gnosis/safe-services

fmrsabino:
Closes #515 

- Add support for serialising and deserialising Block Explorer URI templates from the config service API

### DIFF
--- a/src/json/chains/rinkeby.json
+++ b/src/json/chains/rinkeby.json
@@ -6,6 +6,10 @@
     "value": "https://someurl.com/rpc"
   },
   "blockExplorerUri": "https://blockexplorer.com",
+  "blockExplorerUriTemplate": {
+    "address": "https://blockexplorer.com/{{address}}",
+    "txHash": "https://blockexplorer.com/{{txHash}}"
+  },
   "transactionService": "https://safe-transaction.rinkeby.staging.gnosisdev.com",
   "nativeCurrency": {
     "name": "Ether",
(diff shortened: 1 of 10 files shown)


jpalvarezl (approved):
LGTM, I don't remember from the spec if the clients expect `tx_hash` or `txHash`, but the latter is compliant with the API naming conventions 👍

fmrsabino:
> LGTM, I don't remember from the spec if the clients expect `tx_hash` or `txHash`, but the latter is compliant with the API naming conventions 👍

It'd be `txHash` but since this is set in the admin we can change it without breaking the CGW.

PR #530, opened 2021-07-30
Naive rate-limiter

jpalvarezl:
Ecosystem friday project toy branch

Brief description:
`governor` is a rate limiter that is thread safe. I incorporated it using `rocket` states and created a `request guard` that we can use to mark any endpoints for which we want to return `429`s when abussed. Currently the request `url` is used which could be problematic for a safe monitored from different devices. It would be cool to evaluate a different `key` against which we rate limit. Also, for the sake of testing, the current rate limit is of `1 request per second`. We should define the behaviour better if we decide to keep this.

Current state: `1 request per uri per second`

Next steps:
- Caching 429's long?  
- Can we configure `governor` to rate limit longer?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,12 @@
 version = 3
 
 [[package]]
+name = "ahash"
+version = "0.3.8"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e8fd72866655d1904d6b0997d0b07ba561047d070fbe29de039031c641b61217"
+
+[[package]]
 name = "aho-corasick"
 version = "0.7.15"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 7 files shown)


rmeissner:
Why would we want to ratelimit on the client gateway? The purpose of the client gateway is to utilize heavy caching to prevent the need of ratelimiting.

jpalvarezl:
This was in motivation to the outage we had yesterday. It is unclear if rate limiting would help the situation, but I wanted to have something close to ready in case we evaluated that it would be helpful somehow. We wrote a postmortem which we would have to study more closely to draw better conclusions.

rmeissner:
Rate limiting should probably be configured via our ingress. It would interesting why we would do it via a request guard instead of ingress. In our staging instance you can see that we have already configured a maximum of 20 requests per second per ip.

jpalvarezl:
This was deemed not necessary at this point

PR #533, opened 2021-08-02
Remove blockExplorerUri from ChainInfo payload

fmrsabino:
- Removes `blockExplorerUri` from both the config service and from the payload exposed to the clients since it is no longer required (after the introduction of `blockExplorerUriTemplate`)


jpalvarezl (approved):
LGMT

PR #534, opened 2021-08-02
Removed confusing parts of the `.env.sample` and `README` files

### LINKED ISSUE #528: Update README.md and `.env.sample`
Both files are outdated for `v3.0.0` and don't reflect accurately how to setup or where the documentation is located.

jpalvarezl:
Closes #528

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,13 +1,14 @@
 # Safe Client Gateway
 [![Actions Status](https://github.com/gnosis/safe-client-gateway/workflows/safe-client-gateway/badge.svg?branch=main)](https://github.com/gnosis/safe-client-gateway/actions)
 [![Coverage Status](https://coveralls.io/repos/github/gnosis/safe-client-gateway/badge.svg)](https://coveralls.io/github/gnosis/safe-client-gateway)
+
 ## Motivation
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -38,20 +39,7 @@ For configurations specific to this service the `.env` file can be used. See nex
 
 Place a `.env` file in the root of the project containing URL pointing to the environment in which you want the gateway to run.
 
-The contents of the file should be the following (see `.env.sample` for an example):
-
-```
-TRANSACTION_SERVICE_URL=<Transaction service host>
```
fmrsabino, line 44: Is this removal intentional?

jpalvarezl: Yes, we don't need the TX service URL as part of our env anymore.

fmrsabino: Ah ok 👍 

It's still being used for the tests though: https://github.com/gnosis/safe-client-gateway/blob/9446aafb4e322b004d11660247c0ad7a91094f54/scripts/cache_warmer/start.py#L6

jpalvarezl: yeah :/ that's just for warming up the cache, but I suppose it needs to be adapted for multiple nets. I would fork that into a different issue, as that script is rarely used.

PR #545, opened 2021-08-04
Issue template for `Chain support request`

### LINKED ISSUE #542: Add issue template in github for requesting adding a new chain
The new issue template should be specific to adding new `ChainInfo` provide a list of the information that needs to be provided as well as all the configuration steps required to be performed by the backend team. 

This would be an interim solution, unless we see that it works well.

jpalvarezl:
Closes #542

### DIFF
--- a/.github/ISSUE_TEMPLATE/new_chain_request.md
+++ b/.github/ISSUE_TEMPLATE/new_chain_request.md
@@ -20,41 +20,39 @@ We need to collect the metadata for a chain. The following section contains a sa
 
 ```json
 {
-  {
-      "chainId": "1",
-      "chainName": "Mainnet",
-      "rpcUri": {
-        "authentication": "API_KEY_PATH",
-        "value": "https://mainnet.infura.io/v3/"
-      },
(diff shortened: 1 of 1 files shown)


rmeissner:
Shouldn't this rather be on the config service?

rmeissner:
Also currently we collect this information via a google doc and then add it via the admin. Do we somewhere define if this would become obsolete? Who do we expect to open such issues?

jpalvarezl:
I saw it as a checklist for myself, as (probably) I would have to add these values in the config service admin panel and then the webhook to the core services. I would just like to have help gathering data about the chain from whoever requests it and a simple checklist to see that I have not forgotten anything "configuration-wise"

jpalvarezl:
this would also provide a history of chains we support in the form of github issues, which I personally see as an advantage (we can always call the `/chains/<chain_id>` endpoint but still.

jpalvarezl:
Transferred: https://github.com/gnosis/safe-config-service/pull/192

Issue #546, opened 2021-08-05
Balances, Safe info and txs don't load


tschubotz:
**Describe the bug**
Balances, Safe info and txs don't load

**To Reproduce**
- Open any Safe on mainnet prod, e.g. https://gnosis-safe.io/app/#/safes/0xA063Cb7CFd8E57c30c788A0572CBbf2129ae56B6/ 
-> Doesn't load.

**Expected behavior**
Balances and txs show up

**Desktop (please complete the following information):**
 - OS: MacOS
 - Browser Firefox
 - Version ?

**Additional context**
- https://safe-client.mainnet.gnosis.io/v1/safes/0xA063Cb7CFd8E57c30c788A0572CBbf2129ae56B6/ -> 503
- https://safe-transaction.gnosis.io/api/v1/safes/0xA063Cb7CFd8E57c30c788A0572CBbf2129ae56B6/  -> fine

jpalvarezl:
Just some notes to myself:

The unified version works: https://safe-client.gnosis.io/v1/chains/1/safes/0xA063Cb7CFd8E57c30c788A0572CBbf2129ae56B6/

This endpoint also uses the `all-transactions` endpoint for calculating the `txHistoryTag`

PR #547, opened 2021-08-05
Added a 3 second timeout to contract info request

jpalvarezl:
Addresses part of the issue with #546 

`contract_info` is not fetched with a timeout. Added a 3 second, since this method is repeatedly called in the transaction history list.


jpalvarezl:
Will be fixed for `v3.0.x`

PR #561, opened 2021-08-10
Latest nonce

### LINKED ISSUE #485: Provide latest transaction nonce for clients

Clients need a way to get the latest tx nonce when proposing transactions. On web and iOS it was implemented via `/api/v1/safes/\(safe)/multisig-transactions?limit=1`

As part of removing tx service dependency currently on iOS, we request the first queued transaction and take nonce from it. If there is nothing, we take nonce from the safe info request.

Richard proposed to get the nonce returned as part of the transaction estimation.

TBD

jpalvarezl:
Closes #485 
We default nonce to `0` when there is no multisigTx available (which shouldn't happen)

### DIFF
--- a/src/services/utils.rs
+++ b/src/services/utils.rs
@@ -3,7 +3,7 @@ use crate::models::backend::transactions::{
     MultisigTransaction as BackendMultisigTransaction,
     SafeTransactionEstimation as BackendSafeTransactionEstimation,
 };
-use crate::models::commons::DataDecoded;
+use crate::models::commons::{DataDecoded, Page};
 use crate::models::service::utils::{
     DataDecoderRequest, SafeTransactionEstimation, SafeTransactionEstimationRequest,
 };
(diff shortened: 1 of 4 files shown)


src/services/utils.rs
```diff
@@ -39,21 +43,63 @@ pub async fn estimate_safe_tx_gas(
...
+        .send()
+        .await?;
+
+    Ok(serde_json::from_str::<Page<BackendMultisigTransaction>>(
+        &latest_multisig_tx_response.text().await?,
+    )?
+    .results
+    .first()
+    .map(|it| it.nonce)
+    .unwrap_or(0))
```
fmrsabino, line 104: nitpick: The indentation here looks a bit off (maybe because of fmt). The `Ok()` seems to be only applied to the first statement. I was wondering if we could just extract it in this case.

```rust
let nonce = serde_json::from_str::<Page<BackendMultisigTransaction>>(
        &latest_multisig_tx_response.text().await?,
    )?
    .results
    .first()
    .map(|it| it.nonce)
    .unwrap_or(0);

Ok(nonce)
```

fmrsabino (approved):

src/services/utils.rs
jpalvarezl: I agree, `rustfmt` does not look too nice here. I will add your suggestion, it is also stating more cleary what we expect after the `unwrap_or`

fmrsabino (approved):

PR #563, opened 2021-08-10
Added CLA agreeement action to repo

### LINKED ISSUE #562: Include CLA agreement on PRs

jpalvarezl:
Closes #562


.github/workflows/cla.yml
```diff
@@ -0,0 +1,26 @@
...
+        # Alpha Release
+        uses: gnosis/cla-github-action@master
+        # GitHub token, automatically provided to the action
+        # (No need to define this secret in the repo settings)
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          path-to-signatures: 'signatures/version1/cla.json'
+          path-to-cla-document: 'https://gnosis-safe.io/cla/'
+          branch: 'cla-signatures'
```
fmrsabino, line 23: From the docs it seems that this is going to store the signatures in a branch named `cla-signatures` Under a file named `signatures/version1/cla.json`.

It's a bit weird to have a separate branch to just keep track of the signatures but I guess that's how the action works and why it needs and unprotected branch. But an unprotected branch can be also a trigger for force pushes who change the history of the signatures 🤔

.github/workflows/cla.yml
```diff
@@ -0,0 +1,26 @@
...
+          allowlist: jpalvarezl,rmeissner,Uxio0,bot*,fmrsabino
```
fmrsabino, line 24: If `bot*` is a wildcard I'd be more conservative and only list the ones that we have enabled (`dependabot`)

jpalvarezl:
I just copy and pasted with very minor adjustments to the code from the safe-app-sdk, I will move this back to draft and be more thorough

.github/workflows/cla.yml
jpalvarezl: It seems to be a reserved name: https://github.com/cla-assistant/github-action#1-add-the-following-workflow-file-to-your-repository-in-this-pathgithubworkflowsclayml

.github/workflows/cla.yml
jpalvarezl: I guess the assumption is that we would never use `cla-assitant` as a base branch against which we would open a PR

.github/workflows/cla.yml
fmrsabino: I read the following:

> And also, since there's no way for bot users (such as Dependabot or Greenkeeper) to sign a CLA, you may want to add them in allowlist. You can do so by adding their names in a comma separated string to the allowlist input in the CLA workflow file(in this case dependabot-preview[bot],greenkeeper[bot]). You can also use wildcard symbol in case you want to allow all bot users something like bot*.

So I thought that `bot*` would not work with dependant but maybe it does 👍

fmrsabino (approved):

.github/workflows/cla.yml
jpalvarezl: hahah sorry 🙇 ! you are totally right, I just saw it in the default docs like that too and I assued it. I am wondering how does dependabot match in the safe-apps-sdk repo then 🤔

fmrsabino: Its username is: `dependabot[bot]`. Maybe we can just add that.

PR #574, opened 2021-08-17
Added table tests for asserting env keys and default values

### LINKED ISSUE #494: Adjust `URL` usage in names for env vars and config methods
There is still the methods in `src/config/mod.rs` named `base_config_service_url`, `base_exchange_api_url` and `redis_url` as well as their respective env variables: `CONFIG_SERVICE_URL`, `EXCHANGE_API_BASE_URL` and `REDIS_URL`.

All of them should be renamed into `*URI` or `*uri`

### LINKED ISSUE #510: Cache durations and Request timeouts have different units
It is very confusing that the env vars representing a `cache duration` are in seconds and the ones representing http `request timeouts` are in milliseconds.

The redis API requires seconds and reqwest milliseconds, but we should still not expose this inconsistency in our env variables. 

Change all the values to milliseconds, as it is the unit with the largest resolution of both alternatives.

jpalvarezl:
Closes #494 
Closes #510 

This PR requires coordination with devOps


fmrsabino (commented):
Wouldn't the environment variable changes be independent of the tests? (or do they make sense to go together)

jpalvarezl:
I think it makes sense to add them together, as this is a refactor task, and it is safer to refactor code that's covered by tests.

fmrsabino (approved):

PR #581, opened 2021-08-20
Restructured workflows

### LINKED ISSUE #568: Actions are run twice for internally opened pull request
## Issue

In order to enable external contributions we had to add `on: pull_request` to our workflow. This will trigger a run of a PR opened from a fork, but in the case of an internal branch of the project the event `push` and `pull_request` trigger the workflow twice. 

We need to trigger our workflow only when: 
- push to main
- pushes to PR
- Create a PR (external contributions from forks)
- Tags 

Usage of composite actions could prove useful: https://docs.github.com/en/actions/creating-actions/creating-a-composite-run-steps-action

## Solution 1

Partition our jobs and have a workflow per row as shown in the image below:

image

## Solution 2

Find the combination of trigger events that allows us to run our current workflow in the scenarios listed in the top section of the issue. Something along the lines of:

```yml
on:
  # Trigger the workflow on push or pull request,
  # but only for the main branch
  push:
    branches:
      - main
  pull_request:
    branches:
      - main
  # Also trigger on page_build, as well as release created events
  page_build:
  release:
    types: # This configuration does not affect the page_build event above
      - created
```

For more information: https://docs.github.com/en/actions/reference/events-that-trigger-workflows

jpalvarezl:


.github/workflows/deploy_prod.yml
```diff
@@ -0,0 +1,29 @@
...
+  deploy:
+    runs-on: ubuntu-latest
+    env:
+      DOCKERHUB_ORG: gnosispm
+      DOCKERHUB_PROJECT: safe-client-gateway
+    steps:
+      - name: Checkout branch
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0 # checks out all branches and tags, necessary for versioning /about endpoint
```
fmrsabino, line 20: How does this affect the docker build that happens on `bash scripts/deploy_docker.sh`

.github/workflows/deploy_prod.yml
```diff
@@ -0,0 +1,29 @@
...
+
+      - name: Dockerhub login
+        uses: docker/login-action@v1
+        with:
+          username: ${{ secrets.DOCKER_USER }}
+          password: ${{ secrets.DOCKER_PASSWORD }}
+
+      - name: Deploy tag
+        run: bash scripts/deploy_docker.sh ${GITHUB_REF##*/}
```
fmrsabino, line 29: One suggestion for this is to use `${{ github.event.release.tag_name }}` it represents the tagged version – `vX.Y.Z`

Can probably be done in a separate PR

fmrsabino (approved):

.github/workflows/deploy_prod.yml
jpalvarezl: that would be definitely more readable :+1:

.github/workflows/deploy_prod.yml
jpalvarezl: I think the reason I did this, was so that the script you mention would find the tag. I remember that with a regular checkout of the branch, tags were somehow not included. There is probably a nicer way to do this, as it sounds like a pretty common use case.

.github/workflows/deploy_prod.yml
fmrsabino: I think this is only available on events triggered by the type `release` (not 100% sure)

jpalvarezl:
We decided to not move forward with this

PR #597, opened 2021-08-31
Use cargo-llvm-cov for coverage

fmrsabino:
Make sure these boxes are checked! 📦✅

- [ ] You are using the nightly version of rust `rustup default nighly`
- [ ] You have the latest version of `rustfmt` installed
```bash
$ rustup component add rustfmt
```
- [ ] You ran `cargo fmt` on the code base before submitting


fmrsabino:
Currently we cannot ignore the macros using this tool (which can be achieved with `grcov` with `excl-line` or `excl-br-line`) so I'm closing down this PR.

PR #601, opened 2021-09-01
Changed `baseGas` and `safeTxGas` types to `String`

### LINKED ISSUE #543: `baseGas` and `safeTxGas` types into `String`
Change the types of `base_gas` and `safe_tx_gas` in the `MultisigExecutionDetails` to `String`

jpalvarezl:
Closes #543 

The web team needs to be notified when this is deployed to staging (before we merge)


fmrsabino (approved):
This would be a breaking change for the clients. We should think about versioning endpoints so we have less synchronisation issues and can move more independently with the service development.

PR #611, opened 2021-09-08
`/v1/chains/{chain_id}/safe-apps` forward endpoint

### LINKED ISSUE #595: Add forward endpoint for `/safe-apps` from the config servcie
## Feature

Implement a forward endpoint for the `/safe-apps` config service endpoint.

## Context

The current payload from the config service is as follows:

```json
[
  {
    "id": 0,
    "url": "string",
    "name": "string",
    "iconUrl": "string",
    "description": "string",
    "chainIds": [
      0
    ],
    "provider": {
      "url": "string",
      "name": "string"
    }
  }
]
```

Where the request is (in staging) looks like `https://safe-config.staging.gnosisdev.com/api/v1/safe-apps/?chainId=1`

`chainId` is an integer query parameter for which we filter apps available in certain networks.

## Proposal

We stringgify all the fields that are currently ints, namely `id` and the array `chainIds` turns into an array of stringyfied-ints, like so:

```json
[
  {
    "id": "0",
    "url": "string",
    "name": "string",
    "iconUrl": "string",
    "description": "string",
    "chainIds": [
      "0"
    ],
    "provider": {
      "url": "string",
      "name": "string"
    }
  }
]
```

The request against the config service is cached for as long as we cache `ChainInfo` we would also cache the response but set the duration to something negligible (to avoid issues with flushing).

The CGW **accepts and forwards** the `chainId` query parameter to the config service.

## Questions

Do we need a new payload for the flush endpoint in the CGW to invalidate changes to the underlying data in this endpoint? Do we re-use the `InvalidationPattern::Chains` and change it to be `InvalidationPattern::ConfigService` to simplify things?

jpalvarezl:
Closes #595

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -37,6 +37,7 @@ LOG_ALL_ERROR_RESPONSES=false
 # INDEFINITE_TIMEOUT=1000 # long cache duration of your choice
 # SHORT_ERROR_DURATION=1000
 # LONG_ERROR_DURATION=1000
+# SAFE_APPS_CACHE_DURATION=1000
 
 # Http request time outs
 # The unit of these values is "milliseconds"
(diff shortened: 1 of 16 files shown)


jpalvarezl:
Duplicates #622

PR #612, opened 2021-09-10
Declare futures dependency in Cargo.toml

fmrsabino:
- Given that we use the `futures` crate directly in our codebase, we should declare it directly in order to avoid third-party library changes where this dependency is not guaranteed to be provided.


jpalvarezl:
As far as I remember, from the refactor with Nick, we should definitely use `futures` re-imported by rocket. Transitive dependency usage is prefered in rust, because of rocket itself may not be compatible with a different version of futures, so they are tecnically locking the version of futures they support. 

We do something similar with `eth_abi` and `ethereum_types`, as we would only use a version supported by the library, thusly avoiding runtime errors.

The details about what happens when you have 2 dependencies brought in transitively to project are unknown to me and I would be happy to hear about this from Nick. 

As it stands and AFAIK, using futures from rocket is the correct setup.

fmrsabino:
> As far as I remember, from the refactor with Nick, we should definitely use `futures` re-imported by rocket. Transitive dependency usage is prefered in rust, because of rocket itself may not be compatible with a different version of futures, so they are tecnically locking the version of futures they support.

This should be tackled in cargo itself for dependency resolution. The main issue is that we are assuming that this 3rd party library (rocket) will provide another 3rd party library (Futures) that we directly rely on (note that the distinction is made on the fact that we are directly relying on Futures which is a separate library than Rocket.

Currently rocket does not specify a minimum version to be downloaded for `Futures` (https://crates.io/crates/rocket/0.5.0-rc.1/dependencies).

I guess our guarantee in this case is the generation of the `.lock` file which is tied to a specific version.
 
> We do something similar with `eth_abi` and `ethereum_types`, as we would only use a version supported by the library, thusly avoiding runtime errors.

Why would it be a runtime error? Wouldn't libraries specify the required dependencies (minimum and maximum)?

jpalvarezl:
I would consult with Nick before moving forward, as we used to have this dependency in the past but was removed. I don't remember the exact reason right now.

PR #613, opened 2021-09-10
Sort dependencies lexicographically

fmrsabino:
Sort dependencies in lexicographical order to improve readability and organisation of the dependencies that this project requires

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,43 +7,27 @@ edition = "2018"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
+chrono = { version = "0.4", features = ["serde"] }
+derivative = "2.2.0"
 dotenv = "0.15.0"
 dotenv_codegen = "0.15.0"
-
-regex = "1.5.4"
-lazy_static = "1.4.0"
-
-rocket = { version = "0.5.0-rc.1", features = ["tls", "json"] }
-rocket_codegen = { version = "0.5.0-rc.1" }
-
-# Force rocket peer-dependencies versions
-proc-macro2 = "1.0.28"
-tokio = "1.11.0"
-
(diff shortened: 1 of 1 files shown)


jpalvarezl:
I would love to hear more on the motivation behind this, as previously, the grouping of the dependencies was meant to be topical rather than lexicographical. Is there a standard we'd comply with by introducing this change?

fmrsabino:
> I would love to hear more on the motivation behind this, as previously, the grouping of the dependencies was meant to be topical rather than lexicographical. Is there a standard we'd comply with by introducing this change?

The motivation was when adding a new dependency is that it'd be easy to see where it should be added. In the future it'd also be easy to "locate" it given the lexicographical order.

So given the current state a dependency can be added anywhere under `[dependencies]`?

jpalvarezl:
Well, currently, rather than sorting there is a loose grouping of dependencies. The groups that I identify are `env` , `web framework`, `http client`, `redis` , `eth` and `logging` which are interleaved with miscelaneous stuff. 

I like the grouping, but I also see your point on requiring more clarity on sorting this. I wouldn't mind introducing your proposal, as it would allow for less arbitrary sorting.

As this is still currently a draft, I cannot approve.

fmrsabino:
> Well, currently, rather than sorting there is a loose grouping of dependencies. The groups that I identify are `env` , `web framework`, `http client`, `redis` , `eth` and `logging` which are interleaved with miscelaneous stuff.
> 
> I like the grouping, but I also see your point on requiring more clarity on sorting this. I wouldn't mind introducing your proposal, as it would allow for less arbitrary sorting.
> 
> As this is still currently a draft, I cannot approve.

We can discuss it with the rest of the team 👍

jpalvarezl:
I would say that we just move forward with this :D but if you prefer discussing with the rest that's also fine.

jpalvarezl (approved):

PR #617, opened 2021-09-16
Add description field to Chain

fmrsabino:
- Adds `description` field to `ChainInfo` (both backend and client models)

### DIFF
--- a/src/json/chains/rinkeby.json
+++ b/src/json/chains/rinkeby.json
@@ -2,6 +2,7 @@
   "chainId": "4",
   "chainName": "Rinkeby",
   "l2": false,
+  "description": "Random description",
   "rpcUri": {
     "authentication": "API_KEY_PATH",
     "value": "https://someurl.com/rpc"
--- a/src/json/chains/rinkeby_fixed_gas_price.json
+++ b/src/json/chains/rinkeby_fixed_gas_price.json
@@ -2,6 +2,7 @@
   "chainId": "4",
   "chainName": "Rinkeby",
   "l2": false,
+  "description": "Random description",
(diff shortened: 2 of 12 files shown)


jpalvarezl (commented):
Just one comment, just to be totally certain that I understand this correctly, according to this migration: https://github.com/gnosis/safe-config-service/blob/774e382c36948cfe9937b1c9b25b0959cc01201b/src/chains/migrations/0027_chain_description.py#L16

The field is `blankable` but **not** `nullable` ? Would it make more sense to make it nullable? `description` fields I have the impression are generally optional anyhow, but that could just be observer's bias.

fmrsabino:
> Just one comment, just to be totally certain that I understand this correctly, according to this migration: https://github.com/gnosis/safe-config-service/blob/774e382c36948cfe9937b1c9b25b0959cc01201b/src/chains/migrations/0027_chain_description.py#L16
> 
> The field is `blankable` but **not** `nullable` ? Would it make more sense to make it nullable? `description` fields I have the impresion are generally optional anyhow, but that could just be observer's bias.

That is correct 👍 This requirement came from the client side where it was requested for it to be an empty string instead of `null`: https://github.com/gnosis/safe-config-service/issues/216#issuecomment-919900536

jpalvarezl (approved):

fmrsabino:
@gnosis/safe-web Available on `staging` environment: https://safe-client.staging.gnosisdev.com/v1/chains

jpalvarezl:
This would prevent us from deploying the GW to production until this change reaches production for the config service, right @fmrsabino ? Is there already a date where this will reach production?

PR #631, opened 2021-09-29
Enable public core services url on debug builds

### LINKED ISSUE #629: Introduce `transactionService` and `vpcTransactionService` toggle env var
### Feature

From `v3.6.0` local development won't be possible as we would be using VPC internal URLs even in staging.

We need an easy way to use the external URL for the services.

### Proposal

Introduce env var `USE_EXTERNAL_TX_SERVICE_URI=<bool>` as a mean to toggle between using `transactionService` and `vpcTransactionService` URLs in the `core_uri!` macro.

fmrsabino:
Closes #629 

- The `core_uri` macro now uses the public transaction service url if `debug_assertions` is `True` – this means that the public url will be used by non-optimised builds by default
- To test with the internal cluster url the `--release` flag can be used (in order to use the `release` profile which has `debug_assertions` disabled) – `cargo run --release` (see https://doc.rust-lang.org/cargo/reference/profiles.html#release)

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -1,3 +1,5 @@
+use crate::models::backend::chains::ChainInfo;
+
 macro_rules! concat_parts {
     ($parts_head:expr) => {
         // `stringify!` will convert the expression *as it is* into a string.
@@ -65,13 +67,23 @@ macro_rules! to_hex_string {
     }};
 }
 
+#[cfg(debug_assertions)]
+pub fn get_transaction_service_host(chain_info: ChainInfo) -> String {
+    return chain_info.transaction_service;
+}
+
+#[cfg(not(debug_assertions))]
+pub fn get_transaction_service_host(chain_info: ChainInfo) -> String {
+    return chain_info.vpc_transaction_service;
+}
+
 #[doc(hidden)]
 #[macro_export]
 macro_rules! core_uri {
     ($info_provider:tt, $path:expr) => {{
         let result: ApiResult<String> =
         match $info_provider.chain_info().await {
-            Ok(chain_info) => Ok(format!("{}/api{}",chain_info.vpc_transaction_service, $path)),
+            Ok(chain_info) => Ok(format!("{}/api{}",crate::macros::get_transaction_service_host(chain_info), $path)),
             Err(error) => Err(error,)
         };
         result
--- a/src/services/tests/backend_url.rs
+++ b/src/services/tests/backend_url.rs
@@ -1,3 +1,4 @@
+#[cfg(not(debug_assertions))]
 use crate::models::backend::chains::{
     BlockExplorerUriTemplate, ChainInfo, GasPrice, NativeCurrency, RpcAuthentication, RpcUri, Theme,
 };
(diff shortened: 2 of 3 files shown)


src/macros.rs
```diff
@@ -65,13 +67,23 @@ macro_rules! to_hex_string {
...
 
+#[cfg(debug_assertions)]
+pub fn get_transaction_service_host(chain_info: ChainInfo) -> String {
+    return chain_info.transaction_service;
+}
+
+#[cfg(not(debug_assertions))]
+pub fn get_transaction_service_host(chain_info: ChainInfo) -> String {
+    return chain_info.vpc_transaction_service;
+}
```
jpalvarezl, line 78: I would make these methods private. We don't need to expose them outside the macro context

src/services/tests/backend_url.rs
```diff
@@ -54,7 +54,7 @@ async fn core_uri_success_with_params() {
         exclude_spam
     );
 
-    assert_eq!(url.unwrap(), "http://mainnet-safe-transaction-web.safe.svc.cluster.local/api/v1/safes/0x1230B3d59858296A31053C1b8562Ecf89A2f888b/balances/usd/?trusted=false&exclude_spam=true".to_string());
+    assert_eq!(url.unwrap(), "https://safe-transaction.mainnet.gnosis.io/api/v1/safes/0x1230B3d59858296A31053C1b8562Ecf89A2f888b/balances/usd/?trusted=false&exclude_spam=true".to_string());
```
jpalvarezl, line 57: This test is incorrect 🤔 we need to validate that we are using the vpc URL here, since this test asserts the URL formed to request against the core services. I am started to become more inclined towards having this as an env var. I will add another reason as a final comment on this PR :)

jpalvarezl (commented):
I would double check what happens with `docker compose up`. Judging by the `Dockerfile` we build a `release` binary, which would render the `docker-compose.yml` file unusable unless used along with a custom instance of the config service. The env var would allow us to work around this shortcoming I think 🤔

src/macros.rs
fmrsabino: I think they need to be public to be accessible from the macro itself.

src/services/tests/backend_url.rs
fmrsabino: I'm not sure if I understood why the test is incorrect. If it's being tested under a non optimised build then this should be the result. It might not provide the value that we want true but for that I'd look into adding a test for the optimised builds.

fmrsabino: Pushed a suggestion in 3c212e4

It works for both `cargo test` and `cargo test --release`

I added `--release` to the CI so that the release tests run.

src/macros.rs
fmrsabino: Just confirmed and they do need to be public else `error[E0603]` is raised (`function `get_transaction_service_host` is private`)

src/services/tests/backend_url.rs
fmrsabino: Coverage files are not generated this way. I will take a look ☝️

fmrsabino: Actually I prefer to not go the `cargo test --release` and revert the test changes. So it tests against the debug build. I'm wondering if I can test just these two tests with a specific profile (`release` profile) 🤔

jpalvarezl: I will have a look at this maybe: https://github.com/actions-rs/grcov next week, in case something can be done regarding where the report is generated :)

PR #632, opened 2021-09-29
Balances endpoint optimization

### LINKED ISSUE #588: Split Safe balance from fiat rates
## Current state

Currently the Client Gateway uses `GET /api/v1/safes/{address}/balances/usd/` for retrieving the balances of each safe.

In the transaction-service the following flow is:
- `/balances`:
  1. ERC20 tokens used by safe (cached)
  2. Check balances for each token (node request)
- `/balances/usd` – all steps from `/balances` plus:
  3. Get ETH/USD price (cached for 30 minutes)
  4. Get Token/ETH price (redis cache)

## Problem

In order to improve the granularity of the caching that we perform on the Client Gateway we can think that the endpoint does two main things:

- Fetches the balance (Ether + Tokens) of the Safe
- For each token fetch the correspondent fiat rate (Token / USD)

Currently this is done in a bundled manner from the CGW ie.: the balances and fiat rates are retrieved via a single endpoint which makes it more difficult to tweak how we control the cache validity of balances vs fiat rates

## Solutions 

Splitting the balances with from the fiat rates allows us to better control for how long we want to cache the balances vs for how long we want to cache the fiat rates.

Moreover, the balance updates require a node call so it can be seen as a more expensive call then the rate calls (depending on the service).

Some proposals:
- For each safe, check indexed tokens for that safe and return the rates for that safe. See https://github.com/gnosis/safe-transaction-service/pull/334
- The CGW, knowing the balance information for each safe (using `/balances` endpoint), would request for each token, its fiat rate
- Similar to the point above but bundling all the token fiat rate requests into one request (and cache the bundled response)
- Similar to the point above but cache each token entry of the bundled response so that it can be reused by other safe requests.

fmrsabino:
Closes #588


src/models/converters/balances.rs
```diff
@@ -2,11 +2,30 @@ use crate::models::backend::balances::Balance as BalanceDto;
...
+            .token
+            .as_ref()
+            .and_then(|token| Some(token.decimals))
+            .unwrap_or(native_coin.decimals);
+
+        // We try to convert the token_decimals to f64, if it fails we fallback to the native currency decimals (f64), if it fails the function panics
+        let decimal_exp =
+            f64::from_u64(token_decimals).unwrap_or(f64::from_u64(native_coin.decimals).unwrap());
+
+        let token_balance =
```
rmeissner, line 24: Is this datatype large enough? Is there something like BigInteger or BigNumber in Rust?

src/services/balances.rs
```diff
@@ -2,12 +2,16 @@ use crate::cache::cache_operations::RequestCached;
...
 use crate::models::service::balances::{Balance, Balances};
 use crate::providers::fiat::FiatInfoProvider;
 use crate::providers::info::{DefaultInfoProvider, InfoProvider};
 use crate::utils::context::Context;
 use crate::utils::errors::ApiResult;
-use std::cmp::Ordering;
+use rocket::futures::{stream, StreamExt};
+use std::str::FromStr;
+
+const N_CONCURRENT_TOKEN_REQUESTS: usize = 5;
```
rmeissner, line 14: This should probably be a config/env var

src/services/balances.rs
```diff
@@ -43,28 +47,90 @@ pub async fn balances(
...
+            balance
+                .token_address
+                .to_owned()
+                .unwrap_or("0x0000000000000000000000000000000000000000".to_string())
+        })
+        .collect::<Vec<_>>();
+
+    // We collect the TokenPrice which were successful – unsuccessful ones are ignored
+    let token_prices: Vec<TokenPrice> = stream::iter(token_addresses)
+        .map(|token_address| get_token_usd_rate(context, chain_id, token_address))
```
jpalvarezl, line 62: You can call `.ok()` on `Result` types, then just `filter(|it| it.some())`

src/services/balances.rs
```diff
@@ -43,28 +47,90 @@ pub async fn balances(
...
+        .map(|balance| {
+            balance
+                .token_address
+                .to_owned()
+                .unwrap_or("0x0000000000000000000000000000000000000000".to_string())
+        })
+        .collect::<Vec<_>>();
+
+    // We collect the TokenPrice which were successful – unsuccessful ones are ignored
+    let token_prices: Vec<TokenPrice> = stream::iter(token_addresses)
```
jpalvarezl, line 61: can the type argument be passed to `collect` instead?

src/models/converters/balances.rs
```diff
@@ -2,11 +2,30 @@ use crate::models::backend::balances::Balance as BalanceDto;
...
+    ) -> Balance {
+        let token_decimals = self
+            .token
+            .as_ref()
+            .and_then(|token| Some(token.decimals))
+            .unwrap_or(native_coin.decimals);
+
+        // We try to convert the token_decimals to f64, if it fails we fallback to the native currency decimals (f64), if it fails the function panics
+        let decimal_exp =
+            f64::from_u64(token_decimals).unwrap_or(f64::from_u64(native_coin.decimals).unwrap());
```
jpalvarezl, line 22: Why `panic` instead? We used to have `0.0` so that we don't get rid of the entire reponse if one parsing fails.

fmrsabino: Token decimals as `0` might increase a lot the total balance giving the user the wrong data. I think failing here is reasonable since it relies on the native currency as a backup. If that fails then something might be really off.

src/models/converters/balances.rs
fmrsabino: Regarding Big Numbers in Rust I found the following discussion with some suggestions: https://www.reddit.com/r/rust/comments/a7frqj/have_anyone_reviewed_any_of_the_decimal_crates/

I've decided to go with the the 64 bit version already present for development but we can discuss about possible implementations for representing this value 👍

src/services/balances.rs
```diff
@@ -43,28 +47,90 @@ pub async fn balances(
...
+/// To retrieve the Native Currency fiat price of the chain (eg.: Ether), 0x0000000000000000000000000000000000000000 should be used
+///
+/// # Arguments
+///
+/// * `context`: The context where the request will be executed
+/// * `chain_id`: The chain id on which this request should be executed
+/// * `token_address`: The token address (0x0000000000000000000000000000000000000000 for native currency)
+///
+/// returns: Result<TokenPrice, ApiError>
+///
```
jpalvarezl, line 109: the module `services` is hidden from the cargo documentation generator. Documentation, such as this one is normally placed at the `route` module level.

src/services/balances.rs
fmrsabino: I can add it as a environment variable. Didn't do it because I was trying to avoid some "configuration" overload. But for this one might make sense 👍

src/models/converters/balances.rs
jpalvarezl: We could find a more sensible default for decimals though. Maybe `18`? I think maybe this should be discussed just to be sure that we are in sync with the expected behaviour 🤔

src/services/balances.rs
fmrsabino: I thought this could still be used by editors/IDEs or even just for reading purposes. Eg.: IntelliJ generates the following:
image

But I can remove it 👍

jpalvarezl: You can just move it to `routes/balances` :) Or just adjust the ones there, as those are the ones that get pushed to the docs in github pages

src/models/converters/balances.rs
Uxio0: I would prefer `0.0` as a default and not panicking. I think for cases that wa cannot parse something using a `f64` values couldn't be very high so it won't impact user balance. This is an exception we can recover from, I would reserve `panic` for situations in where we can not

Uxio0:
Looks good! 🚀

PR #635, opened 2021-10-04
Config flag to toggle between `vpcTransactionService` and `transactionService`

### LINKED ISSUE #629: Introduce `transactionService` and `vpcTransactionService` toggle env var

jpalvarezl:
Closes #629 
Closes #631 

In addition to what @fmrsabino suggested, I introduced tests with debug assertion to also validate what is ran in the local setup unoptimized case.

However, after the discussion in the comments we have decided to stick to an implementation based on a environment variable, as this would allow better testability. 

Important note: From this PR onward tests would run in single thread, as changes to env vars behave non-deterministically in an async setup.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -58,7 +58,7 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: test
-          args: --verbose --all --locked
+          args: --verbose --all --locked -- --test-threads 1
         env:
           CARGO_INCREMENTAL: '0'
           RUSTFLAGS: '-Zprofile -Ccodegen-units=1 -Copt-level=0 -Clink-dead-code -Coverflow-checks=off -Awarnings'
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -58,7 +58,7 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: test
-          args: --verbose --all --locked
+          args: --release --verbose --all --locked
         env:
           CARGO_INCREMENTAL: '0'
           RUSTFLAGS: '-Zprofile -Ccodegen-units=1 -Copt-level=0 -Clink-dead-code -Coverflow-checks=off -Awarnings'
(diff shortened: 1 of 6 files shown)


Uxio0 (commented):
I would prefer to make it configurable via environment variable instead of compiling flags. Reasoning:
- More flexible and easy to use
- More visible. This feature could be shadowed in the future and at some point we might not know why the service behaves in a different way when testing locally

(cc @rmeissner  @fmrsabino)

jpalvarezl:
I find myself inclined towards the `env var` impl too. As it would allow setting the env var per test, then we don't need to sacrifice testability. We can assert all execution paths without the need to have optimised and unoptimised builds simultaneously. 

Granted, this PR does not address this, we still only check `test` builds.

jpalvarezl:
I added the env var implementation. However, I had to additionally make the tests run in a single thread, as the setting of the env var per test  was behaving non-deterministically, as our tests are `async` functions.

.github/workflows/rust.yml
```diff
@@ -58,7 +58,7 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: test
-          args: --verbose --all --locked
+          args: --verbose --all --locked -- --test-threads 1
```
rmeissner, line 61: Why are these changes required?

rmeissner (commented):
How does this relate to https://github.com/gnosis/safe-client-gateway/pull/631?  Would this replace that one?

.github/workflows/rust.yml
rmeissner: Ahhh ... didn't see the PR comment when reviewing ... could we add this to the description of the PR?

rmeissner (approved):

jpalvarezl:
I understood #631 PR as a exploratory PR where @fmrsabino evaluated how that alternative would look like. If we merge this PR then we would have decided to go with the env var alternative.

PR #639, opened 2021-10-05
Forward per chain notification request error

### LINKED ISSUE #618: Return better error messages when registering for push notifications
Error messages when registering push notifications are not very useful:

```json
{
  "code":1337, 
  "message":"Push notification registration failed for chain ids: 1, 4, 137"
}
```

Relaying error messages from the tx service. Something like:

```json
{
  "code":1337, 
  "message":"Push notification registration failed for chain ids: 1, 4, 137"
  "details": {
    1: "Cloud messaging token is linked to another device",
    2: "Not valid signature",
    3: "Cloud messaging token is linked to another device"
}
```

**Additional context**

jpalvarezl:
Closes #618


src/services/notifications.rs
```diff
@@ -57,29 +58,43 @@ pub async fn post_registration(
...
-            if let Ok(response) = request.await {
-                // tx service errors
-                if let Err(_) = forward_error(response).await {
-                    output.push(chain_id);
+            match request.await {
+                Ok(response) => {
+                    if !response.status().is_success() {
+                        output.insert(
+                            chain_id,
+                            response.text().await.expect("Error response issue"),
```
rmeissner, line 68: Is this then a json string?

src/services/notifications.rs
```diff
@@ -24,13 +25,13 @@ pub async fn delete_registration(
...
     )?;
 
-    let response = client
+    client
         .delete(url)
         .timeout(Duration::from_millis(default_request_timeout()))
         .send()
         .await?;
 
-    forward_error(response).await
```
rmeissner, line 33: Why was this required before?

jpalvarezl: It was not required. That method was only returning `Ok(())` if the response was success or mapping the error response to an `ApiError` which in the case of this method, would be done by the `responder` implementation for `ApiError` anyway.

This call was totally redundant.

src/services/notifications.rs
jpalvarezl: Yes, this the `json` of the error returned by the core services.

rmeissner:
The expected response defined in the issue https://github.com/gnosis/safe-client-gateway/issues/618 is different

PR #640, opened 2021-10-05
204 status code forced for a pre-flight OPTION requests

### LINKED ISSUE #204: Improve coverage for history tx endpoint
Add tests to functions where mocking is possible. Generally, **not** where `Context` is passed.

### LINKED ISSUE #623: Cors failing for POST requests
When performing the Option request for POST endpoints (e.g. the estimations endpoint) an error is returned. 

A quick work around is to force the response code to be 200 in the CORS trait.

jpalvarezl:
Closes #623 

https://github.com/lawliet89/rocket_cors/blob/master/src/fairing.rs#L90-L98

### DIFF
--- a/src/utils/cors.rs
+++ b/src/utils/cors.rs
@@ -14,7 +14,7 @@ impl Fairing for CORS {
     }
 
     async fn on_response<'r>(&self, request: &'r Request<'_>, response: &mut Response<'r>) {
-        // TODO https://github.com/lawliet89/rocket_cors/blob/master/examples/fairing.rs
+        //https://github.com/lawliet89/rocket_cors/blob/master/examples/fairing.rs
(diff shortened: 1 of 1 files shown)


src/utils/cors.rs
```diff
@@ -30,9 +29,10 @@ impl Fairing for CORS {
             response.set_header(Header::new("Access-Control-Allow-Credentials", "true"));
         }
 
-        if request.method() == Method::Options {
+        if request.method() == Method::Options && request.route().is_none() {
```
rmeissner, line 32: It might make sense to put all of this into 1 `if` (and maybe remove the `response.content_type() == Some(ContentType::JSON)` condition) ... but would have to test this :thinking:

rmeissner (approved):

PR #670, opened 2021-10-26
Added CacheManager to have several redis pool connections

### LINKED ISSUE #344: Redis database id selection for `CacheResponse` and `RequestCached`
The enum `Database` is already defined and builder methods are already in place, the redis database needs to be selected before executing the `cache_op_executor` function.

jpalvarezl:
Closes #344


fmrsabino:
Converted to draft as this will be part of a major release which will come after the planned `3.10.0`

fmrsabino (changes requested):
So the goal of this PR is to introduce multiple Redis databases in a single instance.

`REDIS_DEFAULT_URI` would be used as a default one – `redis/0`
`REDIS_INFO_URI` would be used to store chain information – `redis/1`

After some research it does look like this is not recommended practice and was bad design decision made at the time (stated here by the developer https://groups.google.com/g/redis-db/c/vS5wX8X4Cjg/m/8ounBXitG4sJ).

This is additionally stated in point 3 of https://redis.com/blog/7-redis-worst-practices/

> Salvatore Sanfilippo, the author of Redis, once called numbered databases the worst design mistake he ever made in Redis. This design choice is a cautionary tale in building something that looks like it does one thing but actually does another. Thankfully, while this is becoming less common in the wild, Redis still ships with the ability to switch between different “databases” with the SELECT command. Each database is isolated from a key perspective. So, key foo:bar on database 0 can be completely distinct from foo:bar in database 9. This all sounds rather nice, right? The problem is that these databases are not isolated in any other way. Running KEYS on database 0 will still freeze database 9. In effect, it looks like you can run independent workloads on each database, but in reality they aren’t independent at all.

> A bummer, but not really a worst practice, right? Well, the problem is that numbered databases are not well supported throughout the ecosystem. The first, and probably most dire, nail in the coffin of numbered databases is that they aren’t supported by any clustering system (open source nor Redis Enterprise clustering). In effect, you will never be able to leave a single node of Redis. Also, some modules do not support numbered databases.

What is your perspective on this? I see the advantage of making it easier to separate on a "namespace/feature" level on the database layer but the points stated above make me consider alternative solutions (or to just keep the keys as is).

jpalvarezl:
Thanks for the links @fmrsabino ! I was not aware of any of that. I see that this will bring long term problems that we should definitely avoid.

PR #686, opened 2021-11-09
Tests for `post_confirmation` endpoint

### LINKED ISSUE #679: Unit tests for `routes/post_confirmation`
Cover error (400; already submitted error) and success path

jpalvarezl:
Closes #679

Introduced changes:
- Better usage of `InfoProvider` in the `tx_details` endpoint prevents multiple calls to the config service per request
- Composing with `submit_confirmation` now re-uses and returns directly the `TransactionDetails`, as opposed to the previous decoupled approach
- `trait InfoProvider` now has two new methods `cache` and `client` which return `Arc<dyn Cache>` and `Arc<dyn HttpClient>` respectively. This is to prevent the usage of `RequestContext` everywhere which would encourage the instancing of new `InfoProvider` for a single request, which should be avoided (so we can leverage in memory caches (namely, `Mutex` values in `DefaultInfoProvider`))
- Removed `Database` fields and methods for `CacheResponse` and `RequestCached` as they were not used on won't be in the forseeable future.


src/routes/transactions/tests/routes.rs
```diff
@@ -0,0 +1,166 @@
...
+            .times(1) // From FETCHING CANCELLATION AND FROM ENRICHING TX DETAILS
+            .with(eq(safe_request))
+            .returning(move |_| {
+                Ok(Response {
+                    body: String::from(crate::tests::json::SAFE_WITH_MODULES),
+                    status_code: 200,
+                })
+            });
+
+        // Cancellation tx
```
jpalvarezl, line 98: @fmrsabino It is possible to remove the specific mocking for `Cancelation TX` and `Known Addresses` as we would be able to catch this we the last method. Maybe it is more appropriate to test this in the tests for `TxDetails` ?

jpalvarezl (commented):
Still need to assert the output `json`

.github/workflows/rust.yml
```diff
@@ -53,6 +53,7 @@ jobs:
       REDIS_URI: redis://localhost:6379
       WEBHOOK_TOKEN: test_webhook_token
       CONFIG_SERVICE_URI: https://config.service.url
+      VPC_TRANSACTION_SERVICE_URI: 'false'
```
fmrsabino, line 56: We should write this down in the `README` – how to do setup the test environment locally.

src/routes/transactions/handlers/proposal.rs
```diff
@@ -11,7 +13,7 @@ pub async fn submit_confirmation(
     chain_id: &str,
     safe_tx_hash: &str,
     signature: &str,
-) -> ApiResult<()> {
+) -> ApiResult<TransactionDetails> {
```
fmrsabino, line 16: This is definitely nicer and more in line with the other handlers 🚀

src/routes/transactions/tests/routes.rs
```diff
@@ -0,0 +1,429 @@
...
+use rocket::local::asynchronous::Client;
+use serde_json::json;
+
+#[rocket::async_test]
+async fn post_confirmation_success() {
+    let mock_http_client = {
+        let mut mock_http_client = MockHttpClient::new();
+
+        let mut chain_request = Request::new(config_uri!("/v1/chains/{}/", 4));
+        chain_request.timeout(Duration::from_millis(chain_info_request_timeout()));
```
fmrsabino, line 26: What happens if we don't have this timeout with the mock? This should be negligible right?

src/routes/transactions/tests/routes.rs
```diff
@@ -0,0 +1,429 @@
...
+                    body: String::from(crate::tests::json::CHAIN_INFO_RINKEBY),
+                })
+            });
+
+        // CONFIRMATION REQUEST
+        let mut backend_request = Request::new(
+                "https://safe-transaction.rinkeby.staging.gnosisdev.com/api/v1/multisig-transactions/0x2e4af4b451a493470f38625c5f78f710f02303eb32780896cb55357c00d48faa/confirmations/"
+                .to_string(),
+        );
+        backend_request.body(Some(json!({"signature": "bd42f5c205b544cc6397c8c2e592ca4ade02b8681673cc8c555ff1777b002ee959c3cca243a77a2de1bbe1b61413342ac7d6416a31ec0ff31bb1029e921202ee1c"}).to_string()));
```
fmrsabino, line 43: We should look into creating factories for most of these objects. Right now we are mostly relying on data from a json file and it can be more tricky to see why we need values like `0x2e4af4b451a493470f38625c5f78f710f02303eb32780896cb55357c00d48faa` and `bd42f5c205b544cc6397c8c2e592ca4ade02b8681673cc8c555ff1777b002ee959c3cca243a77a2de1bbe1b61413342ac7d6416a31ec0ff31bb1029e921202ee1c`.

With a Factory, we'd generate random (valid) data for all the fields except for the ones which actually matter for the test.

src/providers/info.rs
```diff
@@ -100,6 +101,9 @@ pub trait InfoProvider {
     async fn address_ex_from_any_source(&self, address: &str) -> ApiResult<AddressEx>;
     async fn address_ex_from_contracts(&self, address: &str) -> ApiResult<AddressEx>;
     fn chain_id(&self) -> &str;
+
+    fn client(&self) -> Arc<dyn HttpClient>;
+    fn cache(&self) -> Arc<dyn Cache>;
```
fmrsabino, line 106: Maybe it's better to add to the PR description the context of this change (since from the PR it seems mostly test related)

src/utils/http_client.rs
```diff
@@ -1,4 +1,4 @@
-use crate::config::default_request_timeout;
+use crate::config::{default_request_timeout, internal_client_connect_timeout};
```
fmrsabino, line 1: Since `internal_client_connect_timeout` is not used in the tests we can include it only on non-test builds (avoids the warning regarding an unused dependency in tests). Example:

```rust
use crate::config::default_request_timeout;
#[cfg(not(test))]
use crate::config::internal_client_connect_timeout;
```

fmrsabino (approved):
Nice. Left some comments about somethings that we can tackle in the future.

src/routes/transactions/tests/routes.rs
jpalvarezl: It will fail to match the `Request` object when trying to find the correct mock for the `get` method call. It matches the whole instance of `Request`.

src/providers/info.rs
jpalvarezl: Yeah, this is the part where the barrier gets a bit blurry, as this was needed so that we could use the same `info_provider` instance throughout the same request. 

I will document this in the description like you suggest 👍

src/utils/http_client.rs
jpalvarezl: Yeah I was also bothered by the warnings where there was seemingly no way around it. Good suggestion!

src/routes/transactions/tests/routes.rs
jpalvarezl: I would love to explore this further. I think it is going to be easier to define these reqs as we write more tests. But it is defintely something we need to tackle early on so we don't litter the code too much.

fmrsabino (approved):
🚀

PR #699, opened 2021-11-23
Add safe_apps_rpc_uri to ChainInfo

fmrsabino:
- `safe_apps_rpc_uri` represents a different RPC url that can be used by the clients in the context of `safe-apps`
-  It was already part of the `/api/v1/chains` of the `safe-config-service` but was never exposed to the clients via the `safe-client-gateway`


jpalvarezl (approved):

fmrsabino:
@gnosis/safe-web Merged and deployment is ongoing. You will be able to see the new payload under https://safe-client.staging.gnosisdev.com/v1/chains

PR #714, opened 2021-11-29
Fix balances route for balances_v2

fmrsabino:
- Use `/v1/safes/{}/balances/` instead of `/v1/safes/{}/balances/usd/` for balances. This was one of the main purposes of the new feature but it was still using the old transaction service route
- Balance requests to the core services now follow `BALANCES_CORE_REQUEST_CACHE_DURATION` which is set to 1 hour by default
- Token requests to the core services are now cached for 10 seconds

### DIFF
--- a/src/config/mod.rs
+++ b/src/config/mod.rs
@@ -78,7 +78,7 @@ pub fn balances_cache_duration() -> usize {
     env_with_default("BALANCES_REQUEST_CACHE_DURATION", 60 * 1000)
 }
 
-pub fn balances_core_cache_duration() -> usize {
+pub fn balances_core_request_cache_duration() -> usize {
     env_with_default("BALANCES_CORE_REQUEST_CACHE_DURATION", indefinite_timeout())
 }
 
--- a/src/config/mod.rs
+++ b/src/config/mod.rs
@@ -78,6 +78,10 @@ pub fn balances_cache_duration() -> usize {
     env_with_default("BALANCES_REQUEST_CACHE_DURATION", 60 * 1000)
 }
 
+pub fn balances_core_cache_duration() -> usize {
+    env_with_default("BALANCES_CORE_REQUEST_CACHE_DURATION", indefinite_timeout())
+}
+
 pub fn safe_app_manifest_cache_duration() -> usize {
     env_with_default("SAFE_APP_MANIFEST_CACHE_DURATION", indefinite_timeout())
 }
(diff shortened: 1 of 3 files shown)


src/config/mod.rs
```diff
@@ -78,6 +78,10 @@ pub fn balances_cache_duration() -> usize {
     env_with_default("BALANCES_REQUEST_CACHE_DURATION", 60 * 1000)
 }
 
+pub fn balances_core_cache_duration() -> usize {
+    env_with_default("BALANCES_CORE_REQUEST_CACHE_DURATION", indefinite_timeout())
```
jpalvarezl, line 82: I would match the method name to the env var name. Also add entry in the test in `config/tests/mod.rs` where we assert default values and env var names.

As an optional comment: I would suggest renaming to `balance_request_cache_duration` and the env var `BALANCE_REQUEST_CACHE_DURATION`

jpalvarezl (approved):
I am approving as to not block this, but I would be happy to see the comments about matching env var to function name and adding an entry to our env table test implemented :) 

Thanks for the quick action 🙌

src/config/mod.rs
```diff
@@ -91,7 +95,7 @@ pub fn safe_apps_cache_duration() -> usize {
 }
 
 pub fn token_price_cache_duration() -> usize {
-    env_with_default("TOKEN_PRICE_CACHE_DURATION", 1) // set to negligible value
+    env_with_default("TOKEN_PRICE_CACHE_DURATION", 10 * 1000)
```
rmeissner, line 98: Should we do the test stuff @jpalvarezl mentioned for this one too?

rmeissner (approved):

src/config/mod.rs
fmrsabino: > I would match the method name to the env var name. Also add entry in the test in config/tests/mod.rs where we assert default values and env var names.

Changed in 527a259

> As an optional comment: I would suggest renaming to balance_request_cache_duration and the env var BALANCE_REQUEST_CACHE_DURATION

There's already a `BALANCES_REQUEST_CACHE_DURATION` which is for the CGW route. That's why I create a new one with `BALANCES_CORE_REQUEST_CACHE_DURATION`

src/config/mod.rs
fmrsabino: I added a default value test here f6b379a

PR #719, opened 2021-11-29
`TransactionStatus::Rejected` returned

### LINKED ISSUE #696: Map rejected transactions as "REJECTED"
Rejected transactions remain `"AWAITING_EXECUTION"` and it is not possible to determine whether it is indeed rejected or not.

When a transaction is/was rejected, the status should be set to `"REJECED"`.

jpalvarezl:
Closes #696


jpalvarezl:
`rejectors` list will be used instead to achieve the current required spec

PR #721, opened 2021-12-01
Add public_rpc_uri to ChainInfo

fmrsabino:
See https://github.com/gnosis/safe-config-service/issues/324 for more details

- Add `public_rpc_uri` to `ChainInfo` – this field represents a public RPC URI that can be set with external frameworks/wallets (eg.: metamask)

### DIFF
--- a/src/routes/chains/converters.rs
+++ b/src/routes/chains/converters.rs
@@ -36,6 +36,16 @@ impl From<BackendChainInfo> for ServiceChainInfo {
                 },
                 value: chain_info.safe_apps_rpc_uri.value,
             },
+            public_rpc_uri: ServiceRpcUri {
+                authentication: match chain_info.public_rpc_uri.authentication {
+                    RpcAuthentication::ApiKeyPath => ServiceRpcAuthentication::ApiKeyPath,
+                    RpcAuthentication::NoAuthentication => {
+                        ServiceRpcAuthentication::NoAuthentication
+                    }
+                    RpcAuthentication::Unknown => ServiceRpcAuthentication::Unknown,
+                },
+                value: chain_info.public_rpc_uri.value,
+            },
             block_explorer_uri_template: ServiceBlockExplorerUriTemplate {
                 address: chain_info.block_explorer_uri_template.address,
                 tx_hash: chain_info.block_explorer_uri_template.tx_hash,
--- a/src/routes/chains/converters.rs
+++ b/src/routes/chains/converters.rs
@@ -16,9 +16,9 @@ impl From<BackendChainInfo> for ServiceChainInfo {
             short_name: chain_info.short_name,
             l2: chain_info.l2,
             description: chain_info.description,
-            rpc_uri: ServiceRpcUri::from(chain_info.rpc_uri),
-            safe_apps_rpc_uri: ServiceRpcUri::from(chain_info.safe_apps_rpc_uri),
-            public_rpc_uri: ServiceRpcUri::from(chain_info.public_rpc_uri),
+            rpc_uri: chain_info.rpc_uri.into(),
+            safe_apps_rpc_uri: chain_info.safe_apps_rpc_uri.into(),
+            public_rpc_uri: chain_info.public_rpc_uri.into(),
             block_explorer_uri_template: ServiceBlockExplorerUriTemplate {
                 address: chain_info.block_explorer_uri_template.address,
                 tx_hash: chain_info.block_explorer_uri_template.tx_hash,
--- a/src/routes/chains/converters.rs
+++ b/src/routes/chains/converters.rs
@@ -1,5 +1,5 @@
 use crate::common::models::backend::chains::{
-    ChainInfo as BackendChainInfo, GasPrice, RpcAuthentication,
+    ChainInfo as BackendChainInfo, GasPrice, RpcAuthentication, RpcUri as BackendRpcUri,
 };
 use crate::routes::chains::models::{
     BlockExplorerUriTemplate as ServiceBlockExplorerUriTemplate, ChainInfo as ServiceChainInfo,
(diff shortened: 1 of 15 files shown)


src/routes/chains/converters.rs
```diff
@@ -36,6 +36,16 @@ impl From<BackendChainInfo> for ServiceChainInfo {
...
+            public_rpc_uri: ServiceRpcUri {
+                authentication: match chain_info.public_rpc_uri.authentication {
+                    RpcAuthentication::ApiKeyPath => ServiceRpcAuthentication::ApiKeyPath,
+                    RpcAuthentication::NoAuthentication => {
+                        ServiceRpcAuthentication::NoAuthentication
+                    }
+                    RpcAuthentication::Unknown => ServiceRpcAuthentication::Unknown,
+                },
+                value: chain_info.public_rpc_uri.value,
+            },
```
jpalvarezl, line 48: I am starting to thing that we could add an implementation of `impl From<RpcUri> for ServiceRpcUri`. Wdyt?

jpalvarezl (approved):
LGTM, a bit out of scope, but I feel we could remove some code duplication by addressing the comment I made :)

src/routes/chains/converters.rs
fmrsabino: Yeah good point. Added in 4fce4f5

jpalvarezl (approved):
Looking good! 🚀

fmrsabino:
@gnosis/safe-web currently being deployed (should take couple minutes). It will be available under `GET https://safe-client.staging.gnosisdev.com/v1/chains`

iamacook:
I will add this to the types now.

katspaugh:
Thanks Fred!

A question.
```
"rpcUri": {
  "authentication": "API_KEY_PATH",
  "value": "https://mainnet.infura.io/v3/"
},
```

When it says `API_KEY_PATH` in the RCP URI, does it mean a key needs to be appended to the path?

Why aren't we using the `{{...}}` template syntax for this?

iamacook:
I added a slash between the RPC uri and token in our code and that caused the bug. [I've fixed it now](https://github.com/gnosis/safe-react/blob/f589b19ecd98a3da423014031391c2f539058c01/src/config/index.ts#L69):

```ts
const formatRpcServiceUrl = ({ authentication, value }: RpcUri, TOKEN): string => {
  const usesInfuraRPC = authentication === RPC_AUTHENTICATION.API_KEY_PATH
  return usesInfuraRPC && TOKEN ? `${value}${TOKEN}` : value
}

export const getRpcServiceUrl = (): string => {
  const { rpcUri } = getChainInfo()
  return formatRpcServiceUrl(rpcUri, INFURA_TOKEN)
}

export const getSafeAppsRpcServiceUrl = (): string => {
  const { safeAppsRpcUri } = getChainInfo()
  return formatRpcServiceUrl(safeAppsRpcUri, SAFE_APPS_RPC_TOKEN)
}
```

The `{{...}}` template syntax was a bit confusing to deal with. [I had to remove the search params and them manually add them myself](https://github.com/gnosis/safe-react/blob/f589b19ecd98a3da423014031391c2f539058c01/src/config/index.ts#L146):

```ts
const [apiUrl] = getExplorerUriTemplate().api.split('?')

const params: Record<string, string> = {
  module: 'contract',
  action: 'getAbi',
  address: contractAddress,
  ...(apiKey && { apiKey }),
}

const response = await fetch(`${apiUrl}?${new URLSearchParams(params)}`)
```

fmrsabino:
> Thanks Fred!
> 
> A question.
> 
> ```
> "rpcUri": {
>   "authentication": "API_KEY_PATH",
>   "value": "https://mainnet.infura.io/v3/"
> },
> ```
> 
> When it says `API_KEY_PATH` in the RCP URI, does it mean a key needs to be appended to the path?

Yes

> Why aren't we using the `{{...}}` template syntax for this?

The discussion regarding this feature happened here: https://github.com/gnosis/safe-config-service/issues/171

> _Edit:_ In our code, we indeed just append the key at the end:
> 
> ```
> export const getSafeAppsRpcServiceUrl = (): string =>
>   useInfuraRPC() ? `${getConfig().safeAppsRpcServiceUrl}/${SAFE_APPS_RPC_TOKEN}` : getConfig().safeAppsRpcServiceUrl
> 
> export const getRpcServiceUrl = (): string =>
>   useInfuraRPC() ? `${getConfig().rpcServiceUrl}/${INFURA_TOKEN}` : getConfig().rpcServiceUrl
> ```
> 
> And this already caused a bug because we didn't have a trailing slash before, and with the remote config there were two slashes. So we should probably switch to the template like the other URL values.

Just appending `str` might skip some url validations that some components do (most of the times present in the standard library). Do you have a similar component? Either way we can always add/remove a trailing slash from our side as long as the clients are aligned on this.

fmrsabino:
> I added a slash between the RPC uri and token in our code and that caused the bug. [I've fixed it now](https://github.com/gnosis/safe-react/blob/f589b19ecd98a3da423014031391c2f539058c01/src/config/index.ts#L69):

Our goal is to have this as easy to use for the clients to consume the payload. If you need to have a trailing slash we can enforce this via the admin.

> The `{{...}}` template syntax was a bit confusing to deal with. [I had to remove the search params and them manually add them myself](https://github.com/gnosis/safe-react/blob/f589b19ecd98a3da423014031391c2f539058c01/src/config/index.ts#L146):

The `{{key}}` pattern was an idea to have something that doesn't occur often in the url so that you can do a `Find + Replace`. Do you have a suggestion for a pattern that works well of you? (mobile is already using this though)

katspaugh:
I find nothing confusing about the {{...}} syntax.

iamacook:
I predominantly used the legacy code. I'll implement your suggestion.

katspaugh:
Thanks! :)

PR #724, opened 2021-12-02
Sort new /balances by fiat balance

### LINKED ISSUE #722: Balances endpoint response is no longer sorted fiat value
Related safe-react issue: https://github.com/gnosis/safe-react/issues/3083

We should sort by fiat value

fmrsabino:
Closes #722
- The returned balances should be sorted by `fiat_balance`


src/routes/balances/handlers_v2.rs
```diff
@@ -52,7 +55,7 @@ pub async fn balances(
     let token_prices: Vec<TokenPrice> =
         get_token_prices(context, &info_provider, &backend_balances).await;
 
-    let service_balances: Vec<Balance> = backend_balances
+    let mut service_balances: Vec<Balance> = backend_balances
```
jpalvarezl, line 58: Is it necessary for this to be mutable? I would expect `total_fiat` requiring to be `mut`

fmrsabino: It is required because of the sorting of the vector.

jpalvarezl (approved):

PR #726, opened 2021-12-03
Included `currentNonce` in `estimation` endpoint respnse

### LINKED ISSUE #711: Return current safe contract nonce as part of the estimation
Currently, estimation returns `latest_nonce` and `safe_tx_gas` fields (https://gnosis.github.io/safe-client-gateway/docs/routes/safes/models/struct.SafeTransactionEstimation.html)

It would be helpful to get the contract `nonce` field as part of the estimation. It can improve UX when editing transaction parameters. We can display an explicit error if a user tries to specify a nonce that is already used.

Related EPIC: https://github.com/gnosis/safe/issues/503

jpalvarezl:
Closes #711

### DIFF
--- a/src/routes/safes/tests/routes.rs
+++ b/src/routes/safes/tests/routes.rs
@@ -8,14 +8,12 @@ use crate::routes::safes::models::{
 };
 use crate::tests::main::setup_rocket;
 use crate::utils::errors::{ApiError, ErrorDetails};
-// use crate::utils::errors::{ApiError, ErrorDetails};
 use crate::utils::http_client::{MockHttpClient, Request, Response};
 use core::time::Duration;
 use mockall::predicate::eq;
 use rocket::http::{ContentType, Header, Status};
 use rocket::local::asynchronous::Client;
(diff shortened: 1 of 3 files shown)


src/routes/safes/tests/routes.rs
```diff
@@ -350,7 +348,7 @@ async fn get_owners_not_found() {
 
 #[rocket::async_test]
 async fn post_safe_gas_estimation() {
-    let safe_address = "0xD9BA894E0097f8cC2BBc9D24D308b98e36dc6D02";
```
fmrsabino, line 353: Out of curiosity, why did you change the safe-address in the test? 😅

fmrsabino (approved):

src/routes/safes/tests/routes.rs
jpalvarezl: Ah just to emphasise that the addres in the Request URL didn't have to match the `to` field in the body. Also now the `safe_address` is an actual test safe :D

PR #727, opened 2021-12-06
Use imports for links for rustdoc

fmrsabino:
- Use imports to provide link support in `rustdoc`

### DIFF
--- a/src/routes/about/routes.rs
+++ b/src/routes/about/routes.rs
@@ -1,15 +1,19 @@
+use rocket::response::content;
+
 use crate::cache::cache_operations::CacheResponse;
 use crate::config::{about_cache_duration, webhook_token};
 use crate::providers::info::{DefaultInfoProvider, InfoProvider};
 use crate::routes::about::handlers;
+use crate::routes::about::models::About;
+use crate::routes::about::models::ChainAbout;
+use crate::routes::safes::models::Implementation;
 use crate::utils::context::RequestContext;
 use crate::utils::errors::ApiResult;
 use crate::utils::http_client::Request;
-use rocket::response::content;
 
 /**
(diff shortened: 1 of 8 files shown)


jpalvarezl (commented):
Very and nice thorough work 🙌 !

I have only one leftover request. The tests are showing all the unused imports as warnings, which makes a bit harder to capture more helpful warnings in the future. 

To tackle this, I suggest to add the following `#[cfg_attr(test, allow(unused_imports))]` in every `routes/<app_name>/mod.rs` on top of the `pub mod routes;` line. For `transactions`, as an example: https://github.com/gnosis/safe-client-gateway/blob/2e9c65f9282d859935f8d86e776a24c0129e3c6d/src/routes/transactions/mod.rs#L6

This would only ignore warnings related to imports for test builds. `cargo run` would still show unused imports in these modules (I've double checked)

fmrsabino:
> Very and nice thorough work 🙌 !
> 
> I have only one leftover request. The tests are showing all the unused imports as warnings, which makes a bit harder to capture more helpful warnings in the future.
> 
> To tackle this, I suggest to add the following `#[cfg_attr(test, allow(unused_imports))]` in every `routes/<app_name>/mod.rs` on top of the `pub mod routes;` line. For `transactions`, as an example:
> 
> https://github.com/gnosis/safe-client-gateway/blob/2e9c65f9282d859935f8d86e776a24c0129e3c6d/src/routes/transactions/mod.rs#L6
> 
> This would only ignore warnings related to imports for test builds. `cargo run` would still show unused imports in these modules (I've double checked)

After discussing this issue, we noticed that these warnings were also part of the production build which is not intended. Therefore we decided to not continue with this and, instead, treat broken links as errors. See https://github.com/gnosis/safe-client-gateway/pull/729

PR #729, opened 2021-12-07
Set broken_intra_doc_links linter to level deny

fmrsabino:
- Set `broken_intra_doc_links` to level `deny` – see https://doc.rust-lang.org/beta/rustdoc/lints.html#broken_intra_doc_links
- This makes the build return an error if there's a broken link in the docs

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,4 +1,5 @@
 #![deny(unused_must_use)]
+#![deny(rustdoc::broken_intra_doc_links)]
 
 extern crate dotenv;
 extern crate log;


src/main.rs
```diff
@@ -1,4 +1,5 @@
 #![deny(unused_must_use)]
+#![deny(rustdoc::broken_intra_doc_links)]
```
jpalvarezl, line 2: this can probably be added to the previous `deny` with a comma

jpalvarezl (approved):
LGTM! 🙌

src/main.rs
fmrsabino: I think I actually prefer on separate lines. Makes it easier for the diff what do you think?

PR #736, opened 2021-12-09
`txData::to` request full address search for contract/token info

### LINKED ISSUE #732: Provide name and logo for the transaction details txData.to field
When showing advanced transaction details, we'd like to show the `txData.to` with the known address name and logo, if it's available.

Currently it only returns the address value.

[Example Response](https://safe-client.gnosis.io/v1/chains/1/transactions/multisig_0x84443F61efc60D10DA9F9a2398980CD5748394BB_0x0f234d2881036da130be836d4a1721a2f9846d6589ace5bbcb1ed7cbadd7c1fc)

[Current implementation](https://github.com/gnosis/safe-client-gateway/blob/main/src/routes/transactions/converters/details.rs#L31)

Linked issue: https://github.com/gnosis/safe-ios/issues/1571

jpalvarezl:
Closes #732

### DIFF
--- a/src/routes/transactions/converters/details.rs
+++ b/src/routes/transactions/converters/details.rs
@@ -28,7 +28,10 @@ impl MultisigTransaction {
             tx_status: self.map_status(&safe_info),
             tx_info: self.transaction_info(info_provider).await,
             tx_data: Some(TransactionData {
-                to: AddressEx::address_only(&self.safe_transaction.to),
+                to: info_provider
+                    .address_ex_from_any_source(&self.safe_transaction.to)
+                    .await
+                    .unwrap_or(AddressEx::address_only(&self.safe_transaction.to)),
                 value: self.safe_transaction.value.to_owned(),
                 hex_data: self.safe_transaction.data.to_owned(),
                 data_decoded: self.safe_transaction.data_decoded.clone(),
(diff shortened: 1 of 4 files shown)


src/routes/transactions/converters/details.rs
```diff
@@ -28,7 +28,10 @@ impl MultisigTransaction {
             tx_status: self.map_status(&safe_info),
             tx_info: self.transaction_info(info_provider).await,
             tx_data: Some(TransactionData {
-                to: AddressEx::address_only(&self.safe_transaction.to),
+                to: info_provider
```
rmeissner, line 31: Do you think it makes sense to have an helper or flag in the method for this? (default on failure)

rmeissner (approved):

src/routes/transactions/converters/details.rs
jpalvarezl: I thought about the helper method. as in `AddressEx::any_source` but I would need to pass in the `info_provider` anyway, which is not that bad, jsut thought it would be a bit too roundabout. 

Btw, I noticed we already return this info in `txInfo.to` 🤔 it appears we are duplicating this here.

jpalvarezl: The clients have a single screen showing `txData` therefore redundancy makes sense.

PR #748, opened 2021-12-15
Atempt http error deserialization

### LINKED ISSUE #745: Error deserialization works from cache but not direct forward
Reproduce as mentioned in the meeting. 

Look into:

https://github.com/gnosis/safe-client-gateway/blob/main/src/utils/errors.rs

and https://github.com/gnosis/safe-client-gateway/blob/a2593880f28713c78a670f6ab9d06fed784bfebc/src/cache/cache_op_executors.rs#L66

jpalvarezl:
Closes #745 

The issue was that we were storing in the cache the error json in the body of the request to the backend services, while forwarding the error by just passing the same body as a flat string. 

This change introduces the following:
- We attempt to deserialize into the error structure we know from the core services (`ErrorDetails`)
- if the error contains a structure we don't know, we revert to the old behaviour
- I added a tests to cover both branches

### DIFF
--- a/src/utils/errors.rs
+++ b/src/utils/errors.rs
@@ -85,7 +85,7 @@ impl ApiError {
         )
     }
 
-    fn new(status_code: u16, message: ErrorDetails) -> Self {
+    pub fn new(status_code: u16, message: ErrorDetails) -> Self {
         Self {
             status: status_code,
             details: message,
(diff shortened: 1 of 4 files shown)


src/utils/errors.rs
```diff
@@ -85,7 +85,7 @@ impl ApiError {
         )
     }
 
-    fn new(status_code: u16, message: ErrorDetails) -> Self {
+    pub fn new(status_code: u16, message: ErrorDetails) -> Self {
```
rmeissner, line 88: This is because of tests?

rmeissner (approved):

src/utils/errors.rs
jpalvarezl: No, is because of this line: https://github.com/gnosis/safe-client-gateway/blob/48138f5c7f3ef757747fb0fb6dfac1ab6164dacb/src/cache/cache_op_executors.rs#L69

PR #759, opened 2021-12-20
Closes #757: Fix wrong `latestNonce`

### LINKED ISSUE #757: Inconsistent logic in propose endpoint
**Describe the bug**
The propose endpoint uses an inconsistent logic to determine `latestNonce`. 

If there are transactions with a nonce higher or equal to the current Safe nonce, it returns this nonce, which then indicates to the clients that they should use `latestNonce + 1` for the next transaction. But if there is no such transaction the endpoint returns the current Safe nonce, therefore the `+ 1` would skip one nonce that should be used. Another issue is that it defaults to `0` in case that no known tx is available.

**Expected behavior**
A nonce should be returned that can be immediately used without further adjustments.

**Additional context**
Fallback to current nonce was introduced with #754

**Proposed solution**
- Revert #754 
- Add v2 estimate endpoint with  `recommendedNonce` field instead of `latestNonce` that returns the nonce that should be used for the next transaction

rmeissner:
Closes #757 

- Revert #754 
- Add `v2` estimation endpoint that replaces `latestNonce` with `recommendedNonce`
 
Example:
Endpoint: `POST - /v2/chains/<chain_id>/safes/<safe_address>/multisig-transactions/estimations`

Example request body:
  ```json
  {
    "to": "0xD9BA894E0097f8cC2BBc9D24D308b98e36dc6D02",
    "value": "0",
    "data": "0x095ea7b3000000000000000000000000ae9844f89d98c150f5e61bfc676d68b4921559900000000000000000000000000000000000000000000000000001c6bf52634000",
    "operation": 0
  }
 ```

Assuming the current nonce of the Safe is `7` and the transaction with the highest nonce stored on the service is `76` the following response will be returned:
 ```json
  {
    "currentNonce": 7,
    "recommendedNonce": 77,
    "safeTxGas": "63417"
  }
 ```

`recommendedNonce` can be directly used for proposing a new Safe transaction without further adjustments.

### DIFF
--- a/src/routes/safes/handlers/estimations.rs
+++ b/src/routes/safes/handlers/estimations.rs
@@ -31,9 +31,7 @@ pub async fn estimate_safe_tx_gas(
     )?;
     let current_nonce = info_provider.safe_info(safe_address).await?.nonce;
 
-    let last_known_nonce =
-        fetch_latest_nonce(context.http_client(), latest_multisig_tx_url).await?;
-    let latest_nonce = max(current_nonce, last_known_nonce);
+    let latest_nonce = fetch_latest_nonce(context.http_client(), latest_multisig_tx_url).await?;
     let safe_tx_gas = fetch_estimation(
         context.http_client(),
         estimation_url,
--- a/src/routes/safes/handlers/estimations.rs
+++ b/src/routes/safes/handlers/estimations.rs
@@ -6,32 +6,37 @@ use crate::common::models::backend::transactions::{
 };
 use crate::common::models::page::Page;
 use crate::providers::info::{DefaultInfoProvider, InfoProvider};
-use crate::routes::safes::models::{SafeTransactionEstimation, SafeTransactionEstimationRequest};
+use crate::routes::safes::models::{
+    SafeTransactionEstimation, SafeTransactionEstimationRequest, SafeTransactionEstimationV2,
+};
 use crate::utils::context::RequestContext;
 use crate::utils::errors::ApiResult;
 use crate::utils::http_client::{HttpClient, Request};
 use std::cmp::max;
 
-pub async fn estimate_safe_tx_gas(
+async fn get_last_known_nonce_and_estimate(
     context: &RequestContext,
     chain_id: &str,
     safe_address: &str,
     safe_transaction_estimation_request: &SafeTransactionEstimationRequest,
-) -> ApiResult<SafeTransactionEstimation> {
+) -> ApiResult<(String, u64, Option<u64>)> {
     let info_provider = DefaultInfoProvider::new(chain_id, &context);
-    let estimation_url = core_uri!(
-        info_provider,
-        "/v1/safes/{}/multisig-transactions/estimations/",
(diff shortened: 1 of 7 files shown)


src/routes/safes/tests/routes.rs
```diff
@@ -553,7 +554,7 @@ async fn post_safe_gas_estimation_no_queued_tx() {
 
     let expected = serde_json::from_value(json!( {
         "currentNonce": 7,
-        "latestNonce": 7,
+        "latestNonce": 0,
```
rmeissner, line 557: This change is the result of reverting #754

src/routes/safes/tests/routes.rs
```diff
@@ -670,7 +671,7 @@ async fn post_safe_gas_estimation_delayed_indexing() {
 
     let expected = serde_json::from_value(json!( {
         "currentNonce": 7,
-        "latestNonce": 7,
+        "latestNonce": 6,
```
rmeissner, line 674: This change is the result of reverting #754

src/routes/safes/handlers/estimations.rs
```diff
@@ -6,48 +6,101 @@ use crate::common::models::backend::transactions::{
...
 use std::cmp::max;
 
-pub async fn estimate_safe_tx_gas(
+async fn get_last_known_nonce_and_estimate(
     context: &RequestContext,
     chain_id: &str,
     safe_address: &str,
     safe_transaction_estimation_request: &SafeTransactionEstimationRequest,
-) -> ApiResult<SafeTransactionEstimation> {
+) -> ApiResult<(String, u64, Option<u64>)> {
```
Uxio0, line 22: Why not using `SafeTransactionEstimationV2` here?

Uxio0 (approved):

src/routes/safes/handlers/estimations.rs
```diff
@@ -6,48 +6,101 @@ use crate::common::models::backend::transactions::{
...
     let info_provider = DefaultInfoProvider::new(chain_id, &context);
-    let estimation_url = core_uri!(
-        info_provider,
-        "/v1/safes/{}/multisig-transactions/estimations/",
-        safe_address
-    )?;
+
+    let current_nonce = info_provider.safe_info(safe_address).await?.nonce;
+
     let latest_multisig_tx_url = core_uri!(
```
Uxio0, line 27: Shouldn't this be inside a provider?

Uxio0 (approved):

src/routes/safes/handlers/estimations.rs
rmeissner: The provider returns the whole info object, but we only need the nonce here

rmeissner: Maybe, I followed the pattern from the existing code

src/routes/safes/handlers/estimations.rs
Uxio0: 🔝

PR #764, opened 2021-12-23
Update README.md

fmrsabino:
- Update test instructions in the `README.md`
- In general, the section should be more compact now and will not provide information on the options that `cargo` offers. For this, the `cargo` documentation provides a more up to date description.
- `REDIS_URI` is the only environment variable that needs to be set for the tests to execute currently.

`README.md` preview – https://github.com/gnosis/safe-client-gateway/blob/update-tests-instructions/README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,18 +42,22 @@ The contents of the file should be the following (see `.env.sample` for an examp
 
 ## Tests
 
-To run all tests use the `cargo test` command. If you want to run a specific subset of tests, then add additionally any info regarding the path of the tests and `cargo` will match it.
+1. In order to run the tests of the project you need to have an instance of Redis running (as some of them test the integration with Redis).
 
-Example: `cargo test converters` will run every tests under the `converters` module. Matching occurs also at a test name level, so by writing the full name of a test, that single test can be run.
-
-As you can see in the `./github/workflows/rust.yml` there are certain basic requirements to run tests:
-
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -42,13 +42,15 @@ The contents of the file should be the following (see `.env.sample` for an examp
 
 ## Tests
 
-1. In order to run the tests of the project you need to have an instance of Redis running (as some of them test the integration with Redis).
+In order to run the test suite of the project:
+
+1. Have an instance of Redis running (as some of them test the integration with Redis).
 
 ```bash
 redis-server
```
Uxio0, line 50: Shouldn't we encourage to use docker, as we do with the other projects?

Uxio0: (I know you are just refactoring a little the document)

fmrsabino: Yes that is a good point. We have `docker-compose` also on the project so it shouldn't require a lot of configuration.

rmeissner (approved):

README.md
fmrsabino: I just checked and `docker-compose` is assigning a random port to `redis` because one is not set by default. I will update it to map the same port – that will make the instructions more clear.

fmrsabino: Opened it here: https://github.com/gnosis/safe-client-gateway/pull/765

Uxio0 (approved):

PR #769, opened 2022-01-03
Pagination fix for `chains` endpoint

### LINKED ISSUE #762: Chains pagination uses safe-config urls
**Describe the bug**
- When requesting the pagination endpoint `/v1/chains` on the `safe-client-gateway`, if a next or previous page is returned, it returns the internal `safe-config-service` url instead of one that has the `safe-client-gateway` as the host.

**To Reproduce**
- Trigger pagination – eg.: by using a limit of 1 chain: https://safe-client.gnosis.io/v1/chains?limit=1
- The `next` field is `http://safe-config-service-web.safe.svc.cluster.local/api/v1/chains/?limit=1&offset=1`

**Expected behavior**
- The `next` field is `https://safe-client.gnosis.io/v1/chains?limit=1&offset=1`

 - Staging or production? – Both.
 - Which chain? – Non applicable.
 - OS: [e.g. iOS] – Non applicable.
 - Browser [e.g. chrome, safari] – Non applicable
 - Version [e.g. 22] –  v3.13.3

jpalvarezl:
Closes #762 

- We bundle `limit` and `offset` in `cursor` to generate the `next` and `previous` links for `Page`s of data (like for tx related endpoints)
- Reused as much code as possible from `txHistory` and `txQueued`
- Tested `chain` endpoints

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -54,6 +54,7 @@ jobs:
       WEBHOOK_TOKEN: test_webhook_token
       CONFIG_SERVICE_URI: https://config.service.url
       VPC_TRANSACTION_SERVICE_URI: 'false'
+      SCHEME: http
 
     steps:
       - name: Checkout branch
--- a/src/routes/chains/handlers.rs
+++ b/src/routes/chains/handlers.rs
@@ -43,7 +43,7 @@ pub async fn get_single_chain(
 }
 
 fn map_link(context: &RequestContext, original_link: Option<String>) -> Option<String> {
-    let a = original_link.as_ref().map(|link| {
+    original_link.as_ref().map(|link| {
         let cursor =
             PageMetadata::from_cursor(link.split("?").collect::<Vec<&str>>().get(1).unwrap_or(&""))
                 .to_url_string();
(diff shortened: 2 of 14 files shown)


src/routes/chains/handlers.rs
```diff
@@ -33,3 +41,16 @@ pub async fn get_single_chain(
     let info_provider = DefaultInfoProvider::new(&chain_id, &context);
     Ok(info_provider.chain_info().await?.into())
 }
+
+fn map_link(context: &RequestContext, original_link: Option<String>) -> Option<String> {
+    original_link.as_ref().map(|link| {
+        let cursor =
+            PageMetadata::from_cursor(link.split("?").collect::<Vec<&str>>().get(1).unwrap_or(&""))
+                .to_url_string();
```
fmrsabino, line 49: nitpick: For this we can also rely on `Url::parse` and then use `query()` to return the query parameter part (suggesting this because parsing URLs can be complex and this might provide extra validations)

.github/workflows/rust.yml
```diff
@@ -54,6 +54,7 @@ jobs:
       WEBHOOK_TOKEN: test_webhook_token
       CONFIG_SERVICE_URI: https://config.service.url
       VPC_TRANSACTION_SERVICE_URI: 'false'
+      SCHEME: http
```
fmrsabino, line 57: Is this required?

fmrsabino (approved):

PR #771, opened 2022-01-04
Added `txQueued` endpoint specific cache duration

### LINKED ISSUE #770: Introduce separate cache timeout for `txQueued` endpoint
Introduce `TX_QUEUED_CACHE_DURATION` defaulting to `REQUEST_CACHE_DURATION` for `RequestCached` in the `txQueued` endpoint

jpalvarezl:
Closes #770


src/routes/transactions/handlers/queued.rs
```diff
@@ -42,6 +42,7 @@ pub async fn get_queued_transactions(
 
     let body = RequestCached::new_from_context(url, context)
         .request_timeout(transaction_request_timeout())
+        .cache_duration(tx_queued_cache_duration())
```
rmeissner, line 45: This is the wrong request. We need to adjust https://github.com/gnosis/safe-client-gateway/blob/7d79372cc95424f18af0ef136c96d1dea8e4d275/src/routes/transactions/routes.rs#L173

rmeissner (changes requested):

src/routes/transactions/handlers/queued.rs
jpalvarezl: Ok will set the cache time out for the response we cache then.

PR #773, opened 2022-01-05
Added `TRANSACTION_SERVICE_AUTH_TOKEN` and the possibility to add headers to requests

### LINKED ISSUE #772: Add token for authentication between CGW<> Core services comms
`/owners` endpoint in the core services is expensive. Its access needs to be restricted. We need to be able to add an auth token in the form of a header.

Link to the core services issue: https://github.com/gnosis/safe-transaction-service/pull/611

- `CORE_SERVICE_AUTH_TOKEN`  needs to be defined. Akin to the `WEBHOOK_TOKEN` (generated with openssl rand -base64 32)
- Implement support for adding request headers in `RequestCached`

jpalvarezl:
Closes #772 

Note: ~~`TRANSACTION_SERVICE_AUTH_TOKEN` is now **expected** . Meaning that if not present, the CGW would panic, failing the request.~~

The env var has been made optional as it is not an integral part of the setup of a CGW instance. However, I left the original error mesasge as a `WARN` level log, to notify developers.

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -17,7 +17,7 @@ ROCKET_ADDRESS=127.0.0.1
 # ROCKET_PORT=8000
 # ROCKET_ADDRESS=localhost
 WEBHOOK_TOKEN=some_random_token
-CORE_SERVICE_AUTH_TOKEN=some_other_random_token
+TRANSACTION_SERVICE_AUTH_TOKEN=some_other_random_token
 # Rocket logs are noise-y, this value filters the logs for errors and our perf monitor
 # Set to "debug" when developing
 # You can select which proportion of the time logs are emited with LOG_THRESHOLD values range [0.0, 1.0]
--- a/.env.sample
+++ b/.env.sample
@@ -17,6 +17,7 @@ ROCKET_ADDRESS=127.0.0.1
 # ROCKET_PORT=8000
 # ROCKET_ADDRESS=localhost
 WEBHOOK_TOKEN=some_random_token
+CORE_SERVICE_AUTH_TOKEN=some_other_random_token
 # Rocket logs are noise-y, this value filters the logs for errors and our perf monitor
 # Set to "debug" when developing
 # You can select which proportion of the time logs are emited with LOG_THRESHOLD values range [0.0, 1.0]
--- a/src/cache/cache_operations.rs
+++ b/src/cache/cache_operations.rs
@@ -13,6 +13,7 @@ use rocket::futures::FutureExt;
 use rocket::response::content;
 use serde::Deserialize;
 use serde::Serialize;
+use std::collections::HashMap;
 use std::future::Future;
 use std::sync::Arc;
 
@@ -155,6 +156,7 @@ pub struct RequestCached {
     pub cache_duration: usize,
     pub error_cache_duration: usize,
     pub cache_all_errors: bool,
+    pub headers: HashMap<String, String>,
 }
 
 impl RequestCached {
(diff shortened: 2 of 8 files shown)


.env.sample
```diff
@@ -17,6 +17,7 @@ ROCKET_ADDRESS=127.0.0.1
 # ROCKET_PORT=8000
 # ROCKET_ADDRESS=localhost
 WEBHOOK_TOKEN=some_random_token
+CORE_SERVICE_AUTH_TOKEN=some_other_random_token
```
fmrsabino, line 20: I'd change this to `TRANSACTION_SERVICE_AUTH_TOKEN` – I think this makes it more explicit for people onboarding to the infrastructure projects on what this is.

src/cache/cache_operations.rs
```diff
@@ -179,6 +182,7 @@ impl RequestCached {
             cache_duration: request_cache_duration(),
             error_cache_duration: request_error_cache_duration(),
             cache_all_errors: false,
+            headers: HashMap::new(),
```
fmrsabino, line 185: nitpick: Not sure if `Default::default()` is more idiomatic in this scenario.

src/config/mod.rs
```diff
@@ -28,6 +28,14 @@ pub fn webhook_token() -> String {
     env::var("WEBHOOK_TOKEN").expect("WEBHOOK_TOKEN missing in env")
 }
 
+pub fn transaction_service_auth_token() -> String {
+    let token = env::var("TRANSACTION_SERVICE_AUTH_TOKEN").unwrap_or_else(|_| {
+        log::warn!("TRANSACTION_SERVICE_AUTH_TOKEN missing in env");
+        String::new()
+    });
+    format!("Token {}", token)
```
fmrsabino, line 36: I think we can move this to the `add_header` just to make it more explicit that the `Token` is actually only required for the header and not part of the token itself wdyt?

jpalvarezl: My rationale here was that `Authorization: Token token` gets handled as key value `("Authorization", "Token token")` therefore, it is always expected like that. Looking at this [link](https://developer.mozilla.org/en-US/docs/Web/HTTP/Authentication#authentication_schemes) I thought that maybe in some cases (in the  future) we would have `Bearer` or `Basic` and it would always be the same.
I am not sure though, I am totally open to suggestions. I would maybe move this `format!` to the `handler.rs` file 🤔 . Wdyt?

src/cache/cache_operations.rs
jpalvarezl: Scratch that 😅 it's totally possible, I agree I would prefer that too 👍

src/config/mod.rs
fmrsabino: > My rationale here was that `Authorization: Token token` gets handled as key value `("Authorization", "Token token")` therefore, it is always expected like that. Looking at this [link](https://developer.mozilla.org/en-US/docs/Web/HTTP/Authentication#authentication_schemes) I thought that maybe in some cases (in the future) we would have `Bearer` or `Basic` and it would always be the same.
> I am not sure though, I am totally open to suggestions. I would maybe move this `format!` to the `handler.rs` file 🤔 . Wdyt?

I was actually just thinking that the token by itself is not directly tied to the header in the `safe-transaction-service` even though it is the only use case so far. I was wondering if the authorisation token was used somewhere else besides the header then the CGW would already have that value in `TRANSACTION_SERVICE_AUTH_TOKEN` and we wouldn't need to strip the `Token ` part away.

This is not an issue. Was just something that I was considering and we can do this at a later point if needed.

fmrsabino (approved):

PR #774, opened 2022-01-06
Feature: Add `client_url` param to safe apps list endpoint and include accessControl in the response

mikheevm:
Relevant ticket: https://github.com/gnosis/safe-config-service/issues/351
**This PR:**
- Add `client_url` param to `safe-apps` endpoint
- Add `accessControl` property to mocked safe apps in tests


mikheevm:

src/common/models/backend/safe_apps.rs
```diff
@@ -18,3 +19,24 @@ pub struct SafeAppProvider {
     pub url: String,
     pub name: String,
 }
+
+#[derive(Deserialize, Debug, PartialEq, Clone)]
+#[serde(tag = "type")]
+pub enum SafeAppAccessControlPolicies {
+    #[serde(rename(deserialize = "NO_RESTRICTIONS"))]
+    NoRestrictions(SafeAppNoRestrictionsPolicy),
+    #[serde(rename(deserialize = "DOMAIN_ALLOWLIST"))]
```
jpalvarezl, line 28: Did you try using `#[serde(rename_all = "SCREAMING_SNAKE_CASE")]` at the `enum` level? Or you want these `type`s have different value when you serialise them?

src/routes/safe_apps/models.rs
```diff
@@ -20,3 +21,27 @@ pub struct SafeAppProvider {
...
+#[serde(rename_all = "camelCase")]
+#[cfg_attr(test, derive(serde::Deserialize))]
+pub struct SafeAppDomainAllowlistPolicy {
+    pub value: Vec<String>,
+}
+
+#[derive(Serialize, Debug, PartialEq, Clone)]
+#[serde(rename_all = "camelCase")]
+#[cfg_attr(test, derive(serde::Deserialize))]
+pub struct SafeAppNoRestrictionsPolicy {}
```
jpalvarezl, line 47: You don't need `{}`, just `replace it with `;` and you should be good :) Then you won't need it in the `converter` method either

src/routes/safe_apps/handlers.rs
```diff
@@ -5,9 +5,16 @@ use crate::routes::safe_apps::models::SafeApp;
...
-
+pub async fn safe_apps(
+    context: &RequestContext,
+    chain_id: &String,
+    client_url: &Option<String>,
+) -> ApiResult<Vec<SafeApp>> {
+    let url = config_uri!(
+        "/v1/safe-apps/?chainId={}&clientUrl={}",
+        chain_id,
+        client_url.clone().unwrap_or("".to_string())
```
jpalvarezl, line 16: if you get rid of `to_string` do you still need to `clone`?

jpalvarezl:
Do you have a ticket in different repo or here in the cgw repo where I can check the spec for this? It would allow me to find issues (should there be any) with the business logic.

mikheevm:
> Do you have a ticket in different repo or here in the cgw repo where I can check the spec for this? It would allow me to find issues (should there be any) with the business logic.

Updated the description, the ticket is https://github.com/gnosis/safe-config-service/issues/351

src/common/models/backend/safe_apps.rs
mikheevm: > Or you want these types have different value when you serialise them?

If I understood you right, this is exactly what I want. I followed the pattern in https://github.com/gnosis/safe-client-gateway/blob/a36739ba10752ab34ded695a0de87df53c86495a/src/common/models/backend/transfers.rs#L8

src/routes/safe_apps/handlers.rs
mikheevm: If I remove `to_string` then I get this error: https://stackoverflow.com/questions/32500974/expected-string-found-str-when-matching-an-optional-string

src/common/models/backend/safe_apps.rs
jpalvarezl: Ok so you want camelCased on serializaed and you expect them to be upper case when you deserialise them

mikheevm: no, I expect them to be uppercase (or screaming snake case) when serialized too

jpalvarezl: In that case you would need to implement according to my suggestion: 
```rust
#[serde(rename_all = "SCREAMING_SNAKE_CASE")]
pub enum SafeAppAccessControlPolicies {
   NoRestrictions(SafeAppNoRestrictionsPolicy),
   DomainAllowList(SafeAppDomainAllowlistPolicy)
}
```

jpalvarezl: Just one thing, there are cleaner examples of polymorphism in the CGW, the one you used (`Transfers`) is quite old code and verbose. 

A cleaner newer example you can find here: https://github.com/gnosis/safe-client-gateway/blob/8cbaca58b181e63dcba1a22fb8c6b1d54c3a47db/src/common/models/backend/chains.rs#L45

After reviewing the original ticket in the config service, it seems like you don't even need `SafeAppNoRestrictionsPolicy` defined.

jpalvarezl:
> how do I enable the nightly channel?

I think that's ok, your PR passed the `rustfmt` check so you are good either way :)

src/common/models/backend/safe_apps.rs
```diff
@@ -10,6 +10,7 @@ pub struct SafeApp {
     pub description: String,
     pub chain_ids: Vec<u64>,
     pub provider: Option<SafeAppProvider>,
+    pub access_control: SafeAppAccessControlPolicies,
```
jpalvarezl, line 13: According to https://github.com/gnosis/safe-config-service/issues/351 this fields is technically `nullable` , right?

This should then be of type `Option<SafeAppAccessControlPolicies>`

mikheevm: No, it is not. It's either of these two values

src/routes/safe_apps/handlers.rs
jpalvarezl: I think here maybe doing `client_url.as_deref().unwrap_or("")` is a bit better, as this will both remove your `&` from the `Option` and return you `&str` after the unwarp.

src/routes/safe_apps/models.rs
```diff
@@ -11,6 +11,7 @@ pub struct SafeApp {
     pub description: String,
     pub chain_ids: Vec<String>,
     pub provider: Option<SafeAppProvider>,
+    pub access_control: SafeAppAccessControlPolicies,
```
jpalvarezl, line 14: This may have to be `Option<SafeAppAccessControlPolicies>` too if the value from the backend is also expected to be nullable. In which case I would suggest adding `#[serde(skip_serializing_if = "Option::is_none")]
` to the field.

src/common/models/backend/safe_apps.rs
jpalvarezl: Cool, I will just give a heads-up to the team internally, as we would then panic if we try to deploy this to production the CGW before this is available in the prod instance of the config service too :)

src/routes/safe_apps/handlers.rs
mikheevm: Thanks, I've just updated. I'm curious what's the difference?

jpalvarezl: because your input is `&Option<String>` you were forced to call unwrap with `"".to_string()`. [as_deref](https://doc.rust-lang.org/std/option/enum.Option.html#method.as_deref) will change our type from that to `Option<&String>` and since, `&String` and `&str` are more or less interchangably used, then your `unwrap("")` no longer needs to call `to_string` as `""` is of type `&str`

You can read more about it here:
- https://users.rust-lang.org/t/whats-the-difference-between-string-and-str/10177/2

jpalvarezl (approved):

jpalvarezl:
Looking good :) Thanks @mikheevm 🙌 !

mikheevm:
thanks for the review @jpalvarezl 🙌

PR #777, opened 2022-01-14
Update README.md

fmrsabino:
- Add `TRANSACTION_SERVICE_AUTH_TOKEN` as a required environment variable that needs to be set
- The value of `TRANSACTION_SERVICE_AUTH_TOKEN` needs to be `some_other_random_token` else the tests fail as It won't be able to match with the mocked network request

I think we should just match to whatever is set in the variable and not with the string `some_other_random_token`. Setting the token to any non-empty value is considered correct and the tests should also pass that way.


Uxio0 (commented):
Wouldn't this be a little confusing? We are not using the auth token anywhere (yet)

fmrsabino:
> Wouldn't this be a little confusing? We are not using the auth token anywhere (yet)

Yes I agree. This is given the current state of the CGW which requires this though. Thus my comment on the PR:

> I think we should just match to whatever is set in the variable and not with the string some_other_random_token. Setting the token to any non-empty value is considered correct and the tests should also pass that way.

Makes sense to remove this from the test for now. But in its current state anyone who follows the instructions is going to have these tests failing.

Uxio0:
Then tests should set their own random token. We cannot enforce an option that's not used anywhere

README.md
```diff
@@ -50,10 +50,11 @@ In order to run the test suite of the project:
 redis-server
 ```
 
-2. Make sure that `REDIS_URI` is set and points to the current Redis instance (assuming Redis is runnning on the default port `6379`):
+2. Make sure that the required environment variables are set (the following example assumes that Redis is runnning on the default port `6379`):
 
 ```bash
 export REDIS_URI=redis://localhost:6379
+export TRANSACTION_SERVICE_AUTH_TOKEN=some_other_random_token
```
jpalvarezl, line 57: Maybe it is confusing that we add `TRANSACTION_SERVICE_AUTH_TOKEN` env var to the list of required env vars , while we don't really `panic` if it's not present: 
https://github.com/gnosis/safe-client-gateway/blob/a24617bc68cdbe1af978e81d0d649b8f03be799e/src/config/mod.rs#L36

fmrsabino: This is just for the test setup since the tests require this specific token.

jpalvarezl (approved):

PR #781, opened 2022-01-25
Nested `DELEGATE` internal transactions in `multiSend` flagged as "not trusted"

### LINKED ISSUE #725: `trusted` flag for `DELEGATE_CALL` transactions
## Feature

Propage `trusted` flag for delegate calls to contracts returned by the core services (see [comment](https://github.com/gnosis/safe-client-gateway/issues/725#issuecomment-989806242) for full spec)

>  - `trustedDelegateCallTarget : Option<bool>` if it's a `CALL` tx we don't send it. If it's a `DELEGATE` it is either true or false, according to what is returned by the `contracts` endpoint from the core services.
> - Nested `DELEGATE` in `multiSend`: Always warn about them
> - The flag can be found `TransactionDetails::TransactionData::trustedDelegateCallTarget` (`TransactionData` is nullable, however it is never null for `Module` and `Multisig` Transactions)

jpalvarezl:
Closes #725 

Added logic and unit tests for `DELEGATE` calls in internal `multiSend` transactions


src/common/converters/data_decoded.rs
```diff
@@ -120,6 +123,27 @@ impl DataDecoded {
             Some(index)
         }
     }
+
+    pub fn has_nested_delegated(&self) -> bool {
+        if self.method == MULTI_SEND {
+            if let Some(value_decoded_type) =
```
rmeissner, line 129: If was thinking if we should be more defensive and check for all parameters (on the root) if they have a `valueDecoded` which corresponds to `InternalTransaction`. We should then make sure that this is only checked when `trusted_for_delegate_call == true`. .... But maybe I am too paranoid :grimacing:

rmeissner: And also remove the check on the `method` ... as we should probably check this for all methods ... again not sure if I am too paranoid

rmeissner: I don't think this is a blocker right now, maybe something to play around with on a ecosystem friday.

rmeissner (approved):

mikheevm:
Does it only check the decoded data? What if the internal transaction has an unverified contract and no decoding?

PR #783, opened 2022-01-26
Generalised nested `DELEGATE` check for any method

### LINKED ISSUE #782: Generalise check for nested `DELEGATE` internal transaction for `DataDecoded`
We need to set the flag `isTrustedDelegateCall` to false in the event of a `DELEGATE` in internal transactions occuring in the `data` of a transaction.

This is currently only checked for `multiSend` calls, but we should generalise this check and apply it to every single paramter.

Only in the case where a call to the `/contracts` endpoint returns `true`, we further this check by looking into `DataDecoded` looking for any parameter that corresponds to an `InternalTransaction` and we check for `Operation == DELEGATE` (not only in the `multiSend` case). We should also **not** look into parameter names. 

**Additional context**
Where this check is performed in the `multiSend` case:
https://github.com/gnosis/safe-client-gateway/blob/e43b2175d653cca6ae835c2459668f544ac3b4c3/src/common/converters/data_decoded.rs#L127

jpalvarezl:
Closes #782 

Changes:
- Not only `multiSend` is checked
- We are only checking a single level of decoding
- Check is short-circuited with the `isTrustedDelegate` flag from the backend (only when the former is `true` we check further)


src/common/converters/tests/data_decoded.rs
```diff
@@ -899,3 +900,11 @@ fn nested_delegate_in_multi_send_without_nested_delegate() {
 
     assert_eq!(data_decoded.has_nested_delegated(), false);
 }
+
+#[test]
+fn no_param_data_decoded_has_nested_delegate() {
+    let data_decoded =
+        serde_json::from_value::<DataDecoded>(json!({"method":"parameterlessMethod"})).unwrap();
```
rmeissner, line 907: Do we have a test for a call with nested delegate calls that is not a multisend?

jpalvarezl: not really 🤔 I will add one 👍

jpalvarezl: I am not able to find a suitable example for this. Generally, `execTransaction` comes to mind but I can't seem to find an example that contains `InternalTransaction`s. 

Currently there aren't any checks on the name of the method. I assume you want this test to assert that we don't check only `multiSend`, right ?

rmeissner (approved):

rmeissner (approved):

PR #796, opened 2022-02-09
Add safe address to transaction details

fmrsabino:
- Add `safeAddress` to `TransactionDetails` payload
- `safeAddress` is non-nullable


Uxio0 (approved):
This is the kind of issue that brings more work on refactoring tests than on the issue 😉

PR #799, opened 2022-02-14
Map contracts endpoint error to ApiResult

### LINKED ISSUE #798: `is_trusted_delegate_call` returns wrong result
**Describe the bug**
When the `operation` is `1` (for delegatecall) then `trusted_delegate_call_target` is expected to be present, but this is not the case.

**To Reproduce**
- Call url `https://safe-client.gnosis.io/v1/chains/1/transactions/0x526d5558f0b29145a874226ac8ecb6a0b17d12b4016ced108dd9f5a8363d4497`
- `trusted_delegate_call_target` is not present in the `tx_data`

**Expected behavior**
- `trusted_delegate_call_target` should be available (and have the value of `true` in this case)

 - production
 - Mainnet
 - 3.16.0

**Additional context**
image

Currently the core service errors when requesting the token info: https://safe-transaction.mainnet.gnosis.io/api/v1/contracts/0x40A2aCCbd92BCA938b02010E17A5b8929b49130D/

fmrsabino:
Closes #798 

- `is_trusted_delegate_call` now returns an `ApiResult` – if the call to the `/contracts` endpoint fails, that failure is propagated (thus allowing callers to decide what to do with the error)
- Previously `is_trusted_delegate_call` was returning `None` if the API call to `/contracts` failed

### DIFF
--- a/src/routes/transactions/converters/details.rs
+++ b/src/routes/transactions/converters/details.rs
@@ -29,7 +29,8 @@ impl MultisigTransaction {
             &self.safe_transaction.data_decoded,
             info_provider,
         )
-        .await;
+        .await?;
+
         Ok(TransactionDetails {
             safe_address: self.safe_transaction.safe.to_owned(),
             tx_id: self.generate_id(),
@@ -141,7 +142,8 @@ impl ModuleTransaction {
             &safe_transaction.data_decoded,
             info_provider,
         )
-        .await;
+        .await?;
+
         Ok(TransactionDetails {
             safe_address: self.safe_transaction.safe.to_owned(),
             tx_id: self.generate_id(),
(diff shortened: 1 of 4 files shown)


src/routes/transactions/converters/details.rs
```diff
@@ -1,5 +1,7 @@
 extern crate chrono;
 
+use rocket::futures::future::OptionFuture;
```
rmeissner, line 3: :thinking:

src/routes/transactions/converters/details.rs
```diff
@@ -177,21 +180,18 @@ pub async fn is_trusted_delegate_call(
...
-            })
-            .ok()
+        let contract_info = info_provider.contract_info(to).await?;
+
+        let has_nested_delegate_calls = !data_decoded
+            .as_ref()
+            .map_or(false, |data_decoded| data_decoded.has_nested_delegated());
+
+        let is_trusted_delegate_call =
+            contract_info.trusted_for_delegate_call && has_nested_delegate_calls;
```
jpalvarezl, line 192: The way that it was written before was short-circuiting to prevent the http request in case `contract.trusted_for_delegate_call` was `true`. I don't mind the change, I am just pointing out a big change in behaviour.

jpalvarezl: Never mind this comment

jpalvarezl (approved):

PR #805, opened 2022-02-18
Add /v1/chains/<chain_id>/hooks/events

### LINKED ISSUE #804: Don't use auth tokens on webhook urls
Currently tokens are on the urls, it can lead to leaks and also it's not really standard.

Use the standard HTTP header `Authorization`: https://developer.mozilla.org/en-US/docs/Web/HTTP/Authentication

fmrsabino:
Closes #804

- Add `/v1/chains/<chain_id>/hooks/events` route
- This route does not require a token to be set in the path. However it needs to be set in the header under `Authorization`
- This route validates the `Authorization` header:
  * If the `Authorization` is not set a `400 - Bad Request` is returned
  * else if the `Authorization` is set only only once in the header and it matches the expected value then the request guard is successful
  * else if does not match the expected value a `401 - Unauthorised` is returned

### DIFF
--- a/src/routes/hooks/authorization.rs
+++ b/src/routes/hooks/authorization.rs
@@ -1,0 +1,34 @@
+use rocket::http::Status;
+use rocket::request::{FromRequest, Outcome};
+use rocket::Request;
+
+use crate::config::webhook_token;
+
+pub struct AuthorizationToken {
+    value: String,
+}
+
+#[derive(Debug)]
+pub enum AuthorizationError {
+    Missing,
+    Invalid,
+    BadCount,
+}
+
+#[rocket::async_trait]
+impl<'r> FromRequest<'r> for AuthorizationToken {
+    type Error = AuthorizationError;
+
+    async fn from_request(request: &'r Request<'_>) -> Outcome<Self, Self::Error> {
+        let keys: Vec<&str> = request.headers().get("Authorization").collect();
+
+        match keys.len() {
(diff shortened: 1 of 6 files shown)


Uxio0:
I think we should change the route, including verbs in the url is not very restful:
- `/v2/chains/<chain_id>/hook/update` -> `/v2/chains/<chain_id>/hooks/events/` (or something similar)

Uxio0:
> If the Authorization is not set a 400 - Bad Request is returned
> else if does not match the expected value a 400 - Bad Request is returned

In this cases, we should return `401 Unauthorized`

> else if multiple Authorization headers are provided a 400 - Bad Request is returned
👍🏻

fmrsabino:
> I think we should change the route, including verbs in the url is not very restful:
> 
> * `/v2/chains/<chain_id>/hook/update` -> `/v2/chains/<chain_id>/hooks/events/` (or something similar)

I was thinking on having just `/v2/chains/<chain_id>/hooks`. Wdyt?

fmrsabino:
> > I was thinking on having just `/v2/chains/<chain_id>/hooks`. Wdyt?
> 
> What I understand if I see `POST /v2/chains/<chain_id>/hooks` is that you are creating a new hook, but looks good to me

`POST /v2/chains/<chain_id>/hooks/payload`? 😅

Uxio0:
`POST /v2/chains/<chain_id>/hooks/payloads` 😉

rmeissner (approved):

Uxio0 (approved):

PR #806, opened 2022-02-18
Update rustfmt rules

fmrsabino:
- In order to be more restrictive on the types of changes changed upstream the following rules were enabled:
  * `reorder_impl_items` – Reorder impl items. type and const are put first, then macros and methods.
  * `normalize_comments` – Convert /* */ comments to // comments where possible
  * `imports_granularity` – Merge imports from the same module into a single use statement. Conversely, imports from different modules are split into separate statements.


.github/workflows/rust.yml
```diff
@@ -71,7 +72,6 @@ jobs:
           profile: minimal
           toolchain: ${{ env.TOOLCHAIN }}
           override: true
-          components: rustfmt
```
fmrsabino, line 74: This was removed as we don't need this component for the tests

.github/workflows/rust.yml
```diff
@@ -61,6 +57,7 @@ jobs:
       - name: Checkout branch
         uses: actions/checkout@v2
 
+      # This can be removed once https://github.com/actions-rs/toolchain/pull/209 is merged and released
```
fmrsabino, line 60: Once this is merged we can remove the `toolchain` parameter from `actions-rs/toolchain@v1` as it will read from the project `rust-toolchain.toml` file

.github/workflows/rust.yml
```diff
@@ -12,24 +12,20 @@ jobs:
...
-        run: |
-          echo "TOOLCHAIN=$(rustup show active-toolchain | cut -d " " -f1)" >> $GITHUB_ENV
-
+        
       - name: Install toolchain
         uses: actions-rs/toolchain@v1
         with:
           profile: minimal
-          toolchain: ${{ env.TOOLCHAIN }}
+          toolchain: nightly
```
rmeissner, line 20: Not a fan of this, as we should specify the toolchain as exact as possible, especially with nightly

rustfmt.toml
```diff
@@ -1 +1,11 @@
-ignore = ["src/json"]
```
rmeissner, line 1: Why did we remove this ignore?

.github/workflows/rust.yml
fmrsabino: Yes I agree. That's the `nightly` version is only used for the `rustfmt` the stable channel is used for compiling and the tests.

rustfmt.toml
fmrsabino: Good catch. I had this on my list because we are currently getting this error on `main`:

```
Warning: can't set `ignore = IgnoreList { path_set: {"src/json"}, rustfmt_toml_path: "" }`, unstable features are only available in nightly channel.
```

I will double check if the rule/path is still valid.

.github/workflows/rust.yml
rmeissner: But even for nightly it would be nice to fix it to a specific nightly version

fmrsabino: Yes I agree. Changed here [b297e7b](https://github.com/gnosis/safe-client-gateway/pull/806/commits/b297e7b732582f22a0a84aa936813ad97b8cabfb)

rustfmt.toml
fmrsabino: Ok this was ignoring the files under `src/json` but this folder no longer exists. Was there any `.rs` file under this folder that was supposed to be ignored by `rustfmt`? 🤔

.github/workflows/rust.yml
jpalvarezl: Should we open an issue instead to have better visibility of when we should do this?

rustfmt.toml
jpalvarezl: This was added to ignore `json` files used for tests, although nowadays they are scattered across tests modules. Maybe we should think about a more ad-hoc solution to flagging modules to be ignored by `rustfmt`. Or we may just as well want rustfmt to format those, if it is fact doing that.

.github/workflows/rust.yml
```diff
@@ -13,23 +13,19 @@ jobs:
...
 
-      - name: Set variables
-        run: |
-          echo "TOOLCHAIN=$(rustup show active-toolchain | cut -d " " -f1)" >> $GITHUB_ENV
-
       - name: Install toolchain
         uses: actions-rs/toolchain@v1
         with:
           profile: minimal
-          toolchain: ${{ env.TOOLCHAIN }}
```
jpalvarezl, line 24: I couldn't find any usages of `TOOLCHAIN` outside this script. I was a bit concerned it could be used else such as the `DockerFile` or our deployment script. It seems that it's only used here, but if you can double check in case I missed something, that'd be great!

jpalvarezl (approved):
If the docs are looking good, then I would say this could be merged.

rustfmt.toml
fmrsabino: > Or we may just as well want rustfmt to format those, if it is fact doing that.

I could not find any information regarding `rustfmt` formatting `json` files (I would say that would be a bit weird too 😅 )

.github/workflows/rust.yml
fmrsabino: It's not being used outside of this file but it is also set for the `tests` job further below.

.github/workflows/rust.yml
fmrsabino: I'm a bit wary of creating issues to track on-going tasks of other repos but no big opinion on this one.

PR #811, opened 2022-02-21
Async redis

### LINKED ISSUE #803: Async usage of redis
Change the `crate::cache::Cache` trait to be `async` in every method.

jpalvarezl:
Closes #803 

Changes:
- Introduced `bb8-redis` for handling `async` connection requests to the pool
-  Redis commands now use their `async` counterparts


src/cache/redis.rs
```diff
@@ -1,95 +1,102 @@
...
-            &mut self.conn(),
-            scan_match_count(&mut self.conn(), pattern, redis_scan_count()),
-        );
+    async fn invalidate_pattern(&self, pattern: &str) {
+        let mut con = self.conn().await;
+        let keys_cmd = scan_match_count_cmd(pattern, redis_scan_count());
+        let mut keys = keys_cmd.iter_async(&mut *con).await.unwrap();
+        pipeline_delete(&mut keys)
+            .await
+            .query_async::<_, ()>(&mut *con)
```
rmeissner, line 71: Should we move this into the `pipeline_delete`?

src/cache/redis.rs
```diff
@@ -1,95 +1,102 @@
...
 
-    fn invalidate_pattern(&self, pattern: &str) {
-        pipeline_delete(
-            &mut self.conn(),
-            scan_match_count(&mut self.conn(), pattern, redis_scan_count()),
-        );
+    async fn invalidate_pattern(&self, pattern: &str) {
+        let mut con = self.conn().await;
+        let keys_cmd = scan_match_count_cmd(pattern, redis_scan_count());
+        let mut keys = keys_cmd.iter_async(&mut *con).await.unwrap();
```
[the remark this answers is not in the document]
jpalvarezl: I moved it out because previously we had to provide a mutable reference to the connection to the method. I thought returning an immutable command ready to be executed by the caller which is a assumed to have a connection is a bit cleaner.

I was also running into some typing issues with the connection, as passing the type alias was seemingly losing some trait implementation info. I can investigate further, if you prefer the previous approach where a mutable reference to the connection is passed down to these methods.

src/cache/redis.rs
jpalvarezl: I explained in this [comment](https://github.com/gnosis/safe-client-gateway/pull/811#discussion_r812624179) why I took it out. Just as for that comment, if you prefer the approach where the mutable reference to the connection is passed, I can try investigating further.

src/cache/redis.rs
```diff
@@ -1,95 +1,99 @@
...
+
+type RedisPool = Pool<RedisConnectionManager>;
+type RedisConnection<'a> = PooledConnection<'a, RedisConnectionManager>;
+
+async fn create_pool() -> RedisPool {
+    let manager = RedisConnectionManager::new(redis_uri())
+        .expect("Establishing connection with redis instance failed");
+    bb8::Pool::builder()
+        .max_size(15) // default is 10
+        .build(manager) // we can technically also set a connection await timeout if necessary
```
fmrsabino, line 15: nitpick: the comment here is unrelated with the `.build()` right?

src/cache/redis.rs
rmeissner: I think it is fair for now. I was more thinking about reusability. Currently you would have to repeat more code than was necessary before.

Cargo.toml
```diff
@@ -7,6 +7,7 @@ edition = "2018"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
+bb8-redis = "0.10.1"
```
rmeissner, line 10: bb8 sucks .... go r2d2

image

image

Cargo.toml
```diff
@@ -26,7 +27,7 @@ regex = "1.5.4"
 reqwest = { version = "0.11.3", features = ["json"] }
 rocket = { version = "0.5.0-rc.1", features = ["tls", "json"] }
 rocket_codegen = { version = "0.5.0-rc.1" }
-redis = { version = "0.21", features = ["r2d2"] }
+redis = { version = "0.21", features = ["tokio-comp"] }
```
fmrsabino, line 30: If `bb8-redis` handles the async pool for us do we still need `tokio-comp`?

fmrsabino: Just saw that they use this as a dependency on the `bb8` side:

https://github.com/djc/bb8/blob/227ee229351106b3435c00ca187a9a1a1b14185c/redis/Cargo.toml#L12

So I'd say that if we use `redis` directly then we should keep this (so that the version is pinned). If we don't then we can remove this dependency completely.

src/cache/redis.rs
fmrsabino: No strong opinion on this one. It's more the defaults and the API itself that can change and I think the best way to see it is by opening the docs 😄

fmrsabino (approved):

fmrsabino (approved):
This is a really nice addition 🙌  🚀

PR #812, opened 2022-02-22
Async redis cmds (sync connection pool)

### LINKED ISSUE #803: Async usage of redis

jpalvarezl:
Closes #803 

- Seemingly working with `async` comands
- No need for additional crates, just additional features from `redis-rs`
- We should probably look into using a [ConnectionManager](https://docs.rs/redis/latest/redis/aio/struct.ConnectionManager.html)


jpalvarezl:
Moving in favour of the solution with an async connection pool

PR #815, opened 2022-02-22
Add /v2/flush

fmrsabino:
- Add a new route `/v2/flush` which requires an `Authorization` header to be set (with the value `Basic <WEBHOOK_TOKEN>`)

### DIFF
--- a/src/routes/hooks/routes.rs
+++ b/src/routes/hooks/routes.rs
@@ -56,3 +56,13 @@ pub fn flush(
     Invalidate::new(invalidation_pattern.0, context.cache()).execute();
     Ok(())
 }
+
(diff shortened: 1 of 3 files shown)


fmrsabino:
> Should we also a version of https://github.com/gnosis/safe-client-gateway/blob/main/src/routes/about/routes.rs#L101 that uses the header?

Yes I think that makes sense. I will target that in a different PR

rmeissner (approved):

fmrsabino:
> > Should we also a version of https://github.com/gnosis/safe-client-gateway/blob/main/src/routes/about/routes.rs#L101 that uses the header?
> 
> Yes I think that makes sense. I will target that in a different PR

https://github.com/gnosis/safe-client-gateway/pull/816

Still missing some tests regarding token validation on this endpoint.

PR #817, opened 2022-02-23
Remove unused dotenv_codegen and proc-macro2

fmrsabino:
These dependencies were not being used:
- `dotenv_codegen`(https://crates.io/crates/dotenv_codegen) – this is used for the `dotenv!` macro which we don't use
- `proc-macro2`(https://crates.io/crates/proc-macro2) – we also don't use the macros under this crate


jpalvarezl (approved):
Good clean-up 🙌 !

PR #818, opened 2022-02-23
Remove unused extern crates

fmrsabino:
- Some `extern crate` are no longer required
- Additionally, a more idiomatic way was introduced in Rust 2018 – https://doc.rust-lang.org/edition-guide/rust-2018/path-changes.html#no-more-extern-crate (adding it to `Cargo.toml` should be sufficient in most scenarios)


jpalvarezl (approved):
🚀

PR #820, opened 2022-02-24
Add nginx configuration

fmrsabino:
- Add an Nginx configuration as a template under `nginx/templates/nginx.conf.template`
- Nginx maps the host port `NGINX_HOST_PORT` to port `80`
- Add the following environment variables:
  * `NGINX_HOST_PORT` – port where Nginx is reachable
  * `NGINX_ENVSUBST_OUTPUT_DIR` – container folder where the generated files should be stored with the environment variables replaced
  * `REDIS_HOST` – redis location (default is `redis`)
  * `REDIS_PORT` – redis port (default is `6379`)
- Use production defaults for `.env.sample`
  * `ROCKET_ADDRESS` is now set as `0.0.0.0`


nginx/templates/nginx.conf.template
```diff
@@ -0,0 +1,64 @@
...
+          add_header              Front-End-Https   on;
+          # we don't want nginx trying to do something clever with
+          # redirects, we set the Host: header above already.
+          proxy_redirect off;
+          # They default to 60s. Increase to avoid WORKER TIMEOUT in web container
+          proxy_connect_timeout 60s;
+          proxy_read_timeout 60s;
+    }
+  }
+}
```
fmrsabino, line 64: @luarx this would be the configuration to be used by devops with `${ROCKET_PORT}` replaced with the production one

.env.sample
```diff
@@ -4,18 +4,31 @@ CONFIG_SERVICE_URI=https://safe-config.staging.gnosisdev.com
 # Production
```
Uxio0, line 4: I wouldn't make our staging urls public

Uxio0 (approved):

.env.sample
fmrsabino: I agree. This was already there but it makes sense to remove it.

PR #827, opened 2022-03-10
Add autodeploy github actions step with nginx auth script

gMonty030:
- Removed Dockerhub autodeploy weebhooks
- Added new step for the autodeploy request after the image is built
- Added auth to autodeploy service, now is more secure the deployment of new versions.
- Added nginx auth script

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -187,17 +187,10 @@ jobs:
     if: github.ref == 'refs/heads/main' || github.ref == 'refs/heads/develop' || startsWith(github.ref, 'refs/tags/')
     steps:
     - uses: actions/checkout@v2
-    - name: Deploy Mainnet
+    - name: Deploy staging
       if: github.ref == 'refs/heads/main'
       run: bash scripts/autodeploy.sh
       env:
         AUTODEPLOY_URL: ${{ secrets.AUTODEPLOY_URL }}
         AUTODEPLOY_TOKEN: ${{ secrets.AUTODEPLOY_TOKEN }}
         TARGET_BRANCH: "staging"
-    - name: Deploy Develop
-      if: github.ref == 'refs/heads/develop'
-      run: bash scripts/autodeploy.sh
-      env:
-        AUTODEPLOY_URL: ${{ secrets.AUTODEPLOY_URL }}
-        AUTODEPLOY_TOKEN: ${{ secrets.AUTODEPLOY_TOKEN }}
-        TARGET_BRANCH: "develop"
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -184,10 +184,10 @@ jobs:
   autodeploy:
     runs-on: ubuntu-latest
     needs: [deploy]
+    if: github.event_name == 'push' && github.ref == 'refs/heads/main'
     steps:
     - uses: actions/checkout@v2
     - name: Deploy staging
-      if: github.event_name == 'push' && github.ref == 'refs/heads/main'
       run: bash scripts/autodeploy.sh
       env:
         AUTODEPLOY_URL: ${{ secrets.AUTODEPLOY_URL }}
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -183,7 +183,7 @@ jobs:
 
   autodeploy:
     runs-on: ubuntu-latest
-    needs: [rustfmt, tests, deploy]
+    needs: [deploy]
     if: github.ref == 'refs/heads/main' || github.ref == 'refs/heads/develop' || startsWith(github.ref, 'refs/tags/')
     steps:
     - uses: actions/checkout@v2
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -180,3 +180,24 @@ jobs:
       - name: Deploy develop # unused
         if: github.ref == 'refs/heads/develop'
         run: bash scripts/deploy_docker.sh develop
+
+  autodeploy:
+    runs-on: ubuntu-latest
+    needs: [rustfmt, tests, deploy]
+    if: github.ref == 'refs/heads/main' || github.ref == 'refs/heads/develop' || startsWith(github.ref, 'refs/tags/')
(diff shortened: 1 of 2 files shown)


.github/workflows/rust.yml
```diff
@@ -180,3 +180,24 @@ jobs:
...
         if: github.ref == 'refs/heads/develop'
         run: bash scripts/deploy_docker.sh develop
+
+  autodeploy:
+    runs-on: ubuntu-latest
+    needs: [rustfmt, tests, deploy]
+    if: github.ref == 'refs/heads/main' || github.ref == 'refs/heads/develop' || startsWith(github.ref, 'refs/tags/')
+    steps:
+    - uses: actions/checkout@v2
+    - name: Deploy Mainnet
```
jpalvarezl, line 190: I think that this name is not very descriptive. As merges into the `main` branch represent deployments into `staging`. The CGW is chain agnostic too, only the legacy was chain specific, but then again even back then, merges to `main` represented `staging` deployments.

.github/workflows/rust.yml
```diff
@@ -180,3 +180,24 @@ jobs:
       - name: Deploy develop # unused
         if: github.ref == 'refs/heads/develop'
         run: bash scripts/deploy_docker.sh develop
+
+  autodeploy:
```
jpalvarezl, line 184: We would be missing a deployment for `tags` which are the ones that represent our `production` deployments.

jpalvarezl (commented):
I hope that my comments are not too out of place. It would be helpful maybe to have a meeting or at least an issue to help better contextualise these changes.

.github/workflows/rust.yml
```diff
@@ -180,3 +180,24 @@ jobs:
...
+    steps:
+    - uses: actions/checkout@v2
+    - name: Deploy Mainnet
+      if: github.ref == 'refs/heads/main'
+      run: bash scripts/autodeploy.sh
+      env:
+        AUTODEPLOY_URL: ${{ secrets.AUTODEPLOY_URL }}
+        AUTODEPLOY_TOKEN: ${{ secrets.AUTODEPLOY_TOKEN }}
+        TARGET_BRANCH: "staging"
+    - name: Deploy Develop
```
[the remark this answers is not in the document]
gMonty030: removed

.github/workflows/rust.yml
gMonty030: Your comments are welcome, indeed these could make the pipeline more understandable

.github/workflows/rust.yml
gMonty030: There is no auto-deploy implemented for production, because of that there is no place for tagged deployments. Maybe this could change in the coming future...

.github/workflows/rust.yml
```diff
@@ -180,3 +180,17 @@ jobs:
       - name: Deploy develop # unused
         if: github.ref == 'refs/heads/develop'
         run: bash scripts/deploy_docker.sh develop
+
+  autodeploy:
+    runs-on: ubuntu-latest
+    needs: [deploy]
+    if: github.ref == 'refs/heads/main' || github.ref == 'refs/heads/develop' || startsWith(github.ref, 'refs/tags/')
```
rmeissner, line 187: Why do we use an if for 3 branches here, but only filter for main in the steps below?

scripts/autodeploy.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
```
rmeissner, line 1: Since we use this in a lot of repos, wouldn't it make sense to have this as a github action that we can share everywhere?

.github/workflows/rust.yml
gMonty030: it's just a quick check of the possible branches that could be used as input.
It can be avoided but make sense if the project is gonna has more than those branches in use

scripts/autodeploy.sh
gMonty030: It could be a good improvement, but I have to research a little bit if there are options to implement a shared mechanism that connects all the repos. I'll take a look

gMonty030:
- - -
   ***I have read the CLA Document and I hereby sign the CLA***
   - - -

gMonty030:

gMonty030:

.github/workflows/rust.yml
fmrsabino, line 187: The following should be enough:

```suggestion
    if: github.event_name == 'push' && github.ref == 'refs/heads/main'
```

This project only uses `main` for deployments to `staging`.

.github/workflows/rust.yml
fmrsabino: @jpalvarezl this update is just to trigger a hook (authenticated) to notify that a new `staging` image is available (that's how I understood it)

fmrsabino (changes requested):
Just a small comment on the condition for the action step.

.github/workflows/rust.yml
gMonty030: Updated.

.github/workflows/rust.yml
gMonty030: Exactly, although the conditions could be more than just a new image, other projects use to have a `develop` branch for `dev-` deployments

.github/workflows/rust.yml
```diff
@@ -184,11 +184,10 @@ jobs:
   autodeploy:
     runs-on: ubuntu-latest
     needs: [deploy]
-    if: github.ref == 'refs/heads/main' || github.ref == 'refs/heads/develop' || startsWith(github.ref, 'refs/tags/')
     steps:
     - uses: actions/checkout@v2
     - name: Deploy staging
-      if: github.ref == 'refs/heads/main'
+      if: github.event_name == 'push' && github.ref == 'refs/heads/main'
```
fmrsabino, line 190: I would move this condition to right after `needs` since we don't need to checkout the branch or start this job if the branch is not `main`

fmrsabino (changes requested):

fmrsabino (approved):
🚀

PR #829, opened 2022-03-21
Extracted max size of request to tokens endpoint

### LINKED ISSUE #828: Transaction flagged as `Custom` when it is in fact a `Transfer`
[staging](https://safe-client.staging.gnosisdev.com/v1/chains/1/transactions/0xc3ee6ea5861954635617edc22da7646a142f0be3d2a9735bdba5736b7d6f8d16) and [production](https://safe-client.gnosis.io/v1/chains/1/transactions/0xc3ee6ea5861954635617edc22da7646a142f0be3d2a9735bdba5736b7d6f8d16)

The token in the transaction is beyond our hardcoded limit for warming up our cache with `TokenInfo` objects.

`staging` works because it has considerably less tokens registered when compared to `production`.

jpalvarezl:
Closes #828 

- Introduction of `TOKEN_CACHE_SIZE_COUNT` to limit the count of tokens requested to warm up our cache
- Set the default value to `20,000`
- Adjusted tests

### DIFF
--- a/src/config/tests/mod.rs
+++ b/src/config/tests/mod.rs
@@ -78,6 +78,11 @@ fn build_usize_test_cases() -> Vec<USizeEnvValue> {
             env_key: String::from("TX_QUEUED_CACHE_DURATION"),
             generator: Box::new(super::tx_queued_cache_duration),
         },
+        USizeEnvValue {
+            expected_default: super::token_cache_size_count(),
+            env_key: String::from("TOKEN_CACHE_SIZE_COUNT"),
+            generator: Box::new(super::token_cache_size_count),
+        },
     ]
 }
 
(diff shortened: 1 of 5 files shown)


src/config/tests/mod.rs
```diff
@@ -78,6 +78,11 @@ fn build_usize_test_cases() -> Vec<USizeEnvValue> {
             env_key: String::from("TX_QUEUED_CACHE_DURATION"),
             generator: Box::new(super::tx_queued_cache_duration),
         },
+        USizeEnvValue {
+            expected_default: super::token_cache_size_count(),
```
fmrsabino, line 82: This is testing more the behaviour of `token_cache_size_count` and not the default itself which might not be what we want.

Shouldn't this test be: "Given that `TOKEN_CACHE_SIZE_COUNT` is empty the value returned by the function is `20000`"?

PR #846, opened 2022-04-06
Removed caching of the POST response from submitting a confirmation

### LINKED ISSUE #845: POST endpoints (confirmation, submit tx) drown errors from the TX service
When submitting and invalid confirmation or transaction errors returned by the transaction service are not correctly forwarded to the client.

Provide at least one of the following: 
The offending line, in the case of the `post_confirmation` can be found [here](https://github.com/gnosis/safe-client-gateway/blob/6e3fa0d5770d185bb0cbe27185d6b33641143b2e/src/routes/transactions/handlers/proposal.rs#L31)

We should remove the usage of the `try` operator (`?`) as it prevents responding with the correct error message.

There are no alternatives for this.

**Additional context**
Nothing additional.

jpalvarezl:
Closes #845

### DIFF
--- a/src/routes/transactions/handlers/proposal.rs
+++ b/src/routes/transactions/handlers/proposal.rs
@@ -6,6 +6,7 @@ use crate::utils::errors::ApiResult;
 use crate::utils::http_client::Request;
 use serde_json::json;
 
+// The transaction details are returned at the route level. The same as for transaction submission
 pub async fn submit_confirmation(
     context: &RequestContext,
     chain_id: &str,
--- a/src/routes/transactions/handlers/proposal.rs
+++ b/src/routes/transactions/handlers/proposal.rs
@@ -1,7 +1,5 @@
 use crate::cache::cache_operations::{Invalidate, InvalidationPattern, InvalidationScope};
 use crate::providers::info::{DefaultInfoProvider, InfoProvider};
-use crate::routes::transactions::handlers::details::get_multisig_transaction_details;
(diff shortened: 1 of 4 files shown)


src/routes/transactions/handlers/proposal.rs
```diff
@@ -13,7 +11,7 @@ pub async fn submit_confirmation(
     chain_id: &str,
     safe_tx_hash: &str,
     signature: &str,
-) -> ApiResult<TransactionDetails> {
+) -> ApiResult<()> {
```
rmeissner, line 14: This would be a breaking change. It is required that we keep returning the tx information

rmeissner (approved):

src/routes/transactions/routes.rs
```diff
@@ -66,21 +66,22 @@ pub async fn post_confirmation<'e>(
...
+    proposal::submit_confirmation(
+        &context,
+        &chain_id,
+        &safe_tx_hash,
+        &request.signed_safe_tx_hash,
+    )
+    .await?;
+
+    let tx_details = CacheResponse::new(&context)
+        .resp_generator(|| details::get_transactions_details(&context, &chain_id, &safe_tx_hash))
```
fmrsabino, line 80: Do we want to cache the transaction details here?

Can't this `POST` affect the `TransactionStatus`?

jpalvarezl: It does. We invalidate the cache right after the post request and only then we are able to reach this statement where we return the tx details

PR #857, opened 2022-04-28
Add option to set special redis mainnet instance

fmrsabino:


src/routes/about/routes.rs
```diff
@@ -101,5 +101,9 @@ pub async fn backbone(
 #[doc(hidden)]
 #[get("/about/redis")]
 pub async fn redis(context: RequestContext, _token: AuthorizationToken) -> ApiResult<String> {
-    Ok(context.cache().info().await.unwrap_or(String::new()))
+    Ok(context
+        .cache(UNSPECIFIED_CHAIN)
```
rmeissner, line 105: Potentially we should add the chain to the url here

src/utils/context.rs
```diff
@@ -1,23 +1,32 @@
...
+use crate::ServiceCache;
+
+pub(crate) const UNSPECIFIED_CHAIN: &'static str = "UNSPECIFIED";
 
 pub struct RequestContext {
     pub request_id: String,
     pub host: String,
     http_client: Arc<dyn HttpClient>,
     cache: Arc<dyn Cache>,
+    mainnet_cache: Arc<ServiceCache>,
```
rmeissner, line 17: Why is this a `ServiceCache` and the other is a `dyn Cache`?

src/providers/info.rs
```diff
@@ -208,7 +212,10 @@ impl InfoProvider for DefaultInfoProvider<'_> {
     }
 
     fn cache(&self) -> Arc<dyn Cache> {
```
rmeissner, line 214: Why dies the info provider expose this?

src/providers/info.rs
```diff
@@ -113,6 +116,7 @@ pub struct DefaultInfoProvider<'p> {
     pub chain_id: &'p str,
     client: Arc<dyn HttpClient>,
     cache: Arc<dyn Cache>,
+    mainnet_cache: Arc<dyn Cache>,
```
rmeissner, line 119: I don't think this should be required here. The Info provider always only operates on one chain. The chain id here is only for the purpose of generating the correct cache keys, right?

src/providers/fiat.rs
```diff
@@ -21,10 +21,10 @@ pub struct FiatInfoProvider {
 }
 
 impl FiatInfoProvider {
-    pub fn new(context: &RequestContext) -> Self {
+    pub fn new(context: &RequestContext, chain_id: &str) -> Self {
```
rmeissner, line 24: Hmmm that one should actually always use the same redis instance, right? As the fiat price conversions are always the same independent of the chain.

src/main.rs
```diff
@@ -44,11 +47,13 @@ async fn rocket() -> Rocket<Build> {
 
     let client = setup_http_client();
     let cache = create_service_cache().await;
+    let mainnet_cache = create_service_cache_mainnet().await;
```
rmeissner, line 50: As this is a quick fix I think this is ok. I would have preferred a more generic way. E.g. a pool manager, that has a mapping of chain id to pool (or service cache how we call it here). All the pools would be instantiated when the app starts, e.g. as defined in an env var that contains a json blob with the expected info :P

src/utils/context.rs
fmrsabino: I'll have a look into this one. The state managed by Rocket cannot have the same type twice (eg.: two `dyn Cache`) but for the `RequestContext` it shouldn't be a problem having what you mentioned.

src/providers/info.rs
fmrsabino: Some routes use it. Eg.: https://github.com/safe-global/safe-client-gateway/blob/67afbd9a4aff5c78fe0d5045d667221e4c1a4dd8/src/routes/safes/handlers/safes.rs#L60

src/providers/info.rs
fmrsabino: Good catch. Removed it 👍

src/routes/about/routes.rs
fmrsabino: For an hotfix I would not do it. The amount of changes is already quite broad.

.env.sample
```diff
@@ -72,6 +72,11 @@ CHAIN_INFO_REQUEST_TIMEOUT=15000
 REDIS_HOST=redis
 REDIS_PORT=6379
 REDIS_URI=redis://${REDIS_HOST}:${REDIS_PORT}
+
+REDIS_HOST_MAINNET=redis
+REDIS_PORT_MAINNET=6379
+REDIS_URI_MAINNET=redis://${REDIS_HOST}:${REDIS_PORT}
```
luarx, line 78: REDIS_URI_MAINNET=redis://${REDIS_HOST_MAINNET}:${REDIS_PORT_MAINNET}?

fmrsabino: Right! Fixed [1805d4c](https://github.com/safe-global/safe-client-gateway/pull/857/commits/1805d4c76f55db6b600c95ed7a10c4d5a30e5cb9)

Uxio0 (approved):

Uxio0:
Looks good as a hotfix

rmeissner (approved):

PR #871, opened 2022-05-11
Pin patch versions of serde and serde_json

fmrsabino:


Uxio0 (approved):
Why is this? Are we having version issues?

fmrsabino:
> Why is this? Are we having version issues?

Some other libraries depend on `serde` and the respective versions were only being updated on the `.lock` file. (eg.: https://github.com/safe-global/safe-client-gateway/pull/868)

Since we use `serde` directly in the project I think it's better to pin them to a specific version (including the patch part).

Uxio0:
> > Why is this? Are we having version issues?
> 
> Some other libraries depend on `serde` and the respective versions were only being updated on the `.lock` file. (eg.: #868)
> 
> Since we use `serde` directly in the project I think it's better to pin them to a specific version (including the patch part).

Makes sense, thanks for explaining!

PR #873, opened 2022-05-12
Use config service to get safe app info

### LINKED ISSUE #855: Try to fetch Safe App information from config service first
Currently the client gateway only looks for the manifest.json of a Safe app based on the Safe app url provided in the `origin` field.

We should first check the config service if it has information about an origin and then fallback to other sources (e.g. manifest)

We should remove this information from the config service completely and only rely on the manifest in all places (e.g. Safe Apps list endpoint)

**Additional context**
- https://github.com/safe-global/safe-client-gateway/blob/main/src/providers/info.rs#L152-L168
- https://app.gitbook.com/o/-MhyDtGxUyODu7d4sgda/s/-MhyEZtd5TVytPJtyS7v-2910905616/~/diff/~/changes/2sfEFsdfUrv0lFHiCLOO/build/sdks/safe-apps/releasing-your-safe-app
- https://gnosisinc.slack.com/archives/C01GL8X8L3H/p1650546491443809

fmrsabino:
Closes https://github.com/safe-global/safe-client-gateway/issues/855

- Use the Safe Config Service to get information about a specific app using the `url` filter
- If the filter doesn't provide any results, or the result is an error, fallback to get information from the manifest

### DIFF
--- a/src/providers/info.rs
+++ b/src/providers/info.rs
@@ -17,6 +17,7 @@ use crate::cache::manager::ChainCache;
 use crate::cache::Cache;
 use crate::common::models::addresses::AddressEx;
 use crate::common::models::backend::chains::ChainInfo;
+use crate::common::models::backend::safe_apps::SafeApp;
 use crate::common::models::backend::safes::MasterCopy;
 use crate::common::models::page::Page;
 use crate::config::{
@@ -71,6 +72,16 @@ pub struct SafeAppInfo {
     pub name: String,
     pub url: String,
     pub logo_uri: String,
+}
+
+impl From<&SafeApp> for SafeAppInfo {
+    fn from(safe_app: &SafeApp) -> Self {
+        SafeAppInfo {
+            name: safe_app.name.clone(),
+            url: safe_app.url.clone(),
+            logo_uri: safe_app.icon_url.clone(),
+        }
+    }
 }
 
 #[derive(Deserialize, Debug, PartialEq)]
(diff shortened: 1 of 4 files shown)


src/providers/info.rs
```diff
@@ -345,6 +347,50 @@ impl DefaultInfoProvider<'_> {
...
+        .await?;
+
+    let config_safe_apps: Vec<SafeApp> = serde_json::from_str::<Vec<SafeApp>>(&result)?;
+
+    match config_safe_apps.first() {
+        None => {
+            bail!("No Safe Apps match the url")
+        }
+        Some(first) => Ok(SafeAppInfo::from(first)),
+    }
```
fmrsabino, line 368: One issue with this is that if for some reason the query parameter is done in a wrong way, all safe-apps are returned since this is just a filter on the Safe Apps collection.

According to our model, multiple apps can also have the same url. If we expect a single URL to match a single Safe App then we can assert that we get a single result back (but we will have the same issue if the total collection of safe apps with no filters is 1)

Uxio0: This works for know. What was the reasoning of letting multiple apps have the same url?

Uxio0 (dismissed):

src/providers/info.rs
fmrsabino: > This works for know. What was the reasoning of letting multiple apps have the same url?

I don't think there's a strong reason for it (as in it was never discussed). But we can align on this and have the `url` as a unique field (probably wouldn't work with `http://example.com` vs `http://example.com/` but we already have that issue when querying)

Uxio0: I faced that problem some time ago... it was very tricky, at last I used something like https://pypi.org/project/url-normalize/

PR #891, opened 2022-05-25
Fix links for cargo doc

fmrsabino:
- https://gnosis.github.io/safe-client-gateway is no longer valid.
- It should be under https://safe.global/safe-client-gateway instead (the action needs to generate the docs with the correct link)


moisses89 (approved):
Looks good.

PR #902, opened 2022-06-08
Add tests for delegates endpoints 

### LINKED ISSUE #705: Unit test coverage for `delegates`

moisses89:
Related to #705 
- test get delegates 
- test post delegates
- test delete delegates


src/routes/delegates/models.rs
```diff
@@ -12,25 +13,25 @@ pub struct Delegate {
...
-    safe: Option<String>,
-    delegate: String,
-    delegator: String,
-    signature: String,
-    label: String,
+    pub safe: Option<String>,
+    pub delegate: String,
+    pub delegator: String,
+    pub signature: String,
+    pub label: String,
```
fmrsabino, line 20: We should not change the field visibility just for the tests. I will double check how this module is structured because the test should have access to the private fields.

src/routes/delegates/tests/routes.rs
```diff
@@ -0,0 +1,283 @@
...
+
+    delegates_request.body(Some(serde_json::to_string(
+        &DelegateCreate{
+            delegate: String::from("0x4CA998cE947Aed03c340141a5491Df539ff1Fd05"),
+            delegator: String::from("0xe6450667b7E9C19845751183f93bc97B01fBAec0"),
+            signature: String::from("0x7c3b61f015633198494c4f6153272e390785a2d1f5c661ac7fa7e53c434cf67d019a7778eb317ab4fb7c4c4cbec0dfa6130094f680da8458d849f58a4a412d291b"),
+            label: String::from("test_delegate"),
+            safe: None
+            }).unwrap())
+        );
```
fmrsabino, line 109: So this test is "installed" on the module that exposes `routes` publicly. Therefore if you remove `pub` from the fields of the module this test won't have access to it.

My suggestion is since we are testing from the route perspective to compare with the expected Json here instead of creating the class itself.

src/routes/delegates/tests/routes.rs
```diff
@@ -0,0 +1,263 @@
...
+            })
+        });
+    // Mock transaction service delegates
+    let mut delegates_request = Request::new(format!(
+        "https://safe-transaction.rinkeby.staging.gnosisdev.com/api/v1/delegates/"
+    ));
+    let mut body_string = String::from(super::BACKEND_CREATE_DELEGATE);
+    body_string = body_string.replace('\n', "");
+    body_string = body_string.replace(' ', "");
+    delegates_request.body(Some(body_string));
```
fmrsabino, line 104: As a reference for the future, we need to come up with a more generic way to compare json on this level.

src/routes/delegates/handlers.rs
```diff
@@ -46,7 +48,7 @@ pub async fn post_delegate(
         request.body(serde_json::to_string(&safe_delegate_create).ok());
         request
     };
-
+    println!("{:?}", request);
```
fmrsabino, line 51: We won't need this 😉

src/routes/delegates/tests/routes.rs
```diff
@@ -0,0 +1,263 @@
...
+    )
+    .await
+    .expect("valid rocket instance");
+    // Create post request
+    let request = client
+        .post(format!("/v1/chains/4/delegates"))
+        .body(String::from(super::BACKEND_CREATE_DELEGATE))
+        .header(Header::new("Host", "test.gnosis.io/api"))
+        .header(ContentType::JSON);
+    println!("{:?}", request);
```
fmrsabino, line 132: This should be removed

fmrsabino (changes requested):

src/routes/delegates/handlers.rs
```diff
@@ -46,7 +48,6 @@ pub async fn post_delegate(
         request.body(serde_json::to_string(&safe_delegate_create).ok());
         request
     };
-
     context.http_client().post(request).await?;
     Ok(())
 }
```
fmrsabino, line 53: We should discard the changes to this file:
- the `Debug` import is not required
- Removing the new newline should also not be part of this PR

fmrsabino (changes requested):

src/routes/delegates/tests/routes.rs
moisses89: I agree It's not an elegant way to compare.

fmrsabino (approved):
🚀 🚀 🚀

PR #906, opened 2022-06-21
Add swagger definition

### LINKED ISSUE #465: Add swagger definition
Currently the client gateway only provides a technical documentation for the endpoints. It would be helpful to have a OpenAPI definition file.

### Uses
- Generate typescript: https://github.com/drwpow/openapi-typescript

### References
- https://github.com/GREsau/okapi

moisses89:
Related with  #465
- add swagger description except the following endpoints: 
``` 
        contracts::routes::post_data_decoder,
        notifications::routes::post_notification_registration,
        safes::routes::post_safe_gas_estimation,
        safes::routes::post_safe_gas_estimation_v2,
        transactions::routes::post_confirmation,
        transactions::routes::post_transaction
``` 
Reason here: https://github.com/safe-global/safe-client-gateway/issues/465#issuecomment-1161605678


src/routes/mod.rs
```diff
@@ -32,46 +32,51 @@ pub mod transactions;
 
 #[doc(hidden)]
 pub fn active_routes() -> Vec<Route> {
-    routes![
+    let no_openapi = routes![
         root,
+        contracts::routes::post_data_decoder,
```
rmeissner, line 37: Why do the post roots don't have an openapi?

src/utils/context.rs
```diff
@@ -80,3 +82,13 @@ impl<'r> FromRequest<'r> for RequestContext {
         });
     }
 }
+
+impl<'a> OpenApiFromRequest<'a> for RequestContext {
```
rmeissner, line 86: Could you add a comment what this is doing?

src/utils/errors.rs
```diff
@@ -15,14 +20,14 @@ use thiserror::Error;
 
 pub type ApiResult<T, E = ApiError> = Result<T, E>;
 
-#[derive(Error, Debug, PartialEq)]
+#[derive(Error, Debug, PartialEq, schemars::JsonSchema)]
```
rmeissner, line 23: Why do we reference the `JsonSchema` here differently then in all other places? (`schemars::JsonSchema` vs `JsonSchema`)

src/utils/errors.rs
```diff
@@ -93,7 +98,18 @@ impl ApiError {
         }
     }
 }
+impl OpenApiResponderInner for ApiError {
```
rmeissner, line 101: What is this for?

src/routes/health/routes.rs
```diff
@@ -3,7 +3,9 @@ use crate::cache::manager::ChainCache;
 use crate::utils::context::RequestContext;
 use crate::utils::errors::ApiResult;
 use rocket::response::content;
+use rocket_okapi::openapi;
 
+#[openapi(tag = "Health")]
```
fmrsabino, line 8: I'd exclude this from Swagger. It's used mostly by other services to check if the service is answering and not relevant for the clients

src/routes/mod.rs
fmrsabino: I think  the question is more about why are `POST` routes excluded from Swagger. Or is really about the `/` endpoint @rmeissner ?

From what I could gather routes that specify a lifetime are not supported by `rocket_okapi`.

fmrsabino: Can we add a comment here linking to the issue about the lifetimes? I would have a group for `root` one for the endpoints which have lifetimes and another one for the ones that we do no want to expose

src/routes/mod.rs
```diff
@@ -32,46 +32,51 @@ pub mod transactions;
...
         transactions::routes::get_module_transactions,
         transactions::routes::get_multisig_transactions,
-        transactions::routes::post_transaction,
-        transactions::routes::post_confirmation,
-        hooks::routes::update,
-        hooks::routes::post_hook_update,
-        hooks::routes::post_hooks_events,
-        hooks::routes::post_flush_events,
-        hooks::routes::flush,
         health::routes::health
```
fmrsabino, line 77: I would not have the `/health` endpoint in Swagger

src/utils/context.rs
fmrsabino: I agree with RIchard here. The comment is a little bit confusing.

From my understanding on how this framework works:
- For the routes that you want to have swagger enabled you need to have `#[openapi]` macro
- For those same routes, the framework needs to know how to represent custom types that implement `FromRequest` (in this case,` RequestContext`)
- By setting mapping  `RequestContext` to `RequestHeaderInput::None` we tell the framework that no user input is required for the header generation (thus it won't show in the swagger API)

src/utils/errors.rs
fmrsabino: I just checked and this trait needs to be implemented for any type returned on the route level.

This does that for the error (i still need to check how is it generating for any `T` in a successful case).

fmrsabino (commented):
My concern with this framework is about compatibility with Rocket versions.

Cargo.toml
```diff
@@ -25,6 +25,7 @@ regex = "1.5.6"
 reqwest = { version = "0.11.10", features = ["json"] }
 rocket = { version = "0.5.0-rc.2", features = ["tls", "json"] }
 rocket_codegen = { version = "0.5.0-rc.2" }
+rocket_okapi = { version="0.8.0-rc.2", features = ["swagger", "rapidoc"] }
```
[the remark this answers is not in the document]
moisses89: yes rapidoc is not necessary.

src/routes/mod.rs
moisses89: You mean in code? Maybe something like this could be enough:
```
        root,
        // rocket_okapi don't support lifetimes
        // https://github.com/GREsau/okapi/issues/84
        contracts::routes::post_data_decoder,
        notifications::routes::post_notification_registration,
        safes::routes::post_safe_gas_estimation,
        safes::routes::post_safe_gas_estimation_v2,
        transactions::routes::post_confirmation,
        transactions::routes::post_transaction,
        // This endpoints shouldn't be exposed on swagger
        about::routes::redis,
        hooks::routes::update,
        hooks::routes::post_hook_update,
        hooks::routes::post_hooks_events,
        hooks::routes::post_flush_events,
        hooks::routes::flush
```

fmrsabino: I meant exactly this. No need to split in the codebase 👍

src/routes/health/routes.rs
fmrsabino: No strong opinion on this one. I see this more as a tool for client side developers but I see your point. Let's leave it 👍

fmrsabino (changes requested):

fmrsabino (dismissed):
Awesome contribution 👏 

One thing that we should keep in mind is how the development of this dependency goes alongside `rocket`. Rocket should be a priority and we should not be blocked by this dependency.

However I'd say that for now this is a non-issue and in a future iteration we can have the swagger as an opt-in feature instead of being a default

src/utils/context.rs
fmrsabino: @moisses89 I think it's better to improve the documentation for this implementation for future reference. What do you think?

moisses89:
> Awesome contribution clap
> 
> One thing that we should keep in mind is how the development of this dependency goes alongside `rocket`. Rocket should be a priority and we should not be blocked by this dependency.
> 
> However I'd say that for now this is a non-issue and in a future iteration we can have the swagger as an opt-in feature instead of being a default

I agree, rocket first :rocket:

fmrsabino (approved):
🚀 🚀

PR #909, opened 2022-06-22
Set decimals for fiatBalance and fiatConversion to 4

### LINKED ISSUE #907: Fiat conversion value with a lot of decimals
**Describe the bug**
The response of request balance include fiatConversion with a lot of decimals and it don't make sense. 
``` 
"fiatConversion": "114.44905823699306444597536025744655245445077974048765880750923682372662602685335837549405990976862218778940838433661"
``` 

**To Reproduce**
- Call url `https://safe-client.gnosis.io/v1/chains/1/safes/0x80D63b12aecF8aE5884cBF1d3536bb0C5f612CfC/balances/EUR?trusted=true&exclude_spam=true`
- The body includes fiatConversion for each asset with a lot of decimals.

**Expected behavior**
A normal quantity of decimals.
`"fiatConversion":"121.2609"`

 - Production and staging
 - Which chain: mainnet
 - Browser: Postman and firefox

**Additional context**
Safe transaction service returns fiatConversion correctly.
`https://safe-transaction.mainnet.gnosis.io/api/v1/safes/0x80D63b12aecF8aE5884cBF1d3536bb0C5f612CfC/balances/usd/?trusted=true&exclude_spam=true`
`"fiatConversion":"121.2609"`

moisses89:
Closes #907
- Set 4 decimals for fiatBalance and fiatConversion
- Modify test to check it

### DIFF
--- a/src/routes/balances/converters.rs
+++ b/src/routes/balances/converters.rs
@@ -3,6 +3,8 @@ use crate::common::models::backend::chains::NativeCurrency;
 use crate::providers::info::{TokenInfo, TokenType};
 use crate::routes::balances::models::Balance;
 
+const SCALE: f64 = 1_0000_f64;
+
 impl BalanceDto {
     pub fn to_balance(&self, usd_to_fiat: f64, native_coin: &NativeCurrency) -> Balance {
         let fiat_conversion = self.fiat_conversion.parse::<f64>().unwrap_or(0.0) * usd_to_fiat;
@@ -43,8 +45,8 @@ impl BalanceDto {
                 logo_uri,
             },
             balance: self.balance.to_owned(),
-            fiat_balance: ((fiat_balance * 10000.0).round() / 10000.0).to_string(),
-            fiat_conversion: ((fiat_conversion * 10000.0).round() / 10000.0).to_string(),
+            fiat_balance: ((fiat_balance * SCALE).floor() / SCALE).to_string(),
+            fiat_conversion: ((fiat_conversion * SCALE).floor() / SCALE).to_string(),
         }
     }
 }
--- a/src/routes/balances/converters.rs
+++ b/src/routes/balances/converters.rs
@@ -43,8 +43,8 @@ impl BalanceDto {
                 logo_uri,
             },
             balance: self.balance.to_owned(),
-            fiat_balance: fiat_balance.to_string(),
-            fiat_conversion: fiat_conversion.to_string(),
+            fiat_balance: ((fiat_balance * 10000.0).round() / 10000.0).to_string(),
+            fiat_conversion: ((fiat_conversion * 10000.0).round() / 10000.0).to_string(),
         }
     }
 }
--- a/src/common/converters/tests/balances.rs
+++ b/src/common/converters/tests/balances.rs
@@ -19,11 +19,11 @@ fn native_token_balance() {
             logo_uri: Some("https://test.token.image.url".to_string()),
         },
         balance: "7457594371050000001".to_string(),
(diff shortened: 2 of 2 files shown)


src/routes/balances/converters.rs
```diff
@@ -43,8 +43,8 @@ impl BalanceDto {
                 logo_uri,
             },
             balance: self.balance.to_owned(),
-            fiat_balance: fiat_balance.to_string(),
-            fiat_conversion: fiat_conversion.to_string(),
+            fiat_balance: ((fiat_balance * 10000.0).round() / 10000.0).to_string(),
+            fiat_conversion: ((fiat_conversion * 10000.0).round() / 10000.0).to_string(),
```
fmrsabino, line 47: This solution should work but has one issue regarding the value that its representing: for fiat balances I would try to not use `round()` as it can round up and thus it'll show a higher value then it actually is.

In this case it's better to use `floor()`.

I also suggest extracting the rounding scale to a private variable in the file like

```rust
const SCALE: f64 = 1_0000_f64;

impl BalanceDto {
  ...
```

So that we know upfront which scale we are using.

fmrsabino (changes requested):
Don't forget to add a test to this one. A unit test that covers just this function should be enough :)

src/routes/balances/converters.rs
moisses89: Ok agree. There is another way to set the number of decimals to 4 and avoid the division, that is using format. 
The issue for me is that if for example we have to round 1.1, format is returning 1.1000.
This result was breaking a test that expect a result with two decimals. What do you think is preferable use the division or we can accept the trailing zeroes and avoid the division?

fmrsabino: > Ok agree. There is another way to set the number of decimals to 4 and avoid the division, that is using format.
> The issue for me is that if for example we have to round 1.1, format is returning 1.1000.

Have you tried using `format!("{:.4}", number_to_be_formatted)`? This should format the number to use 4 decimal places but I think it still causes rounding (up or down) and we want to use `floor` (if you find a way around this let me know 😄 )

> This result was breaking a test that expect a result with two decimals.

But the unit tests were successful for this change 🤔 Do you remember which test broke? It should have been executed.

> What do you think is preferable use the division or we can accept the trailing zeroes and avoid the division?

Since the end result should be a `String` my preference goes to `format!()` but only if it "executes" `floor` and not `round`.

The division is fine 👍

moisses89: > But the unit tests were successful for this change thinking Do you remember which test broke? It should have been executed.
``` 
---- common::converters::tests::balances::native_token_balance stdout ----
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `Balance { token_info: TokenInfo { token_type: NativeToken, address: "0x0000000000000000000000000000000000000000", decimals: 18, symbol: "ETH", name: "Ether", logo_uri: Some("https://test.token.image.url") }, balance: "7457594371050000001", fiat_balance: "2523.7991", fiat_conversion: "338.4200" }`,
 right: `Balance { token_info: TokenInfo { token_type: NativeToken, address: "0x0000000000000000000000000000000000000000", decimals: 18, symbol: "ETH", name: "Ether", logo_uri: Some("https://test.token.image.url") }, balance: "7457594371050000001", fiat_balance: "2523.7991", fiat_conversion: "338.42" }`', src/common/converters/tests/balances.rs:35:5
``` 
Fails because format! is adding zeroes until fill the number of decimals **fiat_conversion: "338.4200" !=  fiat_conversion: "338.42"**

fmrsabino (approved):
🚀

PR #910, opened 2022-06-23
Fix MultisigFilters query param name

### LINKED ISSUE #904: [TBA] Fix data parameter
TBA

See https://github.com/safe-global/safe-react/pull/3870#issuecomment-1133630396

> I've checked the [swagger](https://safe-transaction.rinkeby.staging.gnosisdev.com/) and when directly querying the transaction service, the correct period is returned. However, it isn't [via the gateway](https://safe-client.staging.gnosisdev.com/v1/chains/4/safes/0x9913B9180C20C6b0F21B6480c84422F6ebc4B808/multisig-transactions/?execution_date__gte=2021-05-20T00:00:00.000Z&execution_date__lte=2021-12-31T00:00:00.000Z&executed=true).

fmrsabino:
Closes #904

- `execution_date_lte` needs to be serialized from `execution_date__lte` and not from `execution_date_lte`
- This matches the contract with the clients and also the matching query parameter on the Safe Transaction Service

### DIFF
--- a/src/routes/transactions/filters/multisig.rs
+++ b/src/routes/transactions/filters/multisig.rs
@@ -7,7 +7,7 @@ use std::fmt;
 pub struct MultisigFilters {
     #[field(name = "execution_date__gte")]
     pub execution_date_gte: Option<String>,
-    #[field(name = "execution_date_lte")]
+    #[field(name = "execution_date__lte")]
     pub execution_date_lte: Option<String>,
     pub to: Option<String>,
     pub value: Option<String>,


rmeissner (approved):

src/routes/transactions/filters/multisig.rs
```diff
@@ -7,7 +7,7 @@ use std::fmt;
 pub struct MultisigFilters {
     #[field(name = "execution_date__gte")]
     pub execution_date_gte: Option<String>,
-    #[field(name = "execution_date_lte")]
+    #[field(name = "execution_date__lte")]
     pub execution_date_lte: Option<String>,
```
rmeissner, line 11: I assume that the difference to the rust field name is on purpose, right?

PR #924, opened 2022-07-14
Add /preview endpoint

### LINKED ISSUE #840: Return known addresses and decoded data when available
When getting decoded data in [reponse](https://safe.global/safe-client-gateway/docs/routes/contracts/routes/fn.post_data_decoder.html) we would like to get the known address.

[Problem statement](https://github.com/safe-global/safe-pm/issues/76)
[Mobile issue](https://github.com/safe-global/safe-ios/issues/1710)
[Slack thread](https://gnosisinc.slack.com/archives/CQB729LLF/p1648559663170409)

The nesting should be capped to be at most 1 level deep

Nope

**Additional context**
We want to display decoded data for incoming WalletConnect transactions.

**How it looks currently**

IMG_0861

**What we aim**

Screenshot 2022-05-24 at 11 31 23

Screenshot 2022-05-24 at 11 32 10

**Solution ideas**

Provide `addressInfoIndex` object the same way as it is done for transaction details.

fmrsabino:
Closes #840

- Adds new endpoint under `POST /v1/chains/<chain_id>/transactions/<safe_address>/preview`
- This endpoint is behind a feature flag – to enable it, set `FEATURE_PREVIEW_ENDPOINT` to `true` in your environment
- This endpoint takes some transaction data and returns a `TransactionInfo` (used for rendering the header) and the respective `TransactionData`.
- The clients should provide in the body of the request the following parameters: `to` (mandatory), `data` (optional), `value` (mandatory) and `operation` (mandatory)
- If there's an error while trying to get the decoded data, then the original raw data is returned – ie.: this endpoint does not fail if there's an issue while decoding the data or if it's `null`
- If the `to` address has no mapping (to a known address) then the original `to` address is returned with the other "known address" fields set to `null`

```javascript
POST /v1/chains/<chain_id>/transactions/<safe_address>/preview

client >>> service
{
  "to": <str>,
  "data": <optional-str>,
  "value": <str>,
  "operation": <int>
}

client <<< service
{
  "txInfo": TransactionInfo,
  "txData": TransactionData
}

```


src/routes/transactions/converters/mod.rs
```diff
@@ -21,7 +21,7 @@ use crate::utils::{SAFE_TRANSFER_FROM_METHOD, TRANSFER_FROM_METHOD, TRANSFER_MET
 use rocket::futures::future::OptionFuture;
 
 impl SafeTransaction {
-    async fn transaction_info(
+    pub(crate) async fn transaction_info(
```
fmrsabino, line 24: This is now used by the preview endpoint so I had to expose this to the other crate. It might make sense to review the project structure so that we know what can/should be public to the other crates

src/routes/transactions/tests/preview.rs
```diff
@@ -0,0 +1,141 @@
...
+        data_decoder_request.body(Some(
+            serde_json::to_string::<DataDecoderRequest>(&DataDecoderRequest {
+                data: data.to_string(),
+            })
+            .unwrap(),
+        ));
+        mock_http_client
+            .expect_post()
+            .with(eq(data_decoder_request))
+            .times(0) // TODO core_uri!(info_provider, "/v1/data-decoder/") throws Cached value not available (?)
```
fmrsabino, line 103: This still needs to be tackled. Somehow `core_uri!(info_provider, "/v1/data-decoder/")` is returning `Cached value not available`.

We expect this call to work (and therefore `times(1)`).

src/routes/mod.rs
```diff
@@ -51,6 +54,12 @@ pub fn active_routes() -> Vec<Route> {
         hooks::routes::flush
     ];
 
+    let preview_route = if is_preview_endpoint_enabled() {
```
rmeissner, line 57: Personal preference:
I would initialize the array and push the endpoint depending on the feature flag, rather than having a conditional initialization. This way it is easier to add more endpoints that are secured by feature flags.

src/routes/transactions/handlers/preview.rs
```diff
@@ -0,0 +1,123 @@
...
+
+#[derive(Serialize, Debug, PartialEq)]
+#[serde(rename_all = "camelCase")]
+#[cfg_attr(test, derive(Deserialize))]
+pub struct TransactionPreview {
+    pub tx_info: TransactionInfo,
+    pub tx_data: TransactionData,
+}
+
+pub async fn preview_transaction(
```
rmeissner, line 33: Currently we do all the sub-requests sequentially. Did you check how hard it is to do them in parallel?

src/routes/transactions/handlers/preview.rs
```diff
@@ -0,0 +1,123 @@
...
+
+    let transaction_data =
+        TransactionData {
+            hex_data: preview_request.data.clone(),
+            data_decoded,
+            to: to_address_ex,
+            value: Some(preview_request.value.to_string()),
+            operation: preview_request.operation,
+            trusted_delegate_call_target: is_trusted_delegate_call,
+            address_info_index: OptionFuture::from(safe_transaction.data_decoded.as_ref().map(
```
rmeissner, line 90: Personal preference:

I would do this as a separate statement not inline when creating the struct.

src/routes/transactions/handlers/preview.rs
```diff
@@ -0,0 +1,123 @@
...
+    context: &RequestContext,
+    info_provider: &(impl InfoProvider + Sync),
+    data: &str,
+) -> ApiResult<DataDecoded> {
+    let data_decoder_endpoint = core_uri!(info_provider, "/v1/data-decoder/")?;
+    let client = context.http_client();
+
+    let request = {
+        let mut request = Request::new(data_decoder_endpoint);
+        let body = DataDecoderRequest {
```
rmeissner, line 115: Not sure if it makes sense in this PR, but the endpoint also has a `to` parameter as far as I know which allows for more correct abi decoding.

src/routes/transactions/tests/preview.rs
```diff
@@ -0,0 +1,141 @@
...
+use crate::config::contract_info_request_timeout;
+use crate::providers::info::*;
+use crate::routes::contracts::models::DataDecoderRequest;
+use crate::routes::transactions::handlers::preview::TransactionPreview;
+use crate::routes::transactions::tests::PREVIEW_RESPONSE;
+use crate::tests::main::setup_rocket;
+use crate::utils::errors::ApiResult;
+use crate::utils::http_client::{MockHttpClient, Request, Response};
+
+fn get_chain_info() -> ChainInfo {
```
rmeissner, line 20: Wouldn't it be more consistent to have this as a json file, like we do with the transactions?

src/routes/transactions/tests/preview.rs
```diff
@@ -0,0 +1,141 @@
...
+    ChainInfo {
+        recommended_master_copy_version: "1.1.1".to_string(),
+        transaction_service: "https://safe-transaction.mainnet.gnosis.io".to_string(),
+        vpc_transaction_service: "http://mainnet-safe-transaction-web.safe.svc.cluster.local"
```
rmeissner, line 24: could we not expose internal information here? :D

src/routes/transactions/tests/preview.rs
```diff
@@ -0,0 +1,141 @@
...
+        gas_price: vec![GasPrice::Fixed {
+            wei_value: "1000000".to_string(),
+        }],
+        disabled_wallets: vec![],
+        features: vec![],
+    }
+}
+
+#[rocket::async_test]
+async fn post_preview_success() {
```
rmeissner, line 68: I would expect a couple more tests (e.g. what if a sub-request fails)

src/routes/transactions/tests/preview.rs
```diff
@@ -0,0 +1,141 @@
...
+    let mut mock_info_provider = MockInfoProvider::new();
+    let data = "0x095ea7b3000000000000000000000000ae9844f89d98c150f5e61bfc676d68b4921559900000000000000000000000000000000000000000000000000001c6bf52634000";
+    mock_info_provider
+        .expect_chain_info()
+        .returning_st(move || Ok(get_chain_info()));
+
+    let mock_http_client = {
+        let mut mock_http_client = MockHttpClient::new();
```
rmeissner, line 76: I assume the advantage of assigning the return value of this block is that we can limit the `mut`?

src/routes/mod.rs
fmrsabino: Do you mean having a function that returns routes under feature flags and inside that function we check the feature flags one by one?

src/routes/transactions/handlers/preview.rs
fmrsabino: I didn't, initially for simplicity reasons. I can look into this.

(My preference would be to have this available for the clients as soon as possible so we can do some more testing on `staging`)

src/routes/transactions/handlers/preview.rs
fmrsabino: Yes you are right. Will add it too! 👍

src/routes/transactions/tests/preview.rs
fmrsabino: This should work but I'd need to move the "mock" to the httpclient (which I also prefer)

fmrsabino: The mock on the InfoProvider was done on many other tests. I'd need to review why it was done like that

src/routes/transactions/tests/preview.rs
fmrsabino: Yes I agree. The test setup that we have didn't help and I wanted to provide something that was more in a reviewable state.

But I completely agree with you.

src/routes/transactions/tests/preview.rs
fmrsabino: Yes. The setup is done inside the block (mimicking a builder pattern) and then we expose it as non-mutable reference.

For the test case it might not be required that we are this strict since the scope of the mutable value is just the test function itself.

I'd push it more for readability in this case (ie.: if it makes the test easier to read or not)

src/routes/transactions/handlers/preview.rs
```diff
@@ -0,0 +1,123 @@
...
+}
+
+#[derive(Serialize, Debug, PartialEq)]
+#[serde(rename_all = "camelCase")]
+#[cfg_attr(test, derive(Deserialize))]
+pub struct TransactionPreview {
+    pub tx_info: TransactionInfo,
+    pub tx_data: TransactionData,
+}
+
```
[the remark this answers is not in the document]
fmrsabino: The reason why I didn't do this is because I don't see the value right now on splitting this feature in more files.

I'd keep things as close as possible for a specific feature which allows us to better encapsulate them in the future (example to remove the `pub`).

But there's no strict rule regarding this. We can discuss how to proceed in a meeting 👍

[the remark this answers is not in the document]
moisses89: Ok, understood, it is a very specific model for this feature (preview) only then we could do an exception, migrate this to a new file in the future if was necessary is not too complicated.

src/routes/transactions/tests/preview.rs
fmrsabino: Addressed in [3ddb403](https://github.com/safe-global/safe-client-gateway/pull/924/commits/3ddb4035096ab6434cac56ab9838a9a089bf711b)

src/routes/transactions/handlers/preview.rs
fmrsabino: Addressed in [3ddb403](https://github.com/safe-global/safe-client-gateway/pull/924/commits/3ddb4035096ab6434cac56ab9838a9a089bf711b)

src/routes/transactions/handlers/preview.rs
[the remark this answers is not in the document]
fmrsabino: Exactly. Overall it's better to start with a smaller scope and as soon as we see that a new endpoint requires this structure we see where we should start creating a baseline for the "shared" models.

It'd be more difficult to start with the shared part and then see where we can have them private or not.

src/routes/transactions/handlers/preview.rs
fmrsabino: Added [ec44a5b](https://github.com/safe-global/safe-client-gateway/pull/924/commits/ec44a5b47cda1686e8158df3dc0cfece7371ae2b)

rmeissner (dismissed):

src/config/mod.rs
```diff
@@ -173,6 +173,10 @@ pub fn feature_flag_balances_rate_implementation() -> bool {
     env_with_default("FEATURE_FLAG_BALANCES_RATE_IMPLEMENTATION", false)
 }
 
+pub fn is_preview_endpoint_enabled() -> bool {
+    env_with_default("FEATURE_PREVIEW_ENDPOINT", false)
+}
+
```
moisses89, line 179: One question, should we add `FEATURE_PREVIEW_ENDPOINT` to `.env.sample`? what do you think?

moisses89 (dismissed):

src/config/mod.rs
fmrsabino: Added in [fe778a4](https://github.com/safe-global/safe-client-gateway/pull/924/commits/fe778a476fa653ec794139060cc20ccf6d22ce4a)

moisses89 (approved):
Looks good, great job! :rocket:

PR #925, opened 2022-07-14
Add paginated collectibles

### LINKED ISSUE #920: Add paginated endpoint for collectibles
Long lists of collectibles ERC720 takes a lot of time to return to the client.   
 
- https://github.com/safe-global/safe-pm/issues/89

Include a new version of endpoint to support pagination of transaction service   
- https://github.com/safe-global/safe-transaction-service/issues/239

moisses89:
Closes #920 
- add paginated collectibles 
- add test for paginated collectibles

### DIFF
--- a/src/routes/collectibles/handlers.rs
+++ b/src/routes/collectibles/handlers.rs
@@ -41,14 +41,15 @@ pub async fn collectibles_paginated(
     context: &RequestContext,
     chain_id: &str,
     safe_address: &str,
+    cursor: &Option<String>,
     trusted: Option<bool>,
     exclude_spam: Option<bool>,
-    cursor: &Option<String>,
 ) -> ApiResult<Page<ServiceCollectible>> {
     let info_provider = DefaultInfoProvider::new(chain_id, &context);
     let page_metadata = cursor
         .as_ref()
         .map(|cursor| PageMetadata::from_cursor(cursor));
+
     let url = core_uri!(
         info_provider,
         "/v2/safes/{}/collectibles/?{}&trusted={}&exclude_spam={}", /* paginated service in core is behind v2 api. */
--- a/src/routes/collectibles/handlers.rs
+++ b/src/routes/collectibles/handlers.rs
@@ -1,5 +1,3 @@
-use std::cmp::max;
-
 use crate::cache::cache_operations::RequestCached;
 use crate::cache::manager::ChainCache;
 use crate::common::models::page::{Page, PageMetadata};
@@ -53,7 +51,7 @@ pub async fn collectibles_paginated(
 
     let url = core_uri!(
         info_provider,
-        "/v2/safes/{}/collectibles/?{}&trusted={}&exclude_spam={}", /* paginated service in core is behind v2 api. */
+        "/v2/safes/{}/collectibles/?{}&trusted={}&exclude_spam={}",
         safe_address,
         page_metadata
             .as_ref()
--- a/src/routes/collectibles/handlers.rs
+++ b/src/routes/collectibles/handlers.rs
@@ -10,6 +10,7 @@ use crate::utils::context::RequestContext;
 use crate::utils::errors::ApiResult;
 use crate::utils::urls::build_absolute_uri;
 use rocket::response::content::RawJson;
+use rocket::serde::json::Json;
 
 pub async fn collectibles(
     context: &RequestContext,
@@ -44,7 +45,7 @@ pub async fn collectibles_paginated(
     cursor: &Option<String>,
     trusted: Option<bool>,
     exclude_spam: Option<bool>,
-) -> ApiResult<Page<ServiceCollectible>> {
+) -> ApiResult<Json<Page<ServiceCollectible>>> {
     let info_provider = DefaultInfoProvider::new(chain_id, &context);
     let page_metadata = cursor
         .as_ref()
@@ -69,7 +70,7 @@ pub async fn collectibles_paginated(
 
     let page = serde_json::from_str::<Page<ServiceCollectible>>(&body)?;
 
-    Ok(page.map_inner(|link| {
+    Ok(Json(page.map_inner(|link| {
         map_link(
             context,
             link,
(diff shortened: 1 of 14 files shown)


src/routes/collectibles/routes.rs
```diff
@@ -165,3 +170,45 @@ pub async fn get_collectibles(
...
+#[get("/v2/chains/<chain_id>/safes/<safe_address>/collectibles?<cursor>&<trusted>&<exclude_spam>")]
+pub async fn get_collectibles_paginated(
+    context: RequestContext,
+    chain_id: String,
+    safe_address: String,
+    cursor: Option<String>,
+    trusted: Option<bool>,
+    exclude_spam: Option<bool>,
+) -> ApiResult<content::RawJson<String>> {
+    CacheResponse::new(&context, ChainCache::Other)
```
moisses89, line 200: I will review this, ChainCache?

moisses89: @fmrsabino what do you think? would be necessary caching the response?

Uxio0: I think we should always cache in the cgw, as in the other endpoints

src/routes/collectibles/handlers.rs
```diff
@@ -30,3 +36,76 @@ pub async fn collectibles(
...
+        .map(|cursor| PageMetadata::from_cursor(cursor));
+
+    let url = core_uri!(
+        info_provider,
+        "/v2/safes/{}/collectibles/?{}&trusted={}&exclude_spam={}", /* paginated service in core is behind v2 api. */
+        safe_address,
+        page_metadata
+            .as_ref()
+            .unwrap_or(&PageMetadata::default())
+            .to_url_string(),
```
fmrsabino, line 61: nitpick: I'd move this up where you map from the `cursor` (lines 50 to 52) since we don't need the `page_metadata` for anything else

src/routes/collectibles/handlers.rs
```diff
@@ -30,3 +36,76 @@ pub async fn collectibles(
...
+    exclude_spam: Option<bool>,
+) -> ApiResult<Json<Page<ServiceCollectible>>> {
+    let info_provider = DefaultInfoProvider::new(chain_id, &context);
+    let page_metadata = cursor
+        .as_ref()
+        .map(|cursor| PageMetadata::from_cursor(cursor));
+
+    let url = core_uri!(
+        info_provider,
+        "/v2/safes/{}/collectibles/?{}&trusted={}&exclude_spam={}", /* paginated service in core is behind v2 api. */
```
fmrsabino, line 56: I would remove this comment since this should be self-explanatory

src/routes/collectibles/handlers.rs
```diff
@@ -30,3 +36,76 @@ pub async fn collectibles(
...
+    original_link: Option<String>,
+    chain_id: String,
+    safe_address: String,
+    trusted: Option<bool>,
+    exclude_spam: Option<bool>,
+) -> Option<String> {
+    original_link.as_ref().map(|link| {
+        let cursor =
+            PageMetadata::from_cursor(link.split("?").collect::<Vec<&str>>().get(1).unwrap_or(&""))
+                .to_url_string();
```
fmrsabino, line 96: For the context of this PR this is ok but we need to think of ways to refactor this. This logic is spread across multiple places.

Also URL parsing/manipulation should probably be `Url` which we do have in the project.

src/tests/json/collectibles/collectibles_page.json
```diff
@@ -108,3 +108,4 @@
         }
     }
 ]
+
```
fmrsabino, line 111: We should revert this change (double new line at the end)

src/routes/collectibles/handlers.rs
```diff
@@ -1,10 +1,16 @@
+use std::cmp::max;
```
fmrsabino, line 1: This is not required right?

src/routes/collectibles/routes.rs
```diff
@@ -1,8 +1,16 @@
+use crate::cache::cache_operations::CacheResponse;
+use crate::cache::manager::ChainCache;
```
fmrsabino, line 2: These should also not be needed

src/routes/collectibles/routes.rs
```diff
@@ -1,8 +1,16 @@
...
+use crate::common::models::page::Page;
+use crate::config::chain_info_response_cache_duration;
```
fmrsabino, line 4: This one should also not be needed.

fmrsabino (changes requested):
Looks good 👏 

Just some small requests.

src/routes/collectibles/routes.rs
fmrsabino: > I think we should always cache in the cgw, as in the other endpoints

That's correct. However for the collectibles, the first version of the endpoint only cached the request to the Safe Transaction Service and not the response given by the Safe Client Gateway to the clients.

> @fmrsabino what do you think? would be necessary caching the response?

I would say that we can start by not caching the response to the clients. We might be introducing unnecessary complexity. I'd roll out this endpoint and keep track of its usage.

moisses89: Ok, I agree with only chaching the request.

src/routes/collectibles/handlers.rs
moisses89: yes, It's not required

src/routes/collectibles/handlers.rs
moisses89: I agree, what do you think to create an issue to improve the manage of pagination and refactoring the source?

src/routes/collectibles/handlers.rs
moisses89: Clearest

src/routes/collectibles/handlers.rs
fmrsabino: Yes I think we should create one. No need to specify the solution (just that it needs to be evaluated)

fmrsabino (approved):

PR #929, opened 2022-07-20
Add Swagger link to README

### LINKED ISSUE #917: Add links to Swagger and Rust types to Readme
Please add links to Swagger and Rust types (https://safe.global/safe-client-gateway/docs/routes/index.html) in the Readme and the repo description.

Screenshot 2022-07-10 at 10 08 16

fmrsabino:
Closes #917

- Add link to Swagger (production)
- Update entry: "Client Gateway Wiki" to "Client Gateway Docs" to represent better the resource on this link

The original issue also suggests to add to the project description – I did not do that because I think the description should be a short text (as we have right now) and we'd need to keep track of more places if this url is updated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ This project is a gateway between the Safe clients ([Android](https://github.com
 ## Documentation
 
 - [Client Gateway Docs](https://safe.global/safe-client-gateway/)
-- [Swagger](https://safe-client.gnosis.io/swagger/index.html)
+- [Swagger](https://safe-client.gnosis.io/index.html)
 - [Safe developer documentation](https://docs.gnosis.io/safe/)
 
 ## Quickstart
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -8,7 +8,8 @@ This project is a gateway between the Safe clients ([Android](https://github.com
 
 ## Documentation
 
-- [Client Gateway Wiki](https://safe.global/safe-client-gateway/)
+- [Client Gateway Docs](https://safe.global/safe-client-gateway/)
+- [Swagger](https://safe-client.gnosis.io/swagger/index.html)
```
moisses89, line 12: Maybe we should wait for #927 in production. What do you think?

fmrsabino: Yes makes sense! I was thinking on updating it afterwards but this is also not critical 👍

moisses89 (approved):

PR #931, opened 2022-07-22
Add response types

### LINKED ISSUE #928: Swagger: add response types
Posting here as per @FahdSaif's request.

The swagger currently only shows the shapes of the requests but not the responses.

moisses89:
Related #928 
Add response types for: 
- collectibles v1
- about
- delegates
- safe_apps

### DIFF
--- a/src/routes/about/routes.rs
+++ b/src/routes/about/routes.rs
@@ -1,5 +1,4 @@
-use rocket::response::content;
-
+use super::models::About;
 use crate::cache::cache_operations::CacheResponse;
 use crate::cache::manager::ChainCache;
 use crate::common::routes::authorization::AuthorizationToken;
@@ -9,8 +8,9 @@ use crate::routes::about::handlers;
 use crate::utils::context::RequestContext;
 use crate::utils::errors::ApiResult;
 use crate::utils::http_client::Request;
+use rocket::response::content;
+use rocket::serde::json::Json;
 use rocket_okapi::openapi;
-
 /// `/v1/chains/<chain_id>/about` <br />
 /// Returns [ChainAbout](crate::routes::about::models::ChainAbout)
 ///
@@ -50,8 +50,8 @@ pub async fn get_chains_about(
 /// `/about`
 #[openapi(tag = "About")]
 #[get("/about")]
-pub async fn get_about() -> ApiResult<content::RawJson<String>> {
-    Ok(content::RawJson(serde_json::to_string(&handlers::about())?))
+pub async fn get_about() -> ApiResult<Json<About>> {
+    Ok(Json(handlers::about()))
 }
 /// `/v1/chains/<chain_id>/about/master-copies` <br />
 /// Returns a list of `MasterCopy`
--- a/src/routes/collectibles/handlers.rs
+++ b/src/routes/collectibles/handlers.rs
@@ -16,7 +16,7 @@ pub async fn collectibles(
     safe_address: &str,
     trusted: Option<bool>,
     exclude_spam: Option<bool>,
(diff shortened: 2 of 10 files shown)


src/routes/about/routes.rs
```diff
@@ -93,15 +96,12 @@ pub async fn get_master_copies(
...
-    context: RequestContext,
-    chain_id: String,
-) -> ApiResult<content::RawJson<String>> {
+pub async fn backbone(context: RequestContext, chain_id: String) -> ApiResult<Json<String>> {
     let client = context.http_client();
     let info_provider = DefaultInfoProvider::new(chain_id.as_str(), &context);
     let url = core_uri!(info_provider, "/v1/about/")?;
     let request = Request::new(url);
-    Ok(content::RawJson(client.get(request).await?.body))
+    Ok(Json(client.get(request).await?.body))
```
moisses89, line 104: This should return Json<BackBoneType>

moisses89:
@fmrsabino regarding this fragment of code:
```
CacheResponse::new(&context, ChainCache::from(chain_id.as_str()))
        .duration(about_cache_duration())
        .resp_generator(|| handlers::get_master_copies(&context, chain_id.as_str()))
        .execute()
        .await
```
CacheResponse->execute returns `ApiResult<RawJson<String>>>` 
Instead rewrite execute I would like to add a method that returns from cache the result in String and in each endpoint cast the result to `ApiResult<Json<CorrespondingType>>>`
What do you think about this way to do this?

fmrsabino:
> @fmrsabino regarding this fragment of code:
> 
> ```
> CacheResponse::new(&context, ChainCache::from(chain_id.as_str()))
>         .duration(about_cache_duration())
>         .resp_generator(|| handlers::get_master_copies(&context, chain_id.as_str()))
>         .execute()
>         .await
> ```
> 
> CacheResponse->execute returns `ApiResult<RawJson<String>>>` Instead rewrite execute I would like to add a method that returns from cache the result in String and in each endpoint cast the result to `ApiResult<Json<CorrespondingType>>>` What do you think about this way to do this?

Instead of casting I'd probably have it as a Generic output. This might be a bigger change however so you can also try parsing the `String` to a specific type using `serde` (altough this might result in unecessary serializations when we return from the route:  `String` -> `T` -> `U`)

fmrsabino:
> Closes #928

This PR shouldn't close this issue. The issue is more about having it on all endpoint and this is specific to the About the Collectibles

src/routes/collectibles/handlers.rs
```diff
@@ -26,13 +26,13 @@ pub async fn collectibles(
...
-        RequestCached::new_from_context(url, &context, ChainCache::from(chain_id))
-            .request_timeout(collectibles_request_timeout())
-            .execute()
-            .await?,
-    ))
+    let body = RequestCached::new_from_context(url, &context, ChainCache::from(chain_id))
+        .request_timeout(collectibles_request_timeout())
+        .execute()
+        .await?;
+    Ok(Json(serde_json::from_str::<Vec<ServiceCollectible>>(
```
rmeissner, line 33: This doesn't make any sense. We first serialize them as a string and then convert them back to structs to convert them back to strings. 

With our current caching we cannot use the `Json` type.

rmeissner (changes requested):

src/routes/collectibles/handlers.rs
rmeissner: Just to provider better documentation I would not add this additional parsing overhead

fmrsabino:
We are targeting this work in https://github.com/5afe/safe-client-gateway-nest

Issue #958, opened 2022-08-29
Add support for `url` param for `/v1/chains/<chain_id>/safe-apps`


mikhailxyz:
**Is your feature request related to a problem? Please describe.**
(the start of the sentence was suggested by the issue template) I'm always frustrated when I have to fetch the whole app list to get one app. The [config service](https://safe-config.gnosis.io/) supports the `url` query param for the `safe-apps` endpoint, which allows fetching a single app by exact URL match. It would be awesome if the client gateway also supported this query param.

**Provide references to the feature you are implementing that requires this change**
- https://safe-config.gnosis.io/
- The param is relevant for https://github.com/safe-global/web-core/issues/452 to display a warning if the app is not in the default list

**Describe the solution you'd like**
1:1 support for the URL param as on the config service

**Additional context**
The param is relevant for https://github.com/safe-global/web-core/issues/452 to display a warning if the app is not in the default list

PR #959, opened 2022-08-30
Feature: safe apps url query param

mmv08:
**This PR**:
- Adds support for `url` query param for `safe-apps` endpoint as in #958 

**Additional context**:
We already have a [provider](https://github.com/gnosis/safe-client-gateway/blob/248a9a1bf8b45fb8d6bcb80a5ef22d5fd78768cb/src/providers/info.rs#L157) for safe apps info which theoretically could be reused, but I didn't find a good way to glue it into the current problem

### DIFF
--- a/src/routes/safe_apps/tests/routes.rs
+++ b/src/routes/safe_apps/tests/routes.rs
@@ -1,5 +1,5 @@
 use crate::routes::safe_apps::models::SafeApp;
-use crate::routes::safe_apps::tests::RESPONSE_SAFE_APPS_WITH_TAGS;
+use crate::routes::safe_apps::tests::{RESPONSE_SAFE_APPS_WITH_TAGS, RESPONSE_SAFE_APPS_WITH_URL_QUERY};
 use crate::tests::main::setup_rocket;
 use crate::utils::errors::{ApiError, ErrorDetails};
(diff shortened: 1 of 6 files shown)


src/routes/safe_apps/tests/routes.rs
```diff
@@ -20,7 +22,7 @@ async fn safe_apps() {
     let mut mock_http_client = MockHttpClient::new();
 
     let safe_apps_request = Request::new(config_uri!(
-        "/v1/safe-apps/?chainId={}&clientUrl={}",
+        "/v1/safe-apps/?chainId={}&clientUrl={}&url=",
```
Uxio0, line 25: What's the purpose of adding url here if you are not using it?

mikhailxyz: I do not have a strong reasoning, but I think that it always gets added to config service request with an empty value even if not used: https://github.com/safe-global/safe-client-gateway/blob/248a9a1bf8b45fb8d6bcb80a5ef22d5fd78768cb/src/routes/safe_apps/handlers.rs#L15

Also, it's used the same way in other tests for the `client_url` param

Uxio0: Makes sense then, thanks 😉

fmrsabino (approved):
🚀

PR #970, opened 2022-09-28
Invalidate Mainnet cache in /v2/flush

fmrsabino:
- With the introduction of two caches (one for Mainnet and the other one for all the other networks), The `/v2/flush` route was executing the invalidation only for all the other chains.
- This means no Mainnet data was invalidated when `/v2/flush` was triggered.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2081,7 +2081,7 @@ checksum = "71d301d4193d031abdd79ff7e3dd721168a9572ef3fe51a1517aba235bd8f86e"
 
 [[package]]
 name = "safe-client-gateway"
-version = "3.35.0"
+version = "3.35.1"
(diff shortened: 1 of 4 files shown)


moisses89 (approved):

Uxio0 (approved):

fmrsabino:
> I think we should have a test but I understand is not easy as it's a production related issue and I don't know how mocking works on Rust

A simple test would be to isolate the route from the rest of the application and check how `Invalidate::execute` gets called (in terms of arguments).

However this requires some other changes like making the `AuthorizationToken::value` public so that we could create one for the tests. So it's definitely possible but the scope increases.

rmeissner (approved):

PR #1012, opened 2022-11-25
Group messages by date

fmrsabino:
- Groups messages from `GET /v1/chains/<chain_id>/safes/<safe_address>/messages` by day
- Introduces `DateLabel` as a type that can be returned by the message list – `type: DATE_LABEL`
- The grouping works only with the current page returned, if the next page contains entries on the same day, a `DATE_LABEL` would be returned for those too

### DIFF
--- a/src/routes/messages/frontend_models.rs
+++ b/src/routes/messages/frontend_models.rs
@@ -18,7 +18,7 @@ pub(super) struct Confirmation {
 
 #[derive(Serialize, Debug)]
 #[serde(tag = "type", rename_all = "SCREAMING_SNAKE_CASE")]
-pub(super) enum Message {
+pub(super) enum MessageItem {
     #[serde(rename_all = "camelCase")]
     Message {
         message_hash: String,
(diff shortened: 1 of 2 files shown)


Uxio0:
@fmrsabino Wouldn't it be easier if I add a filter on the tx service?

fmrsabino:
> @fmrsabino Wouldn't it be easier if I add a filter on the tx service?

Can you expand on this? This is about grouping messages that we get from a `Page` into different days ie.: it doesn't work on a `date` level but on a `page` level 🤔

Uxio0:
Ok, I got it wrong then, makes sense! I thought it was a problem to have a missing signature for a day

fmrsabino:
> Ok, I got it wrong then, makes sense! I thought it was a problem to have a missing signature for a day

Ah no. This is purely a UI component part – from a list of messages (on a page) we group them into different days. For each day we return an additional item (header):

```json
{
    "type": "DATE_LABEL",
    "timestamp": 1669161600000
}
```

Uxio0 (approved):

iamacook (approved):

PR #1044, opened 2023-01-12
Add client ip to performance logs

fmrsabino:
- Adds the remote client IP address to the performance logs
- The client IP is set as the last parameter of the performance log
- If it is not possible to retrieve the client IP, then its value is "-1"

Note for log entry parsers:

Instead of:

```
"MT::{method}::{route}::{delta}::{status_code}::{request_uri}::{chain_id}
```

You should consider:

```
"MT::{method}::{route}::{delta}::{status_code}::{request_uri}::{chain_id}::{client_ip}
```

### DIFF
--- a/src/monitoring/performance.rs
+++ b/src/monitoring/performance.rs
@@ -6,6 +6,8 @@ use rocket::{Data, Request, Response};
 
 pub struct PerformanceMonitor();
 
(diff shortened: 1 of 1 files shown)


moisses89 (approved):
Looks good to me! :rocket:

luarx (approved):

PR #1045, opened 2023-01-16
Backmerge release 3.45.0

fmrsabino:
Release: https://github.com/safe-global/safe-client-gateway/releases/tag/v3.45.0


fmrsabino:
3.44.0 release happened. The branch was never backmerged (https://github.com/safe-global/safe-client-gateway/tree/release-3.44.0) – this was an oversight on my side.

The only change from that branch was the version bump so there's no point in keeping it right now since this overrides it.

Uxio0 (approved):

PR #1050, opened 2023-01-20
Chore: merge json test data files for safe apps

mmv08:
Because safe app tags were hidden behind a feature flag, we had separate test files for safe apps with and without tags. [Now](https://github.com/safe-global/safe-client-gateway/pull/1046) tags are a production feature, we can have a single file.


fmrsabino (approved):
🚀🚀

PR #1053, opened 2023-01-23
Feature: Safe apps social links

mmv08:
Implements config-service changes from https://github.com/safe-global/safe-config-service/issues/739


fmrsabino (approved):
🚀 🚀 🚀

PR #1073, opened 2023-02-23
Add support for linux/arm64 cross compilation

fmrsabino:
- Adds support to compile to a `linux/arm64` target from a `linux/amd64` host.
- Specifies the linker for ARM64 targets in `.cargo/config.toml`
- Use the `vendored` feature from the `openssl` crate to compile the OpenSSL source and link it in the target architecture.
- Removes `ca-certificates` `libssl-dev` from the dependencies required from the runtime image (they are not required)
- `docker-publish-staging` now publishes both `linux/amd64` and `linux/arm64` images
- Formats `autodeploy` step (indentation/spaces)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,47 +1,61 @@
 # Build Image
 # match with version in rust-toolchain.toml file
-FROM rust:1.65.0-slim-buster as builder
+# BUILDPLATFORM is available inside Docker context. It contains the name of the host platform
+# See https://docs.docker.com/build/building/multi-platform/#building-multi-platform-images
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -1,6 +1,6 @@
 # Build Image
 # match with version in rust-toolchain.toml file
-FROM rust:1.65.0-slim-buster as builder
+FROM --platform=$BUILDPLATFORM rust:latest as builder
```
fmrsabino, line 3: I will pin the version before opening the PR for review

luarx (approved):

PR #1124, opened 2023-07-31
feature: Helm chart for Kubernetes deployment 

tolbrino:
Make sure these boxes are checked! 📦✅

- [x] You have a nightly compatible version of `rustfmt` installed
```bash
rustup component add rustfmt --toolchain nightly
```
- [x] You ran `cargo +nightly fmt` on the code base before submitting

### What was wrong? 👾

The repository was lacking a Helm chart for Kubernetes deployment.

### How was it fixed? 🎯

The Helm chart was added at `charts/safe-client-gateway`, a location adhering to Helm best-practises. The helper script at `scripts/build-and-publish-helm-chart.sh` can be used for releasing new versions of the Helm chart. Documentation of the chart is at `charts/safe-client-gateway`.

This chart is already used successfully to run the Safe infrastructure in Kubernetes.

## Sibling PRs

- https://github.com/safe-global/safe-wallet-web/pull/2330
- https://github.com/safe-global/safe-transaction-service/pull/1601
- https://github.com/safe-global/safe-config-service/pull/911

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,6 @@ dump.rdb
 **/venv
 
 *.DS_Store*
+
+# helm packages
+safe-client-gateway-*.tgz
--- a/charts/safe-client-gateway/.gitignore
+++ b/charts/safe-client-gateway/.gitignore
@@ -1,0 +1,1 @@
+values-testing.yaml
--- a/charts/safe-client-gateway/Chart.yaml
+++ b/charts/safe-client-gateway/Chart.yaml
@@ -1,0 +1,8 @@
+---
+apiVersion: v2
+name: safe-client-gateway
+version: 0.1.1
+appVersion: v3.53.0
+description: A Helm chart for installing Safe Client Gateway
+type: application
+icon: "https://raw.githubusercontent.com/safe-global/safe-core-sdk/main/assets/logo.png"
--- a/charts/safe-client-gateway/README.md
+++ b/charts/safe-client-gateway/README.md
@@ -1,0 +1,56 @@
+# Safe Client Gateway Chart
+
+This chart packages the Safe Client Gateway resources. The chart assumes that there is already an existing Redis instance available and connection attribute should be passed in the values of the Helm Chart
+
+## Parameters
+
+### Common parameters
+
+| Name               | Description                                        | Value |
+| ------------------ | -------------------------------------------------- | ----- |
+| `nameOverride`     | String to partially override common.names.fullname | `""`  |
+| `fullnameOverride` | String to fully override common.names.fullname     | `""`  |
+
+### Installation Parameters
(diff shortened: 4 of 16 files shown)


tolbrino:

tolbrino:
recheck

fmrsabino:
@tolbrino and @ausias-armesto thank you for your contribution.

> The repository was lacking a Helm chart for Kubernetes deployment.

This is not something that we need for the project. My recommendation is to fork the repository and change it according to your needs.

At the moment we won't specify on how the service should be deployed and the respective resources (that is up to each team using it). This applies to the suggested kubernetes configuration.

tolbrino:
> @tolbrino and @ausias-armesto thank you for your contribution.
> 
> > The repository was lacking a Helm chart for Kubernetes deployment.
> 
> This is not something that we need for the project. My recommendation is to fork the repository and change it according to your needs.
> 
> At the moment we won't specify on how the service should be deployed and the respective resources (that is up to each team using it). This applies to the suggested kubernetes configuration.

Thanks for the feedback. We are already working with our own forks, but that does not help out the Safe community as a whole since discovery of such additional changes is hard.

I understand you might not want to add these files into each repository. How about creating a new repository called `safe-charts` which holds all available charts for the various Safe infrastructure components? You may then also state that the repository is a community effort and not an official distribution, but at least its a single place for the community to find it.

fmrsabino:
> Thanks for the feedback. We are already working with our own forks, but that does not help out the Safe community as a whole since discovery of such additional changes is hard.

We try to address the community needs including feature-discovery. However a discussion about those needs is also required. The introduction of this tool and respective deployment files assumes that this how the project should be deployed and these tools should be used. We do not make any assumptions on how the deployment should work in this repo and we give that choice to the developers who want to run our services themselves. We do provide a Docker image to make it easier to deploy but even then we do not assume that the image itself should be used (e.g. a developer/team might want to use a competing solution to Docker).

> I understand you might not want to add these files into each repository. How about creating a new repository called `safe-charts` which holds all available charts for the various Safe infrastructure components? You may then also state that the repository is a community effort and not an official distribution, but at least its a single place for the community to find it.

If the suggestion is for it to live under the `safe-global` then this needs to be aligned internally. However, due to the open-source nature of the project, something like `safe-charts` can be created by any org. I cannot promise that this is something that will be done under `safe-global` (for the reasons that I mentioned).

Also, there was no discussion on why this change is necessary for the project and an evaluation on the community support behind it. Adding Helm to the repository assumes that Helm is the desired solution that we want for the community and so far that discussion did not take place. Since this repository mostly deals with the business logic for client-side rendering, the respective deployment of such service should in this case be considered on a separate repo (as you suggested).

PR #1126, opened 2023-08-25
Update README.md

fmrsabino:
- Updates the README.md file to state the current status of the project (as not being maintained anymore).
- Adds reference to the new NestJS implementation and adds recommendation to use that repository instead

Preview: https://github.com/safe-global/safe-client-gateway/blob/update-readmne/README.md


iamacook (approved):

fmrsabino:
It looks like we are having issues in the CI because grcov now requires Rust 1.70.0.

This dependency was downloaded via Github Actions and seems that it was not pinned correctly leading to updates that could break the pipeline.

I've removed it here: https://github.com/safe-global/safe-client-gateway/pull/1127

